PR #116, opened 2026-04-11
Fix - Use correct Jamf API paths for computer and mobile device groups

jp-cpe:
`jamf-cli pro backup` was returning 404's for computer groups and not exporting any mobile device groups at all 

===
1. Computer smart/static groups — Backup used GET /v1/computer-groups/{id}, which doesn’t exist, so every detail call 404’d. It now uses the real APIs: /v2/computer-groups/smart-groups and /v2/computer-groups/static-groups for list and get.

2. Mobile smart groups — Detail URL was wrong (/v1/mobile-device-groups/{id} instead of .../smart-groups/{id}). List rows use groupId / groupName, but backup only read id / name, so it treated every group as missing and exported nothing without recording a failure.

3. Mobile static groups — There was no backup row at all; we added /v1/mobile-device-groups/static-groups (list + get) and output under static-groups/mobile.

4. Tests — Added/updated backup tests (including mocks for the duplicate smart-groups / static-groups resource names).

### DIFF
--- a/internal/commands/pro_backup.go
+++ b/internal/commands/pro_backup.go
@@ -254,7 +254,19 @@ func listModernItems(ctx context.Context, client registry.HTTPClient, def Resour
 	var items []resourceItem
 	for _, m := range all {
 		id := extractID(m)
+		if id == "" {
+			// Mobile device group list APIs use groupId / groupName (see MobileDeviceGroup.yaml SmartGroup / StaticGroup).
+			id = extractField(m, "groupId")
+		}
 		name := extractName(m)
+		if name == "" {
+			if n, ok := m["groupName"].(string); ok && n != "" {
+				name = n
+			}
+		}
+		if name == "" {
+			name = id
+		}
 		if id != "" {
 			items = append(items, resourceItem{ID: id, Name: name})
 		}
--- a/internal/commands/pro_backup_test.go
+++ b/internal/commands/pro_backup_test.go
@@ -280,8 +280,9 @@ func TestBackup_ComputerGroups(t *testing.T) {
 			"/v2/computer-groups/static-groups": {200, `{"totalCount":1,"results":[{"id":"20","name":"Lab Macs","count":5}]}`},
 			// v2 static group detail
 			"/v2/computer-groups/static-groups/20": {200, `{"id":"20","name":"Lab Macs","description":"Computer lab","siteId":"-1"}`},
-			// mobile smart groups (empty — shares Name "smart-groups" with computer smart groups)
-			"/v1/mobile-device-groups/smart-groups": {200, `[]`},
+			// mobile smart/static (empty — same Resource Name as computer groups; FilterResources runs all)
+			"/v1/mobile-device-groups/smart-groups":  {200, `[]`},
+			"/v1/mobile-device-groups/static-groups": {200, `{"totalCount":0,"results":[]}`},
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


neilmartin83 (commented):
## PR #116 Review: Fix Jamf API paths for computer and mobile device groups

### What it does

Fixes four bugs in `pro backup` where group resources were either 404ing or silently exporting nothing:

1. **Computer smart groups** — was hitting `/v1/computer-groups/{id}` (doesn't exist) → now `/v2/computer-groups/smart-groups`
2. **Computer static groups** — same wrong v1 path, and was sharing the same list endpoint as smart groups → now `/v2/computer-groups/static-groups`
3. **Mobile smart groups** — detail URL was `/v1/mobile-device-groups/{id}` instead of `/v1/mobile-device-groups/smart-groups/{id}`, AND the list response uses `groupId`/`groupName` instead of `id`/`name`, so every entry was silently dropped
4. **Mobile static groups** — `ResourceDef` entry was entirely missing

### Correctness

The resource path fixes in `pro_resources.go` are straightforward and correct. The `listModernItems` changes in `pro_backup.go` correctly handle the `groupId`/`groupName` field names in the mobile device group API responses.

I traced through the fallback chain for all cases:

- **Normal resource** (`id`+`name` present): all fallbacks skipped, no behavior change
- **Mobile group** (`groupId`+`groupName`, no `id`/`name`): `extractID` → `""`, falls through to `groupId`; `extractName` → `""` (its internal `extractID` fallback also returns `""` since there's no `id` field), falls through to `groupName`. Correct.
- **Missing name** (has `groupId` but no `groupName`): final `name = id` fallback kicks in, file gets named by ID. Reasonable.

No regressions for existing resources — Classic resources use `listClassicItems` (untouched), and modern resources with standard `id`/`name` fields never reach the fallback code.

### Issues

**1. Style inconsistency in `groupName` extraction** (`pro_backup.go:261-264`)

`groupId` uses the existing `extractField` helper, but `groupName` does a manual type assertion:

```go
id = extractField(m, "groupId")       // uses helper
// ...
if n, ok := m["groupName"].(string); ok && n != "" {  // manual
    name = n
}
```

Should be:

```go
if n := extractField(m, "groupName"); n != "" {
    name = n
}
```

`extractField` already handles the string case identically, and handles float64 too (irrelevant for names, but no harm).

**2. Test mock response shape inconsistency**

In `TestBackup_ComputerGroups`, the mobile smart groups mock returns a raw array:
```go
"/v1/mobile-device-groups/smart-groups": {200, `[]`},
```

While mobile static groups in the same test returns paginated:
```go
"/v1/mobile-device-groups/static-groups": {200, `{"totalCount":0,"results":[]}`},
```

Both work because `FetchAllPaginated` auto-detects the format, but the inconsistency is unnecessary. Pick one style for empty responses. The paginated format is more realistic for these endpoints.

### Minor observations (not blocking)

- The `groupId`/`groupName` fallback is applied to all modern resources in `listModernItems`, not just mobile device groups. It's harmless (dead code for everything else), but if more API shape quirks accumulate, consider adding optional `IDField`/`NameField` to `ResourceDef` instead of growing the fallback chain.
- The old static-groups/computers entry shared the exact same `ListPath` and `GetPath` as smart-groups/computers (`/v1/computer-groups` for both) — meaning the old code was exporting the same data into both directories. The fix correctly separates them.

### Verdict

Solid bug fix, well-scoped, good test coverage. The two items above are minor — fix the `groupName` extraction for consistency and unify the empty mock shapes, then this is good to merge.

ktn-jamf (changes requested):
> [!WARNING]
> **Needs changes** — Fixes four broken API paths for computer and mobile device group backup (404s and silent data loss).
> Blocking: (1). Nice-to-have: (2), (3).

**Scope of review:** 3 files changed, +162/−5 lines. Read the full diff, `pro_backup.go` (listModernItems and extractField/extractName helpers), `pro_resources.go` (ResourceDef table), and `pro_backup_test.go` (new tests). Verified CI status.

**CI Status:** ❌ Failing — lint step (`gofumpt`) fails on `pro_backup_test.go:365`.

**Prior review status:** @neilmartin83 left a COMMENT review flagging two style issues (groupName extraction inconsistency, mock response shape inconsistency). Both are still unaddressed. I'm promoting the `extractField` one below as it's a real inconsistency; the mock shape issue I'm including as nice-to-have.

---

### Findings

**1. (blocking, correctness) — `pro_backup_test.go:365`: gofumpt formatting failure breaks CI**

CI is red because `gofumpt` flags line 365. This is the only lint failure blocking merge.

**Suggested fix:**
```bash
gofumpt -w internal/commands/pro_backup_test.go
```

**Fixed when:** `golangci-lint run ./internal/commands/...` passes with zero issues and CI goes green.

---

**2. (nice-to-have, flagged by @neilmartin83) — `pro_backup.go:263-265`: groupName extraction bypasses the `extractField` helper**

`groupId` is extracted via the existing `extractField` helper, but `groupName` uses a manual type assertion. This is inconsistent within the same function — the pattern should match:

```diff
- 		if name == "" {
- 			if n, ok := m["groupName"].(string); ok && n != "" {
- 				name = n
- 			}
- 		}
+ 		if name == "" {
+ 			if n := extractField(m, "groupName"); n != "" {
+ 				name = n
+ 			}
+ 		}
```

`extractField` already handles the string assertion identically and also handles `float64` (irrelevant for names, but no harm). The manual assertion is dead-simple so this isn't functionally different — it's purely about consistency within the function.

---

**3. (nice-to-have, flagged by @neilmartin83) — `pro_backup_test.go:284`: inconsistent empty mock response shapes**

In `TestBackup_ComputerGroups`, the mobile smart groups mock returns a raw array:
```go
"/v1/mobile-device-groups/smart-groups": {200, `[]`},
```
While mobile static groups returns the paginated envelope:
```go
"/v1/mobile-device-groups/static-groups": {200, `{"totalCount":0,"results":[]}`},
```

Both work because `FetchAllPaginated` auto-detects, but the inconsistency is confusing for the reader — pick one style. The paginated format is more realistic for these endpoints.

---

### What's done well

- The root cause analysis is thorough — four distinct bugs (wrong list paths, wrong detail paths, missing field mappings, missing resource definition) all addressed together.
- Good test coverage with separate tests for computer groups and mobile device groups, including realistic mock data that exercises the `groupId`/`groupName` fallback path.
- The `name = id` final fallback at line 267-269 is a nice defensive touch for any future API shape quirks.

neilmartin83:
Superseded by #153, which fixes the same computer-group and mobile-device-group backup failures via a generator-driven registry — every resource now derives its paths from the specs automatically, so this class of drift can't recur.

PR #117, opened 2026-04-11
fix(report): drop stale devices from update-status --scan-failures

### LINKED ISSUE #108: report update-status --scan-failures: Apple TVs show blank serial and device name
## Description

When running `pro report update-status --scan-failures`, Apple TVs appear in the failure rows with empty **Serial** and **Name** fields, while iPads and iPhones show the expected values.

## Steps to reproduce

```bash
jamf-cli pro report update-status --scan-failures
```

Apple TV rows in the failure detail table:

```
DEVICE NAME   SERIAL   OS VERSION   DEVICE TYPE   ...
              (blank)  tvOS 17.x    Mobile
```

## Expected behaviour

Apple TV serial numbers and device names should be populated the same way they are for iPads and iPhones.

## Root cause (suspected)

`buildDeviceLookup` in `pro_report_updates.go` builds the name/serial lookup by fetching `/v2/mobile-devices` (for serial numbers) and `/v2/mobile-devices/detail?section=GENERAL` (for display names). Apple TVs are enrolled as mobile devices, but one of the following is likely true:

1. The `/v2/mobile-devices` list does not return Apple TVs, so `mobileSerials` is never populated for them.
2. The managed software update plan failure records reference Apple TV device IDs using a key that differs from the `mobileDeviceId` field used in the detail response, causing the lookup to miss.
3. The `general.displayName` field is absent or structured differently for Apple TVs in the detail response.

The IDs from the update plan failures are looked up in the `lookup` map keyed by device ID — if Apple TVs don't land in that map, the enrichment silently produces blank fields.

## Environment

- jamf-cli v1.6.0
- Apple TV (tvOS)

neilmartin83:
## Summary

- `pro report update-status --scan-failures` now skips devices whose ID no longer exists in inventory, instead of showing blank rows
- Table header counts reflect post-filter totals
- Stderr message logged when stale devices are skipped (e.g. `Skipped 40 failed plan(s) for devices no longer in inventory.`)

## Root cause

The Jamf Pro managed software updates API (`/v1/managed-software-updates/plans` and `/v1/managed-software-updates/update-statuses`) retains references to devices that have been deleted or re-enrolled. The `device.href` in these records points to a 404. Previously these produced blank rows in both the error devices and failed plans tables — no name, serial, or OS version.

Verified against a live instance: 590 of 647 Apple TV plan entries resolve correctly via `/v2/mobile-devices`; the remaining 57 are stale references (40 deleted devices + 17 with mismatched `objectType` in the plans API pointing to non-existent mobile device IDs).

Closes #108

## Test plan

- [x] `make test` — all pass
- [x] New test `TestRunReportUpdateStatus_StaleDevicesDropped` — captures JSON output and asserts stale device 999 is absent from both `error_devices` and `failed_plans` arrays
- [x] Verified against live Jamf Pro instance (`acs-cob.datajar.mobi`)

### DIFF
--- a/internal/commands/pro_report_updates.go
+++ b/internal/commands/pro_report_updates.go
@@ -305,12 +305,17 @@ func runReportUpdateStatus(ctx context.Context, client registry.HTTPClient, scan
 	}
 
 	// Enrich error devices with inventory data.
+	// Devices whose ID no longer exists in inventory (stale plan/status
+	// references) are silently dropped — they have no useful data to show.
 	var lookup map[string]updateDeviceMeta
 	var errorRows []map[string]any
 	if len(errors) > 0 {
 		lookup = fetchUpdateDeviceLookup(ctx, client)
 		for _, e := range errors {
-			meta := lookup[e.deviceID]
+			meta, found := lookup[e.deviceID]
+			if !found {
+				continue
+			}
 			dt := meta.deviceType
 			if dt == "" {
 				dt = normalizeDeviceType(e.deviceType)
@@ -325,6 +330,9 @@ func runReportUpdateStatus(ctx context.Context, client registry.HTTPClient, scan
 				"product_key": e.productKey,
 				"updated":     e.updated,
 			})
+		}
+		if stale := len(errors) - len(errorRows); stale > 0 {
+			fmt.Fprintf(os.Stderr, "Skipped %d error device(s) no longer in inventory.\n", stale)
 		}
 	}
 
@@ -362,7 +370,10 @@ func runReportUpdateStatus(ctx context.Context, client registry.HTTPClient, scan
 			lookup = fetchUpdateDeviceLookup(ctx, client)
 		}
 		for _, fp := range failedPlans {
-			meta := lookup[fp.deviceID]
+			meta, found := lookup[fp.deviceID]
+			if !found {
+				continue
+			}
 			dt := meta.deviceType
 			if dt == "" {
 				dt = normalizeDeviceType(fp.deviceType)
@@ -380,6 +391,9 @@ func runReportUpdateStatus(ctx context.Context, client registry.HTTPClient, scan
 				"error":       fp.errors,
 				"last_event":  lastEvt,
 			})
+		}
+		if stale := len(failedPlans) - len(failedPlanRows); stale > 0 {
+			fmt.Fprintf(os.Stderr, "Skipped %d failed plan(s) for devices no longer in inventory.\n", stale)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


ktn-jamf (approved):
> [!TIP]
> **Merge-ready** — Drops stale device references from `pro report update-status --scan-failures` output, fixing blank rows for deleted/re-enrolled devices. No critical or important issues found.

## Findings

**Nice-to-have:**

🟩 **(1)** (performance) — `pro_report_updates.go:344-358`: Event fetching still includes stale device plan UUIDs

The `planUUIDs` list is built from `failedPlans` *before* the stale-device filter runs at line 372-376. This means `BoundedParallelFetch` will still make API calls to `/v1/managed-software-updates/plans/{uuid}/events` for plans that belong to stale devices — those event results are fetched, stored in `lastEvents`, then never used because the corresponding row gets dropped during enrichment.

This is harmless but wasteful. With live data showing 40+ stale devices, that's 40+ unnecessary API calls (rate-limited to 3 concurrent).

**Suggested fix:** Move event fetching after the stale-device filter, or pre-filter `failedPlans` against the lookup before building `planUUIDs`. Example:

```diff
+ // Pre-filter stale devices from failedPlans before event fetching.
+ if lookup == nil {
+     lookup = fetchUpdateDeviceLookup(ctx, client)
+ }
+ filteredPlans := make([]failedPlan, 0, len(failedPlans))
+ for _, fp := range failedPlans {
+     if _, found := lookup[fp.deviceID]; found {
+         filteredPlans = append(filteredPlans, fp)
+     }
+ }
+ stalePlans := len(failedPlans) - len(filteredPlans)
+ failedPlans = filteredPlans

  var planUUIDs []string
  for _, fp := range failedPlans {
      ...
  }
```

This also simplifies the later enrichment loop since you wouldn't need the `if !found { continue }` there anymore.

---

🟩 **(2)** (testing) — `pro_report_updates_test.go:240-242`: Test mutates package-level `outputFmt` without `t.Parallel()` guard

The test sets `outputFmt = "json"` and restores via `defer`. This is safe today because no tests in this file use `t.Parallel()`, but it's fragile. This matches the existing pattern in the file, so purely a follow-up observation.

---

## What's done well

- The fix correctly addresses the root cause at the right layer — rather than trying to fix the Apple TV lookup (which would still break for truly deleted devices), it handles the general case of stale references.
- Good test design: the mock includes both a valid device (ID `2`) and a stale device (ID `999`) across both error statuses and failed plans, verifying that filtering works in both code paths.
- The stderr messages are a nice touch for operational visibility without polluting stdout/machine-parsable output.

LGTM — both findings are nice-to-haves for a follow-up.

PR #115, opened 2026-04-10
feat: Platform API integration into power commands, audit, reports, and device views

neilmartin83:
## Summary

Closes #33

- **`pro overview`** — Platform section: blueprint/benchmark counts, compliance %
- **`pro device`** — DDM declarations (grouped by blueprint), blueprint scoping, compliance benchmarks per device
- **`pro audit --checks platform`** — 7 health checks: undeployed blueprints, deployment failures, stale blueprints, benchmark updates available, monitor-only benchmarks, empty scope, failed DDM declarations
- **`pro report`** — 4 new reports: `blueprint-status`, `compliance-rules`, `compliance-devices`, `ddm-status` (aggregated by blueprint, standalone filtered, actionable errors only)
- **`pro blueprints clone`** — copy blueprints with automatic `payloadIdentifier` randomization for legacy profiles
- **`pro backup/diff`** — blueprints and compliance-benchmarks as exportable/diffable resources
- **`pro group-tools --unused`** — now checks policies, profiles, restricted software, ebooks, patch policies, prestages, AND platform blueprints/benchmarks
- **`pro ddm-reports errors`** — devices with actionable DDM failures (filters `Info.DeclarationNotInstalled`)
- **Spinner** for Platform SDK HTTP calls
- **Pluralization fix** — `ddm-statuss` → `ddm-status`, `startup-statuss` → `startup-status`, `apns-client-push-statuss` → `apns-client-push-status`

DDM reports resolve declaration identifiers to their source blueprint name, classify sources (blueprint/system/standalone), and filter non-actionable reason codes from all display surfaces.

## Test plan

- [x] `make build && make test && make lint` — all pass, zero issues
- [x] 16 new platform tests (`pro_platform_test.go`): audit checks, overview, clone payload randomization, DDM classification, scope helpers
- [x] Live tested against EU platform gateway with real blueprints, benchmarks, and devices
- [x] Verified `--checks platform` audit, `report blueprint-status`, `report ddm-status`, `blueprints clone`, `group-tools --unused`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #119, opened 2026-04-11
feat: add destructive-cooldown to enforce delay between destructive operations

### LINKED ISSUE #67: Feature: destructive-cooldown profile setting to prevent rapid destructive operations
## Problem

There is currently no client-side protection against rapid execution of destructive commands (delete, wipe, lock) from unintended script loops or runaway automation. Server rate limiting is reactive and doesn't distinguish operation severity — it treats rapid deletes the same as rapid reads.

## Proposed Solution

Add an optional `destructive-cooldown` setting to config profiles:

```yaml
profiles:
  production:
    url: https://prod.jamfcloud.com
    auth-method: oauth2
    client-id: keychain:jamf-cli/production/client-id
    client-secret: keychain:jamf-cli/production/client-secret
    destructive-cooldown: 10s   # NEW
```

When set, the CLI enforces a minimum delay between any operation marked `IsDestructive`:
- Before executing, check time since last destructive op for this profile
- If within cooldown window, sleep for remaining duration (with message to stderr)
- Update timestamp after successful execution

## Behavior

```
$ jamf-cli -p production pro computers delete 123 --yes
Deleted computer 123.

$ jamf-cli -p production pro computers delete 456 --yes
Cooldown: waiting 7s before destructive operation...
Deleted computer 456.
```

## Implementation

| File | Change |
|------|--------|
| `internal/config/config.go` | Add `DestructiveCooldown time.Duration` to `Profile` struct |
| `internal/cooldown/cooldown.go` (new) | Cooldown state: `Load()`, `Check(profile)`, `Record(profile)` |
| `generator/parser/generator.go` | Inject cooldown check in `IsDestructive` template block |
| `internal/commands/pro_bulk_commands.go` | Add cooldown between items in bulk destructive ops |

**State file:** `~/.config/jamf-cli/cooldown.json`
```json
{
  "production": "2026-04-05T14:32:01Z",
  "staging": "2026-04-05T14:30:45Z"
}
```

## Config Options

| Setting | Type | Description |
|---------|------|-------------|
| `destructive-cooldown` | duration | Minimum time between destructive ops (e.g., `5s`, `1m`, `30s`) |

## Edge Cases

- **Parallel shells**: File-based state with advisory locking prevents bypass
- **Clock skew**: Use monotonic comparison where possible
- **Ctrl+C during wait**: Cooldown not recorded, next attempt waits again
- **Bulk commands**: Cooldown applies between each item, not just at start

## Not Included (Future Consideration)

- Per-resource-type cooldowns (e.g., longer for `EraseDevice` than `delete`)
- Global cooldown across all profiles
- `--skip-cooldown` override flag (intentionally omitted for safety)

neilmartin83:
## Summary

- Adds a per-profile `destructive-cooldown` config setting that enforces a minimum delay between destructive operations (delete, erase, wipe, lock, restart, shutdown, delete-multiple)
- Default delay is **10 seconds** when no profile setting is configured
- `--no-input` always bypasses the cooldown unconditionally — CI/CD pipelines are never affected
- Set `destructive-cooldown: 0s` to disable for a specific profile
- State is persisted to `~/.config/jamf-cli/cooldown.json`, keyed by profile name

## Config example

```yaml
profiles:
  production:
    url: https://prod.jamfcloud.com
    auth-method: oauth2
    client-id: keychain:jamf-cli/production/client-id
    client-secret: keychain:jamf-cli/production/client-secret
    destructive-cooldown: 10s   # optional; 10s is the default
```

## Behaviour matrix

| Scenario | Cooldown |
|----------|---------|
| Interactive, no profile config | 10s (default) |
| Interactive, `destructive-cooldown: 30s` | 30s |
| Interactive, `destructive-cooldown: 0s` | none (disabled) |
| `--no-input` | none, always |
| Env-var auth (no profile) | none |

## Live tests (nmartin.jamfcloud.com)

**Default 10s cooldown — back-to-back deletes:**
```
$ bin/jamf-cli pro categories delete 1582 --yes
Deleted successfully

$ bin/jamf-cli pro categories delete 1583 --yes
Cooldown: waiting 9.948s before destructive operation...
Deleted successfully
```

**Dry-run — exits before cooldown, no state written:**
```
$ bin/jamf-cli pro categories delete 1584 --yes -n
[dry-run] Would delete category 1584
```

**`--no-input` — no wait:**
```
$ bin/jamf-cli pro categories delete 1585 --yes
Deleted successfully

$ bin/jamf-cli pro categories delete 1586 --yes --no-input
Deleted successfully
```

**`destructive-cooldown: 0s` — explicitly disabled:**
```
$ bin/jamf-cli pro categories delete 1587 --yes
Deleted successfully

$ bin/jamf-cli pro categories delete 1588 --yes
Deleted successfully
```

**`destructive-cooldown: 3s` — custom duration:**
```
$ bin/jamf-cli pro categories delete 1584 --yes
Deleted successfully

$ time bin/jamf-cli pro categories delete 1589 --yes
Cooldown: waiting 2.936s before destructive operation...
Deleted successfully
        3.097 total
```

**State file after delete:**
```json
{"default":"2026-04-11T08:56:10.684187+01:00"}
```

## Implementation notes

- New `internal/cooldown` package handles all state load/save/enforce/record logic
- Injected into modern and classic generator templates — all generated destructive commands get it automatically
- `Record` fires on both DELETE 204 responses and successful non-DELETE destructive ops (erase, wipe, lock, etc.)
- `pro send-command` bulk loop also enforces cooldown between each destructive MDM command (EraseDevice, DeviceLock)
- Closes #67

## Test plan

- [x] Delete two resources back-to-back — second should wait *(live tested)*
- [x] `--no-input` — no wait, no prompt *(live tested)*
- [x] `destructive-cooldown: 0s` in profile — no wait *(live tested)*
- [x] `destructive-cooldown: 3s` in profile — 3s wait *(live tested)*
- [x] Dry-run (`-n`) — exits before cooldown fires *(live tested)*
- [x] Verify `~/.config/jamf-cli/cooldown.json` is written after each destructive op *(live tested)*


danjamf:
Not a critique on the code but just the options. Would it make sense to have a profile where `--no-input` also renders to a cooldown? I'm thinking agentic hookups to CLI or other script calls that would likely be using that switch and go bad quickly

ktn-jamf:
> [!WARNING]
> **Needs changes** — Adds per-profile destructive-cooldown that enforces a minimum delay between destructive operations (delete, erase, wipe, lock, etc.), configurable via `destructive-cooldown` in profile config. Solid design and clean integration into both modern and classic generator templates.
> Blocking: (1), (2). Nice-to-have: (3), (4).

**Scope of review**: All 6 hand-written source files (cooldown package, root.go, config.go, registry.go, pro_bulk_commands.go, both generator templates). Sampled 4 generated files (categories, classic_classes, erase_device_computers, advanced_mobile_device_searches) to verify template output. Verified `make test` and `make verify-generated` pass. Skipped line-by-line review of the remaining ~100 generated files since they are mechanically produced from verified templates.

**Prior review status**: One PR comment from @danjamf (not an inline code review) raising a design question about whether `--no-input` should optionally enforce cooldown for agentic scenarios. This is addressed in finding (3).

---

**Findings**

🟧 **(1) (test coverage) — No unit tests for `internal/cooldown/cooldown.go`**

This is the core safety mechanism of the PR and it has zero test coverage. The `Enforce` function has 5 distinct code paths (noInput bypass, zero-duration bypass, default duration, custom duration, empty profile bypass) and `Record` has 2 (empty profile, normal). The behaviour matrix in the PR description was verified manually, but there's nothing preventing regressions.

**Suggested fix:** Add `internal/cooldown/cooldown_test.go` covering at minimum:
```go
func TestEnforce_NoInput(t *testing.T)           // noInput=true → immediate return
func TestEnforce_ZeroDuration(t *testing.T)      // *0 configured → immediate return
func TestEnforce_EmptyProfile(t *testing.T)      // profileName="" → immediate return
func TestEnforce_NoPriorOp(t *testing.T)         // no state file entry → immediate return
func TestEnforce_CooldownElapsed(t *testing.T)   // enough time passed → immediate return
func TestEnforce_CooldownNotElapsed(t *testing.T) // not enough time → sleeps (use short duration like 50ms)
func TestEnforce_DefaultDuration(t *testing.T)   // configured=nil → uses 10s default
func TestRecord_EmptyProfile(t *testing.T)       // no-op
func TestRecord_WritesState(t *testing.T)        // verify file written with correct timestamp
```

Use `t.TempDir()` via `XDG_CONFIG_HOME` env override to isolate state files. The `statePath()` function already respects `XDG_CONFIG_HOME`, so this is straightforward.

**Fixed when:** `internal/cooldown/cooldown_test.go` exists with tests covering all 5 Enforce paths and both Record paths, and `make test` passes.

🟧 **(2) (correctness) — `cooldown.Record` called on non-204 DELETE responses regardless of actual success**

In the modern template (`generator/parser/generator.go:1542-1546`), for DELETE operations that don't return 204, the fallback path calls `Record` if `PrintResponse` returns nil:

```go
// generated output (e.g. categories.go:468-471):
err = ctx.Output.PrintResponse(resp)
if err == nil {
    cooldown.Record(ctx.ProfileName)
}
return err
```

While the HTTP client currently errors on `>= 400` status codes (so this path only fires for 2xx non-204 responses), this creates a hidden coupling. If the client's error-handling behavior ever changes, or if a server returns an unexpected 2xx status for a failed delete, `Record` would fire incorrectly. The same template path is also used for non-DELETE destructive ops (erase, wipe, lock) where the response semantics may differ.

**Suggested fix:** Guard `Record` on response status code in the template:
```diff
- err = ctx.Output.PrintResponse(resp)
- if err == nil {
-     cooldown.Record(ctx.ProfileName)
- }
- return err
+ err = ctx.Output.PrintResponse(resp)
+ if err == nil && resp.StatusCode >= 200 && resp.StatusCode < 300 {
+     cooldown.Record(ctx.ProfileName)
+ }
+ return err
```

Then `make generate` to propagate.

**Fixed when:** `Record` is only called for 2xx responses in the template fallback path, and `make verify-generated` passes.

🟩 **(3) (design) — Consider @danjamf's feedback on `--no-input` cooldown opt-in**

@danjamf raised a valid point: agentic/scripted callers often use `--no-input` but may still benefit from cooldown protection. The current design unconditionally bypasses cooldown for `--no-input`, which is reasonable for CI/CD but leaves agentic automation unprotected.

A potential middle ground: add a `destructive-cooldown-force: true` profile option that enforces cooldown even when `--no-input` is set. This keeps the default safe for CI/CD while giving admins an opt-in for agentic profiles. This doesn't need to block the current PR but is worth a follow-up issue.

**Fixed when:** This is a design discussion item. Considered addressed when the team has a decision documented (issue or PR comment) on whether to add forced cooldown for `--no-input` scenarios.

🟩 **(4) (code quality) — `noInputCooldown2` variable naming in classic template**

In `generator/classic/generator.go:538`, the variable `noInputCooldown2` is needed because `noInput` is scoped inside the `if !flagYes` block. The `2` suffix is a bit cryptic.

**Suggested fix:**
```diff
- noInputCooldown2, _ := cmd.Flags().GetBool("no-input")
- if err := cooldown.Enforce(ctx.ProfileName, noInputCooldown2, ctx.DestructiveCooldown); err != nil {
+ noInputForCooldown, _ := cmd.Flags().GetBool("no-input")
+ if err := cooldown.Enforce(ctx.ProfileName, noInputForCooldown, ctx.DestructiveCooldown); err != nil {
```

**Fixed when:** Variable has a descriptive name instead of a numeric suffix.

---

**What's done well:**

- ✅ Clean separation of concerns — the `cooldown` package is self-contained with no dependencies on the rest of the codebase, making it easy to test and reason about independently.
- ✅ The three-way nil/zero/value semantics for `*time.Duration` are well-designed: `nil` = default 10s, `*0` = disabled, `*N` = custom. This avoids the common trap of conflating "not set" with "zero".

**CI Status**: Tests pass locally. No CI checks visible on this PR yet.

neilmartin83:
@danjamf 👋🏼 been having a think - if people use unattended agents to run this (I know, but people gonna people... 🤣 ), cooldown will just delay the action for 10s and still do it. `--no-input` is for CI/CD pipelines and having it bypassing the cooldown period would save on runtime costs. And even with cooldown active for an agent, the person might not notice it waiting due to it.

I hope the other safety nets like `--yes` and `--confirm-destructive` will also help there, and agents will act with good ethics and not just add those to commands without permission. This is where people need to be careful - ideally they would only let agents use `jamf-cli` with a read-only API client or one that can't do deletes/send MDM commands etc.

I think it's an extremely valid concern and well worth opening up a separate issue about - maybe there should be something that can detect agent usage and outright block destructive commands unless some config option is explicitly set by the user. What do you think?

danjamf:
> @danjamf 👋🏼 been having a think - if people use unattended agents to run this (I know, but people gonna people... 🤣 ), cooldown will just delay the action for 10s and still do it. `--no-input` is for CI/CD pipelines and having it bypassing the cooldown period would save on runtime costs. And even with cooldown active for an agent, the person might not notice it waiting due to it.
> 
> I hope the other safety nets like `--yes` and `--confirm-destructive` will also help there, and agents will act with good ethics and not just add those to commands without permission. This is where people need to be careful - ideally they would only let agents use `jamf-cli` with a read-only API client or one that can't do deletes/send MDM commands etc.
> 
> I think it's an extremely valid concern and well worth opening up a separate issue about - maybe there should be something that can detect agent usage and outright block destructive commands unless some config option is explicitly set by the user. What do you think?

Agree with your comments. An unattended agent would better be configured without the destructive permissions.

neilmartin83:
Would you say that's the end user's responsibility (with a tool like this, great power...)? Or should we look for opportunities to put further guard rails in place, beyond the flags we have?

danjamf:
Perhaps renaming of the `full-admin` setup role to something that impresses the destructive nature? I know standard is the default, but the creation of `full-admin` in profiles is perhaps too easy a thing to do? Either way, not related to this FR and PR fix

ktn-jamf (approved):
**Re-review: Merge-ready** ✅

Adds per-profile `destructive-cooldown` enforcing a minimum delay between destructive CLI operations (delete, erase, wipe, lock, restart, shutdown). All prior review findings have been addressed.

**Prior review finding status:**

| Finding | Status |
|---|---|
| (1) No unit tests for `internal/cooldown` | **Fixed** — `cooldown_test.go` added with 9 tests covering all 5 `Enforce` paths and both `Record` paths. All pass. Uses `t.TempDir()` + `XDG_CONFIG_HOME` for isolation. |
| (2) `Record` called without status code guard | **Fixed** — Template fallback path now guards on `resp.StatusCode >= 200 && resp.StatusCode < 300`. Verified in both the template (`generator/parser/generator.go:1544`) and generated output (`categories.go:469`). |
| (3) @danjamf feedback on `--no-input` cooldown | **Addressed** — Discussion resolved in PR comments; team agreed this is a follow-up concern, not a blocker. |
| (4) `noInputCooldown2` naming | **Fixed** — Renamed to `noInputForCooldown` in classic template (`generator/classic/generator.go:538`) and propagated to generated files. |

**Additional verification on the fix commit (`70ae9fb`):**

- The merge conflict with `main` (Platform API integration from PR #115) was resolved correctly — `PlatformClient` field and SDK construction coexist with the cooldown profile resolution in `root.go`.
- `make verify-generated` passes — all 205 generated files (157 modern + 48 classic) are consistent with templates.
- All cooldown tests pass (`go test ./internal/cooldown/... -v`).
- CLAUDE.md updates from the merge are unrelated to this PR (Platform API documentation from #115).

All findings resolved — approving.

PR #121, opened 2026-04-12
feat: blueprint DDM components, profile import with native conversion, scope management

neilmartin83:
## Summary

- **Blueprint component scaffold generator** — parses OpenAPI specs for DDM component types, walks schemas to produce example JSON, generates `internal/blueprintcomponents/scaffolds.go` via `make generate`. New CLI commands: `pro blueprints components scaffold <id>` and `pro blueprints apply --scaffold --component <id>`.
- **Configuration profile conversion** (`internal/profileconvert/`) — converts `.mobileconfig` files and raw plists to DDM blueprint components. Strips Apple metadata keys, removes empty values, generates deterministic payload identifiers. Optional `--strip-defaults` fetches Apple schemas from GitHub to remove keys at default values.
- **Automatic legacy-to-DDM conversion** — `import-profile` automatically promotes compatible payloads to native DDM components instead of wrapping everything in `com.jamf.ddm-configuration-profile`:
  - `com.apple.mobiledevice.passwordpolicy` → `passcode-settings` (12 key mappings, boolean inversions, customRegex restructure)
  - `com.apple.applicationaccess` safari keys → `safari-settings` (7 keys, cookie policy enum conversion)
  - `com.apple.applicationaccess` deferral keys → `software-update-settings` (forceDelayed*/enforced*Delay → Deferrals)
  - `com.apple.applicationaccess` RSR keys → `software-update-settings` (allowRapidSecurityResponseInstallation/Removal → RapidSecurityResponse)
  - `com.apple.SoftwareUpdate` → `software-update-settings` (AutomaticActions, AllowStandardUserOSUpdates, Beta)
  - Multiple converters targeting the same component (e.g. deferrals + RSR + SoftwareUpdate) are deep-merged into a single component
  - Payloads without a converter stay wrapped in the configuration-profile component
  - A single profile with mixed payloads produces multiple components in one blueprint step
  - `--legacy` flag bypasses all DDM conversion and wraps everything in a single configuration-profile
  - Unsupported payload types are filtered by default (`--include-unsupported` to override)
  - `--strip-defaults` strips Apple defaults before DDM conversion so default-valued keys are not actively managed
- **Cross-instance blueprint cloning** — portable export format enriches scope group UUIDs with names/types, `apply` resolves names on the target instance. Scope override flags (`--computer-group`/`--mobile-device-group`) and multi fan-out support.
- **Scope management**: `pro blueprints scope add`/`remove` with name resolution and `--id`/`--name` lookup on deploy/undeploy.

## Test plan

- [x] `make build && make test && make lint` — all pass, 0 lint issues
- [x] Live-tested `import-profile` against all 48 computer profiles and 7 mobile device profiles on pro-stuartdea
- [x] Verified passcode policy conversion (mobile) produces native `passcode-settings` component
- [x] Verified restrictions profile conversion produces `safari-settings` + `software-update-settings` + configuration-profile wrapper
- [x] Verified software-update-settings renders correctly in Jamf UI (full schema with `Included: false` for non-managed sections)
- [x] Verified `com.apple.SoftwareUpdate` profile converts to native `software-update-settings` (AutomaticActions, Beta, AllowStandardUserOSUpdates)
- [x] Verified RSR keys from restrictions profile merge into `software-update-settings` (RapidSecurityResponse)
- [x] Verified mixed deferrals + SoftwareUpdate profile deep-merges into single `software-update-settings` component
- [x] Verified `--legacy` flag wraps all payloads without DDM conversion
- [x] Verified unsupported payload types are filtered by default (no more 400 validation errors)
- [x] Verified scope group resolution (Classic group names → platform UUIDs)
- [x] Verified cross-instance export→apply round-trip (pro-stuartdea → pro-jamfmsptemplate)
- [x] Unit tests for DDM converters covering all key mappings, boolean inversions, partial conversions, merging, edge cases
- [x] Deleted all test blueprints after verification


neilmartin83:
## Code review summary

Reviewed the full diff (4606 additions). Core conversion logic and code generation are solid — well-tested profileconvert package, clean generator pipeline, correct SchemaFetcher concurrency pattern.

### Issues found and fixed (3790d0a)

| # | Issue | Fix |
|---|-------|-----|
| 1 | `TestFetchPlatformOverview_NilClient` was a no-op (assigned to var then discarded) — triggered `unusedwrite` diagnostic | Replaced with `TestFetchPlatformOverview_UsesAllMockData` that exercises the function with populated mock data |
| 2 | RSQL escaping in `resolveGroupPlatformID` only escaped `\` and `"` — group names with `*`, `(`, `)`, `;` would produce malformed filters | Added those characters to the `strings.NewReplacer` |
| 3 | `--type` flag on `configuration-profile` and `import-profile` accepted arbitrary strings silently (e.g. `--type macos` would hit wrong API path) | Added early validation: must be `"computer"` or `"mobile"` |
| 4 | `html.UnescapeString` for XML entity decoding could confuse readers about why HTML unescaping is used on XML | Added clarifying comment |
| 5 | `newUUID()` silently ignored `crypto/rand.Read` errors | Changed to panic — if the OS entropy source is broken, silently producing bad UUIDs is worse |
| 7 | `downloadClassicProfile` and `import-profile` didn't check HTTP status before parsing response body — a 404 produced a confusing "no `<payloads>` found" error | Added `resp.StatusCode != http.StatusOK` check with clear "not found (HTTP 404)" message |
| 8 | No unit tests for `extractAndResolveScope` (most complex helper in the PR) | Added 5 test cases: group resolution, missing scope, malformed scope, all-computers warning, dropped unsupported elements |
| 9 | No unit tests for `downloadClassicProfile` | Added 4 test cases: success, 404, missing payloads, nil client |

### Not changed (acceptable as-is)

- 10MB `readResponseBody` limit — generous but per-request single-user CLI, not a risk
- Apple schema URL pinned to `release` branch — reasonable default, would need a flag to make configurable

All tests pass (`go test ./...`), `go vet` clean, live-tested scaffold/conversion/validation commands.

ktn-jamf (commented):
> [!WARNING]
> **Needs changes** — Adds blueprint DDM component scaffolding, mobileconfig-to-DDM conversion with native payload promotion, cross-instance cloning, and scope management.
> Blocking: (1), (2). Nice-to-have: (3), (4), (5).

## Scope of review

Reviewed the full 8085-line diff across 27 files. Read the `internal/scope/types.go` for context (not in diff). Cross-referenced `CLAUDE.md`, repo conventions, and the author's self-review comment. CI passes.

## Prior review status

No formal GitHub reviews exist. The author posted a self-review comment (4231051782) listing 9 issues found and fixed in commit `3790d0a`. All items appear addressed in the current diff (RSQL escaping, `--type` validation, `newUUID` panic, HTTP status checks, test coverage for `extractAndResolveScope` and `downloadClassicProfile`).

## Findings

---

**1** (design) — `pro_blueprints.go`: `readResponseBody` duplicates existing HTTP body reading

`readResponseBody` at line ~3048 of the diff is a new standalone function:
```go
func readResponseBody(resp *http.Response) ([]byte, error) {
    return io.ReadAll(io.LimitReader(resp.Body, 10<<20))
}
```

This is called in three places (`downloadClassicProfile`, `import-profile`, `resolveGroupPlatformID`). The `internal/client` package already handles response reading for all generated commands. Introducing a parallel body-reading function here means: (a) the 10MB limit is only enforced in this file, not project-wide, (b) there's no consistent error handling for oversized responses, and (c) any future improvements to body reading (compression, logging) won't apply here.

**Suggested fix:** Either reuse the HTTP client's response handling, or if raw access is needed, move `readResponseBody` to `internal/client/` as an exported utility so the limit and error behavior are centralized.

**Fixed when:** `readResponseBody` is either removed in favor of existing infrastructure, or promoted to a shared utility. No inline 10MB constant in `pro_blueprints.go`.

---

**2** (correctness) — `pro_blueprints.go:extractAndResolveScope`: scope XML parsing uses string indexing, misses nested `<scope>` in `<limitations>` or `<exclusions>`

`extractAndResolveScope` finds the scope section via `strings.Index(xmlStr, "<scope>")` and then the end via `strings.Index(xmlStr[scopeStart:], "</scope>")`. This naive string search finds the **first** `</scope>` after the opening tag. The `<limitations>` and `<exclusions>` sections are nested *inside* `<scope>`, so they don't have their own `<scope>` tags — but if any Classic API response nests a `<scope>` inside scope (as some Jamf API responses do for nested scoping), this would silently truncate.

**Suggested fix:**
```diff
- scopeEnd := strings.Index(xmlStr[scopeStart:], "</scope>")
+ scopeEnd := strings.LastIndex(xmlStr[scopeStart:], "</scope>")
```

**Fixed when:** The scope end index finds the last `</scope>` after the opening tag, or the function uses `xml.Decoder` for proper nesting.

---

**3** (design) — `pro_blueprints.go`: duplicate code between `scope add` and `scope remove` commands

`newBlueprintsScopeAddCmd` and `newBlueprintsScopeRemoveCmd` share ~40 lines of identical boilerplate: flag declarations, `requirePlatformClient` gate, `hasNames` check, `resolveGroupNames` calls for both computer and mobile groups, and the "at least one flag" validation.

**Suggested fix:** Extract a `resolveAllGroupFlags` helper:
```go
func resolveAllGroupFlags(ctx context.Context, client registry.HTTPClient, groupIDs, computerGroups, mobileDeviceGroups []string) ([]string, error) {
    ids := append([]string{}, groupIDs...)
    resolved, err := resolveGroupNames(ctx, client, computerGroups, "COMPUTER")
    if err != nil { return nil, err }
    ids = append(ids, resolved...)
    resolved, err = resolveGroupNames(ctx, client, mobileDeviceGroups, "MOBILE")
    if err != nil { return nil, err }
    ids = append(ids, resolved...)
    if len(ids) == 0 {
        return nil, fmt.Errorf("specify at least one --group-id, --computer-group, or --mobile-device-group")
    }
    return ids, nil
}
```

**Fixed when:** The group resolution logic is shared between scope add, scope remove, and `apply --computer-group/--mobile-device-group`.

---

**4** (reliability) — `internal/profileconvert/defaults.go`: `SchemaFetcher` silently swallows all HTTP errors

`fetchAndParse` returns `nil, nil` for network errors, non-200 status codes, and read errors. The caller (`FetchDefaults`) then returns `nil, nil` and `StripConfigDefaults` skips the payload with a "no Apple schema available" message. This is the right degradation behavior for `--strip-defaults` (best-effort), but it makes debugging failures impossible. A schema fetch failing silently on every payload type because of a corporate proxy, TLS issue, or GitHub rate limit will produce a confusing "no Apple schema available" message 40+ times.

**Suggested fix:** Return the error for non-404 failures so the caller can log it:
```diff
  if resp.StatusCode == http.StatusNotFound {
      return nil, nil
  }
  if resp.StatusCode != http.StatusOK {
-     return nil, nil
+     return nil, fmt.Errorf("unexpected status %d from %s", resp.StatusCode, url)
  }
```
`StripConfigDefaults` already handles the error path by logging it and continuing, so this won't break the degradation behavior.

**Fixed when:** Non-404 HTTP errors are propagated with status code and URL, and appear in stderr output for debugging.

---

**5** (simplification) — `generator/blueprintcomponents/generator.go`: `walkBoolean` defaults to `true` for booleans without a default

```go
func walkBoolean(schema *openapi3.Schema) any {
    if schema.Default != nil {
        return schema.Default
    }
    return true
}
```

Defaulting to `true` is arbitrary and could produce scaffolds that enable features unintentionally (e.g., a new component with a `Dangerous` boolean field with no schema default would scaffold as `true`).

**Suggested fix:** Default to `false` for booleans without an explicit default in the schema. This is the safer starting point.

**Fixed when:** `walkBoolean` returns `false` when no default is specified.

---

## What's done well

- The DDM converter architecture is genuinely well-designed. The `convertFunc` signature with `(config, remaining, warnings, error)` allows clean chaining of partial converters, and the deep-merge strategy for multiple converters targeting the same component ID (e.g., deferrals + RSR + SoftwareUpdate all merging into `software-update-settings`) is elegant.

- The portable export format with enriched scope groups (`blueprintExportScopeGroup` with name + type) enabling cross-instance cloning is a strong UX improvement. The format detection in `isPortableScopeFormat` and transparent fallback to legacy UUID format is clean.

- Test coverage is thorough — the DDM converter tests (1447 lines) cover all key mappings, boolean inversions, partial conversions, merging, converter error fallback, and end-to-end orchestration with mixed payloads.

ktn-jamf (approved):
> [!TIP]
> **Merge-ready** — All prior findings addressed. No new issues in the fix commit.

## Re-review: Prior finding status

| # | Finding | Status |
|---|---------|--------|
| (1) | `readResponseBody` duplicates existing HTTP body reading | **Fixed** — Promoted to `internal/client/client.go` as exported `ReadResponseBody`. All 3 call sites updated, old function deleted, `"io"` import removed. |
| (2) | `extractAndResolveScope` XML `strings.Index` truncation risk | **Fixed** — Changed to `strings.LastIndex` so nested `</scope>` tags don't cause silent truncation. |
| (3) | Duplicate group resolution in scope add/remove | **Fixed** — Extracted `resolveAllGroupFlags` helper. Both scope commands use it. The `apply` command correctly keeps its own inline resolution (different semantics). |
| (4) | `SchemaFetcher` silently swallows non-404 HTTP errors | **Fixed** — Non-404 errors now return `fmt.Errorf(...)`. Degradation behavior preserved. |
| (5) | `walkBoolean` defaults to `true` for booleans without schema default | **Fixed** — Now returns `false`. Test updated. |

## New commit analysis

Fix commit (`9087abf`): 5 files, +36/−36 net. Each change is minimal and surgical. No new logic, no new code paths, no regressions.

**This covers all findings.**

PR #122, opened 2026-04-12
feat: compliance-benchmarks export, clone, portable apply, baseline scaffold

neilmartin83:
## Summary

- **`export <title>`** — portable JSON/YAML output replacing device group IDs with `{name, deviceType, groupType}` structs; output pipeable directly to `apply` for cross-instance cloning
- **`clone <source-title> <new-title>`** — creates a new benchmark from an existing one; `--computer-group` flag overrides target device groups on the destination
- **`delete [<id>]` + `--name`** — delete now takes a raw ID as primary arg (faster, no resolution); `--name` flag retained for title-based lookup
- **`apply` portable format** — reads `benchmarkPortableInput` (groups as `{name, type}` not IDs), resolves names→IDs on the target instance; `--computer-group` overrides groups; `--scaffold` prints a static template
- **`apply --scaffold-from-baseline <id>`** — generates a pre-populated scaffold from a baseline: injects sources, all rules (enabled by default), ODV placeholder enrichment (Placeholder → Value → `"<odv-value>"` fallback), and title/description pulled from baseline metadata
- **`enforcementMode`** default changed from `AUDIT` to `MONITOR` in all scaffold output

## Test plan

- [x] `TestBenchmarkToPortable_ResolvesGroupNames` — group IDs resolved to name+type in export
- [x] `TestBenchmarkToPortable_PreservesODV` — ODV values preserved through export round-trip
- [x] `TestCBScaffold_StaticTemplate` — static scaffold has MONITOR enforcement mode
- [x] `TestCBScaffoldFromBaseline` — baseline ID injected, title/description populated, all 3 ODV enrichment cases (placeholder / value fallback / sentinel)
- [x] `TestCBScaffoldFromBaseline_UnknownID` — unknown baseline ID surfaces error
- [x] `TestCBExport` — export command round-trips through portable format
- [x] `TestCBClone` / `TestCBClone_WithComputerGroupOverride` — clone creates new benchmark with/without group override
- [x] `TestCBDeleteByID` / `TestCBDeleteByName` / `TestCBDeleteNoArgs` — delete routing
- [x] `TestCBApply_ResolvesGroupNames` / `TestCBApply_ComputerGroupOverride` — apply resolves + overrides groups


ktn-jamf (changes requested):
> **Needs changes** — Adds export, clone, delete-by-ID, portable apply format, and baseline scaffold to compliance-benchmarks.
> Blocking: (1). Nice-to-have: (2), (3), (4).

**Scope of review**: All 862 added lines and 43 deleted lines across 3 files. Reviewed the full diff and checked surrounding context for `printScaffold`, `printExport`, `outputFmt`, `cbResolveTargetGroups`, and pattern consistency with existing Platform/Protect commands.

---

### Findings

**🟧 (1) (design) — Breaking change to `apply` input format with no backward compatibility**

The `apply` command previously accepted `jamfplatform.CBEngineBenchmarkRequestV2` directly, where `target.deviceGroups` was `[]string` (raw IDs). This PR changes it to `benchmarkPortableInput`, where `target.deviceGroups` is `[]benchmarkPortableGroup` (objects with `name`/`deviceType`/`groupType`). Since compliance-benchmarks shipped in v1.7.0, any user who saved scaffold output or crafted JSON matching the old format will get a parse error.

This is intentional (the portable format is better for cross-instance use), but the PR should either:
- Accept both formats (try portable first, fall back to the old SDK struct if group parsing fails), or
- Document this as a breaking change in the PR description and release notes so it's a deliberate decision, not a surprise.

**Suggested fix** — option A, backward compatibility:
```diff
  var input benchmarkPortableInput
  if err := unmarshalInput(data, &input); err != nil {
      return fmt.Errorf("parsing input: %w", err)
  }
+ // If target groups is empty, try parsing as legacy format ([]string IDs)
+ if len(input.Target.DeviceGroups) == 0 {
+     var legacy jamfplatform.CBEngineBenchmarkRequestV2
+     if err := unmarshalInput(data, &legacy); err == nil && len(legacy.Target.DeviceGroups) > 0 {
+         // Legacy format: pass IDs through directly, skip resolution
+         ... handle legacy path ...
+     }
+ }
```

Or option B — just acknowledge in the PR description that this is a breaking change to `apply` input. Since the feature is new (v1.7.0) and compliance benchmarks require platform gateway auth (narrow user base), this may be acceptable.

**Fixed when**: Either the apply command accepts both old `[]string` and new `[]benchmarkPortableGroup` formats for `target.deviceGroups`, OR the PR description explicitly calls out the breaking change and the team has consciously accepted it.

---

**🟩 (2) (reliability) — `cbScaffoldFromBaseline` silently swallows `ListBaselines` error**

`pro_compliance_benchmarks.go:608`: The `ListBaselines` call failure is silently eaten:
```go
if bls, err := cliCtx.PlatformClient.ListBaselines(ctx); err == nil {
```

If the list call fails (network blip, permission issue), the scaffold silently falls back to `"My Benchmark"` as the title with an empty description. The user gets no indication that metadata enrichment failed. A stderr warning would be more helpful.

**Suggested fix:**
```diff
- if bls, err := cliCtx.PlatformClient.ListBaselines(ctx); err == nil {
+ bls, blsErr := cliCtx.PlatformClient.ListBaselines(ctx)
+ if blsErr != nil {
+     fmt.Fprintf(os.Stderr, "Warning: could not fetch baseline metadata for title/description: %v\n", blsErr)
+ }
+ if blsErr == nil {
      for _, bl := range bls.Baselines {
```

**Fixed when**: A warning is emitted to stderr when `ListBaselines` fails, so the user knows why the scaffold has a generic title.

---

**🟩 (3) (correctness) — `printScaffold` YAML encoding doesn't call `Close()`**

`pro_platform_helpers.go:64-67`: The JSON path explicitly disables HTML escaping (`SetEscapeHTML(false)`) to preserve characters like `<`, `>`, `&` in scaffold placeholders. The YAML path doesn't need this (YAML encoders don't HTML-escape by default), so this is technically fine. But the `yaml.Encoder` also doesn't have a `Close()` call, which the `gopkg.in/yaml.v3` docs recommend to flush the stream.

**Suggested fix:**
```diff
  case "yaml":
      enc := yaml.NewEncoder(os.Stdout)
      enc.SetIndent(2)
-     return enc.Encode(v)
+     if err := enc.Encode(v); err != nil {
+         return err
+     }
+     return enc.Close()
```

**Fixed when**: `enc.Close()` is called after `enc.Encode(v)` to flush the YAML encoder stream.

---

**🟩 (4) (simplification) — Duplicate group resolution logic in `clone`**

`pro_compliance_benchmarks.go:268-278`: The clone command has inline group resolution logic that duplicates `cbResolveTargetGroups`. Since `cbResolveTargetGroups` already handles the "if overrideNames is non-empty, use those, otherwise use the portable groups" pattern, clone could delegate to it — but the source groups are raw IDs (strings), not `benchmarkPortableGroup`. Converting to portable groups just to call the helper would be awkward, so the current inline approach is pragmatically fine. This is a very minor "awareness" note, not something worth changing.

**Fixed when**: N/A — acceptable as-is.

---

### What's done well

- **Test coverage is thorough**: 10 focused test cases covering the happy paths, error paths, ODV enrichment cascade (placeholder → value → sentinel), and group override behavior. The `TestCBScaffoldFromBaseline` test in particular is well-structured, testing all three ODV branches in a single test with clear assertions.

- **`benchmarkToPortable` graceful degradation**: When a device group ID isn't found in the lookup map, it falls back to using the raw ID as the name instead of erroring. This is the right call — partial export is better than a hard failure when a group was deleted between benchmark creation and export.

- **Scaffold from baseline is a genuinely useful workflow**: Pre-populating all rules from a baseline with sensible ODV defaults removes a major friction point in benchmark creation. The three-tier ODV enrichment (Placeholder → Value → sentinel) is well thought through.

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Adds export, clone, portable apply format, backward-compatible legacy input, delete-by-ID, and baseline scaffold to compliance-benchmarks. All prior findings addressed.

**Re-review scope**: Incremental diff (`d5e01e9` vs `1c2563e`) — 87 lines added across 3 files. Verified each prior finding against the fix commit. CI is green.

---

### Prior review status

| Finding | Status |
|---|---|
| (1) Breaking change to `apply` input format | ✅ Fixed — backward compat via `cbPortableGroupsValid` + legacy SDK format fallback |
| (2) `ListBaselines` silently swallowed | ✅ Fixed — stderr warning added |
| (3) YAML encoder missing `Close()` | ✅ Fixed |
| (4) Duplicate group resolution in clone | ✅ Fixed — extracted `cbResolveNameList` helper |

### Detail on finding (1) fix

The backward compat implementation is well-designed:

1. Tries portable format first
2. Validates groups have non-empty names (catches YAML leniency producing empty-name structs from `[]string` input)
3. Falls back to the old `CBEngineBenchmarkRequestV2` struct, passing raw IDs through without resolution
4. `--computer-group` override works correctly in both formats

Two new tests cover the happy path (`TestCBApply_LegacyFormat`) and the override interaction (`TestCBApply_LegacyFormatWithGroupOverride`).

### No new issues from the fix commit

The incremental changes are clean and introduce no new concerns.

PR #123, opened 2026-04-12
feat: add --scaffold flag to all protect apply commands

### LINKED ISSUE #113: feat: add --scaffold flag to Protect apply commands
## Summary

Pro `apply` and `patch` commands have a `--scaffold` flag that prints an empty JSON template the user can fill in and pipe back. Protect `apply` commands have no equivalent.

## Proposed Behavior

Add `--scaffold` to every `protect <resource> apply` command. When passed, it should print an empty JSON (or YAML with `-o yaml`) template matching the SDK input type for that resource, then exit — without reading stdin or making any API calls.

Example:

```sh
bin/jamf-cli protect analytics apply --scaffold
bin/jamf-cli protect analytics apply --scaffold -o yaml
```

Output could then be piped directly to an editor or used as a starting point:

```sh
bin/jamf-cli protect analytics apply --scaffold | pbcopy
```

## Motivation

The workflow for Protect commands currently requires the user to know the input schema ahead of time or use `export` on an existing resource. `--scaffold` closes the gap for net-new resources.

## Implementation Notes

- The scaffold template should be generated from the SDK input struct (zero-value JSON marshal is the simplest approach).
- Should respect `-o yaml` (same as `export` commands).
- Pattern should match how `scaffoldJSON()` works in the Pro generator (`generator/parser/generator.go`).
- Needs to be added per-command in each `protect_*.go` file (or factored into a shared helper in `protect_helpers.go`).

neilmartin83:
## Summary

- Adds `--scaffold` to all 13 `protect <resource> apply` commands
- Prints a zero-value JSON template of the SDK input struct and exits — no auth, no API calls
- Respects `-o yaml` (uses `printExport`, same as export commands)
- `analytic-sets apply` and `plans apply` scaffold the export struct (the format `apply` actually accepts), not the raw SDK input

## Test plan

- [x] `bin/jamf-cli protect analytics apply --scaffold` prints JSON template
- [x] `bin/jamf-cli protect plans apply --scaffold -o yaml` prints YAML template
- [x] `bin/jamf-cli protect analytic-sets apply --scaffold` prints export struct
- [x] All other apply commands accept `--scaffold` without error

Closes #113


ktn-jamf (approved):
> **Merge-ready** — Adds `--scaffold` flag to all 13 `protect <resource> apply` commands, printing a zero-value JSON/YAML template without requiring auth. No critical or important issues found.

## Scope of review

- 14 files changed, +66/-1
- All 13 protect apply commands reviewed for pattern consistency
- Compared against existing platform `--scaffold` pattern (`pro_blueprints.go`, `pro_compliance_benchmarks.go`, `pro_platform_device_groups.go`)
- Verified auth bypass in `root.go` (`PersistentPreRunE` line 435) handles the new flags correctly
- Checked `printExport` vs `printScaffold` functional differences

## Findings

**Nice-to-have:**

🟩 (1) (consistency) — Flag help text says "JSON" but the feature supports YAML via `-o yaml`

All 13 new flags use `"Print an empty JSON template and exit"` as the description. Since `printExport` respects `-o yaml`, saying "JSON" is slightly inaccurate. A description like `"Print an empty template for the input format and exit"` would be more accurate. Cosmetic — not worth blocking.

🟩 (2) (testing) — No tests added for the new scaffold behavior

The existing test pattern in `pro_device_actions_test.go` shows how scaffold tests work. Adding similar tests for at least one protect apply scaffold would prevent regressions. Not blocking since the pattern is mechanical and the auth bypass is already tested implicitly.

## What's done well

- ✅ Smart choice to use `printExport` instead of `printScaffold` — gives protect scaffolds `-o yaml` support for free, a genuine UX improvement over the platform pattern.
- ✅ Correct use of export structs for `analytic-sets` and `plans` — scaffolds the format that `apply` actually accepts, not the raw SDK input type.
- ✅ The auth bypass in `root.go` line 435 already handles all new commands generically — no changes needed to the auth path.

PR #130, opened 2026-04-13
fix: preserve PayloadUUID and PayloadIdentifier on classic config profile updates

### LINKED ISSUE #129: Preserve PayloadUUID and PayloadIdentifier when updating Classic config profiles
## Problem

When updating a Classic API configuration profile (macOS or mobile), the CLI sends the new mobileconfig payload as-is without preserving the existing profile's `PayloadUUID` and `PayloadIdentifier`. This causes macOS/iOS devices to treat the updated profile as an entirely new installation rather than an update to the existing one, resulting in "ghost profiles" — the old profile lingers on devices even though the server considers it replaced.

This affects all three update code paths:
- **`upload`** (`internal/commands/pro_profiles_upload.go`) — takes a raw `.mobileconfig` file, wraps it in XML, and PUTs it
- **`apply`** (generated) — upserts full XML; when replacing an existing profile, sends the payload unchanged
- **`update`** (generated) — PUTs full XML with no payload modification

## Expected behavior

On every **update/replace** operation (not create), the CLI should:

1. Fetch the existing profile's `general/payloads` from the server (`GET /JSSResource/{resource}/id/{id}/subset/General`)
2. Parse the existing plist payload to extract the top-level `PayloadUUID` and `PayloadIdentifier`
3. Parse the new mobileconfig/payload plist
4. Replace the new payload's `PayloadUUID` and `PayloadIdentifier` with the existing values
5. Re-serialize and send the modified payload

This should happen unconditionally — there is no valid reason to change a profile's UUID during an update. If a user wants a fresh identity on devices, the correct workflow is delete + create.

## Affected resources

- `osxconfigurationprofiles` (classic-macos-config-profiles)
- `mobiledeviceconfigurationprofiles` (classic-mobile-config-profiles)

## Implementation notes

- Shared helper function, likely in `internal/profileconvert/`, that handles: fetch existing payloads → parse plist → extract UUIDs → inject into new plist → return modified bytes
- `upload` command calls this helper before building the XML envelope
- Classic generator template (`generator/classic/generator.go`) emits UUID preservation logic in the apply and update paths, gated to config profile resources only (the template already knows the resource type)
- Only the top-level `PayloadUUID` and `PayloadIdentifier` need preservation (nested payload-level UUIDs within `PayloadContent` are not affected)

## Reference

The [jamf-upload](https://github.com/grahampugh/jamf-upload) project implements this correctly:
- [JamfComputerProfileUploaderBase.py](https://github.com/grahampugh/jamf-upload/blob/main/JamfUploaderProcessors/JamfUploaderLib/JamfComputerProfileUploaderBase.py) — `get_existing_uuid_and_identifier()` + `replace_identifiers_in_mobileconfig()`
- [JamfMobileDeviceProfileUploaderBase.py](https://github.com/grahampugh/jamf-upload/blob/main/JamfUploaderProcessors/JamfUploaderLib/JamfMobileDeviceProfileUploaderBase.py) — same pattern

neilmartin83:
## Summary

- Adds `internal/profileconvert/classic_uuid.go` with shared helpers to fetch the existing profile's plist from the Classic API, extract the top-level `PayloadUUID` and `PayloadIdentifier`, and inject them into the incoming payload before any update/replace operation
- Updates the classic generator template (`generator/classic/generator.go`) so all generated `update` commands for config profile resources (`IsConfigProfile: true`) automatically call these helpers
- Updates `pro_profiles_upload.go` to call the same helpers on the upload-update path
- Generated files (`classic_macos_config_profiles.go`, `classic_mobile_config_profiles.go`, `classic_registry.go`) are regenerated to reflect the template changes

## Test plan

- [x] `make test` passes (including new `classic_uuid_test.go` unit tests for UUID extraction and injection)
- [x] `make generate` produces no diff against the committed generated files
- [x] Updating a macOS config profile via `upload`, `update`, and `apply` preserves `PayloadUUID`/`PayloadIdentifier` — profile is treated as an update on devices, no ghost profiles

Closes #129

### DIFF
--- a/generator/classic/generator.go
+++ b/generator/classic/generator.go
@@ -722,6 +722,10 @@ import (
 	"os"
 	"strings"
 {{- end }}
+{{- if anyIsConfigProfile . }}
+	"bytes"
+	"net/url"
+{{- end }}
 
 	"github.com/spf13/cobra"
 
@@ -730,9 +734,6 @@ import (
 	"github.com/Jamf-Concepts/jamf-cli/internal/xmlconv"
 {{- end }}
 {{- if anyIsConfigProfile . }}
-	"bytes"
-	"net/url"
-
 	"github.com/Jamf-Concepts/jamf-cli/internal/profileconvert"
 {{- end }}
 )
@@ -863,6 +864,9 @@ func resolveClassicNameToIDForApply(ctx context.Context, client registry.HTTPCli
 // fetchClassicProfileByName fetches a Classic config profile by name and returns
 // its numeric ID and existing payload plist in a single API call. Returns ("", nil)
 // when not found; callers should proceed without UUID injection if payload is nil.
+//
+// Errors (including non-404 server errors) are silently swallowed — UUID
+// preservation is best-effort and must never block an update.
 func fetchClassicProfileByName(ctx context.Context, client registry.HTTPClient, apiPath, name string) (id string, payloadPlist []byte) {
 	path := fmt.Sprintf("/JSSResource/%s/name/%s", apiPath, url.PathEscape(name))
 	resp, err := client.Do(ctx, "GET", path, nil)
--- a/generator/classic/generator.go
+++ b/generator/classic/generator.go
@@ -155,7 +155,16 @@ func templateFuncs() template.FuncMap {
 			return r.HasOperation("create") && r.HasOperation("update") && r.HasLookup("name")
 		},
 		"needsBytes": func(r ClassicResource) bool {
-			return r.HasOperation("create") && r.HasOperation("update") && r.HasLookup("name")
+			return (r.HasOperation("create") && r.HasOperation("update") && r.HasLookup("name")) ||
+				(r.IsConfigProfile && r.HasOperation("update"))
+		},
+		"anyIsConfigProfile": func(resources []ClassicResource) bool {
+			for _, r := range resources {
+				if r.IsConfigProfile {
+					return true
+				}
+			}
+			return false
 		},
 		"hasDeleteByName": func(r ClassicResource) bool {
 			return r.HasOperation("delete") && r.HasLookup("name")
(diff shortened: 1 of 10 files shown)


ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Preserves `PayloadUUID` and `PayloadIdentifier` during Classic config profile updates/applies/uploads, preventing ghost profiles on managed devices. CI green, well-tested, clean architecture.

## Scope of review

- **764 additions, 96 deletions** across 10 files
- Reviewed: generator template changes, generated output, new `profileconvert` package, `pro_profiles_upload.go` changes, scaffolds.go diff
- Verified: template/generated consistency, CDATA wrapping logic, error handling strategy, test coverage
- CI: passing
- Prior reviews: none

## Findings

💡 **(1)** (scope) — `internal/blueprintcomponents/scaffolds.go`: scaffold default changes are unrelated

The `scaffolds.go` changes flip ~40 boolean scaffold defaults from `true` to `false` across DDM settings (math, passcode, safari, software-update). This is a separate behavioral change — scaffold output for `pro blueprints apply --scaffold` will now produce different defaults. Not harmful, but muddies the commit history for the UUID fix.

**Suggested fix:** Split into a separate commit with its own message (e.g., `fix(scaffolds): use restrictive defaults for DDM component scaffolds`). Fine as a follow-up.

**Fixed when:** Scaffolds change is in its own commit with a descriptive message.

---

💡 **(2)** (dead code) — `internal/profileconvert/classic_uuid.go:65`: `ExtractProfileIdentifiers` is exported but unused

`ExtractProfileIdentifiers` is tested but never called outside tests. If intended for future use, that's fine — but it's currently dead code.

**Suggested fix:** Either remove it or add a brief comment noting its intended use. No urgency.

**Fixed when:** Function is either removed or has a caller, or a comment explains its purpose.

---

## What's done well

✅ **Best-effort error strategy** — UUID injection never blocks updates. Every error path returns the input unchanged rather than failing the command. Right call: preserving UUIDs is an optimization, not failing to preserve them shouldn't prevent the update.

✅ **Single-call optimization** — `fetchClassicProfileByName` returns both ID and existing payload in one API call, avoiding an extra round-trip. `lookupClassicProfileByName` in upload was similarly updated.

✅ **Thorough test coverage** — `classic_uuid_test.go` covers happy path, empty/nil existing, invalid existing, no UUIDs in existing, invalid new plist, and content preservation with nested `PayloadContent` arrays.

✅ **Clean template integration** — `IsConfigProfile` is derived from the path string in the parser, keeping YAML schema unchanged. Template guards are well-scoped with `{{ else }}` fallback preserving original behavior for non-profile resources.

PR #190, opened 2026-05-08
feat(output): add --select flag for multi-field projection

ktn-jamf:
## Summary

- Adds `--select id,general.name,udid` for jq-style multi-field projection on any command
- Supports prefix-matching: `--select general` returns every `general.*` field after flattening
- Builds on the `output.Projector` scaffold from #189 (compact)
- When both `--select` and `--compact` are set, `--select` wins

## Why

`--field` extracts a single scalar (great for shell pipelines: `--field id | xargs ...`). `--select` is the JSON-shaped sibling: keep the structure, drop everything you don't need. Together they cover the two ways agents narrow Jamf Pro's enormous responses.

## Examples

```bash
# Single device with just id, name, serial
jamf-cli pro computers get --name "MacBook Pro" --select id,general.name,general.serial_number

# List everything in a record's general section
jamf-cli pro computers get --name "MacBook Pro" --select general

# Combine: --select wins over --compact
jamf-cli pro computers list --compact --select profiles  # returns profiles arrays
```

## Test plan

- [x] `go test ./internal/output/...` — 7 new test cases for select (exact, prefix, dot-path, missing, whitespace, select+compact)
- [x] `go test ./...` — full suite green
- [x] `make lint` — 0 issues
- [x] `bin/jamf-cli --help` shows `--select` line

## Stacking

Based on `feat/compact-flag` (#189). Once that lands, this PR's diff will only show the select-specific additions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #192, opened 2026-05-08
feat(commands): add doctor diagnostic command

ktn-jamf:
## Summary

- New `jamf-cli doctor [profile]` command — prints config path, resolved active profile (and the signal that selected it), every JAMF_* / JAMFPROTECT_* / JAMFSCHOOL_* env var (secrets fingerprinted, never full values), per-profile credential resolution status, and a 5-second unauthenticated HEAD probe of the configured URL
- Inspired by [cli-printing-press](https://github.com/mvanhorn/cli-printing-press)'s `auth doctor`
- Skips auth (added to `chainSkip`) so it works precisely when credentials are misconfigured — the case where it's most useful

## Why

When an agent or human hits 401, the question is "is the token missing, expired, or shadowed by a stale env var?". Today that takes `cat ~/.config/jamf-cli/config.yaml`, `env | grep JAMF`, and a curl. `jamf-cli doctor` is one command.

## Example output (`-o table` for the human view)

```
jamf-cli v1.16.1

CONFIG
  path:    /Users/keaton/.config/jamf-cli/config.yaml
  status:  present

ACTIVE PROFILE
  name:        my-pro
  source:      JAMF_PROFILE env var
  product:     pro
  url:         https://example.jamfcloud.com
  auth-method: oauth2
  client-id:   env:JAMF_CLIENT_ID  (resolved, fingerprint: abcd••••)
  client-secret: keychain:jamf-cli/my-pro-secret  (resolved, fingerprint: wxyz••••)

ENVIRONMENT
  JAMF_URL                       (unset)
  JAMF_PROFILE                   my-pro
  JAMF_TOKEN                     (unset)
  JAMF_CLIENT_ID                 set, fingerprint: abcd••••
  ...

CONNECTIVITY
  HEAD https://example.jamfcloud.com  →  200 OK (148ms)
```

JSON output via `-o json` (the default) returns the same data as a structured `doctorReport`.

## Security

- Secrets are never printed in full. `fingerprint()` returns first 4 chars + `••••`. Values shorter than 4 chars are fully redacted.
- The connectivity probe is HEAD-only, no auth headers, 5s timeout — pure DNS/TLS reachability.
- Reads `~/.config/jamf-cli/config.yaml` and OS env vars only. No writes.

## Test plan

- [x] `go test ./internal/commands/... -run 'Doctor|Fingerprint|ProbeEnvVars|ProbeProfileCredentials|ResolveProfileNameForDoctor'` — 11 tests covering precedence, fingerprint redaction, env-var probing, missing-profile note, credential resolution success/failure
- [x] `go test ./...` — full suite green (including `TestApplyRootGroups_AllCommandsGrouped` after adding `doctor` to `rootGroupMap`)
- [x] `make lint` — 0 issues
- [x] Manual: `jamf-cli doctor`, `jamf-cli doctor -o table`, `jamf-cli doctor missing-profile`

## Notes

This is the fourth and last "cheap win" pulled from cli-printing-press. Independent of #189 / #190 / #191 — can land in any order.

### DIFF
--- a/internal/commands/doctor.go
+++ b/internal/commands/doctor.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"sort"
@@ -59,12 +60,14 @@ type doctorReport struct {
 }
 
 type profileReport struct {
-	Name        string         `json:"name"`
-	Source      string         `json:"source"`
-	Product     string         `json:"product,omitempty"`
-	URL         string         `json:"url"`
-	AuthMethod  string         `json:"authMethod"`
-	Credentials []credentialOK `json:"credentials,omitempty"`
+	Name         string         `json:"name"`
+	Source       string         `json:"source"`
+	Product      string         `json:"product,omitempty"`
+	URL          string         `json:"url"`
+	EffectiveURL string         `json:"effectiveUrl,omitempty"` // URL after --url/JAMF_URL override
+	URLSource    string         `json:"urlSource,omitempty"`    // "profile", "JAMF_URL env", "--url flag"
+	AuthMethod   string         `json:"authMethod"`
+	Credentials  []credentialOK `json:"credentials,omitempty"`
 }
 
 type credentialOK struct {
@@ -73,6 +76,9 @@ type credentialOK struct {
 	Resolved    bool   `json:"resolved"`
 	Fingerprint string `json:"fingerprint,omitempty"`
 	Error       string `json:"error,omitempty"`
+	// EnvOverride names the env var that, if set, would shadow this
+	// profile credential at auth time. Empty when nothing shadows it.
+	EnvOverride string `json:"envOverride,omitempty"`
 }
 
 type envEntry struct {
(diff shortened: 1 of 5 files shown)


neilmartin83:
Approved — live tested against `platform-nmartin` tenant.

**Verified:**
- `doctor [profile]` resolves correctly across all three sources (positional arg, `JAMF_PROFILE` env, config default-profile) with `source` field reflecting which won
- Fingerprint redaction: 4-char prefix + `••••` for keychain-resolved client-id/client-secret, full bullet-redact for env values <4 chars (verified with `JAMF_TOKEN=abc` → `•••`)
- Missing profile: emits actionable `notes` entry, skips connectivity, doesn't crash
- Connectivity probe is HEAD-only, no `Authorization` header, 5s timeout, redirects not followed — confirmed in source at `internal/commands/doctor.go:305`
- Skips auth chain (works precisely when creds are broken — exactly when needed)
- Plays nice with PR 189/190/191: `--compact`, `--select profile.name,connectivity`, `--field version`, `-o table` all behave as expected
- `go test ./internal/commands/... -run 'Doctor|Fingerprint|Probe...'` green

**Gap (non-blocking, file follow-up):**
`profileReport` has no `tenant-id` field. For `auth-method: platform` profiles the table omits it entirely, even though the env section lists `JAMF_TENANT_ID`. Anyone diagnosing "why does my platform auth 401" will miss a missing/wrong tenant-id. Add `TenantID string` to `profileReport`, populate from `p.TenantID` (already on `config.Profile` at `config.go:32`), render conditionally in the table block. Low risk, high signal for platform users.

Same for `product` — field exists in struct (`doctor.go:65`) and rendering branch exists at `doctor.go:418`, but profiles don't carry `Product` for platform-auth so it never prints. Worth confirming intended behavior.

Ship — these are polish, not blockers.

🤖 Live-tested with [Claude Code](https://claude.com/claude-code)

neilmartin83 (approved):
Live-tested against platform-nmartin tenant. Profile resolution precedence, fingerprint redaction, missing-profile handling, HEAD-only unauthenticated probe, --compact/--select/--field interop, and unit tests all clean. Filed nit on missing tenant-id in profileReport (platform-auth gap) — non-blocking.

PR #203, opened 2026-05-11
feat(classic): add ebooks, user-groups, vpp scope support

neilmartin83:
## Summary

- Adds `ebooks` to `specs/classic/resources.yaml` — full CRUD, `id`/`name` lookups, scope support; generates `classic_ebooks.go`
- Adds `usergroups` to `specs/classic/resources.yaml` — full CRUD + `apply`; generates `classic_user_groups.go`
- Adds scope support to `vppassignments` and `vppinvitations` (both require full-doc PUT — no `/subset/Scope` endpoint)
- Fixes `NamedItem.ID` and `classicResourceXML.General.ID` typed as `int` — Classic API returns UUID strings for scope item IDs; changed both to `string`
- Adds `--user-group`, `--jss-user-group`, `--jss-user` as valid scope target/exclusion flags, routing to `jss_user_groups`/`jss_users` in scope XML (VPP-style user-based targeting)
- Adds `ResolveByList` to `scope.Resource` — resources without a `/name/` endpoint list all records and match by name
- Adds `NoSubsetPut` to `scope.Resource` — resources without a `/subset/Scope` PUT endpoint fetch the full document, splice in the updated scope at byte level, and PUT the whole document back
- Wires both flags through YAML manifest → `ClassicResource` → generator template → generated `scope.Resource` literals

## Test plan

- [x] `make test` passes
- [x] `bin/jamf-cli pro classic-ebooks scope get "<name with colon>"` returns scope without parse error
- [x] `bin/jamf-cli pro classic-user-groups list` returns JSS user group records
- [x] `bin/jamf-cli pro classic-vpp-assignments scope get "<name>"` resolves name via list, returns scope
- [x] `bin/jamf-cli pro classic-vpp-assignments scope add "<name>" --user-group "<group>"` adds to `jss_user_groups` target
- [x] `bin/jamf-cli pro classic-vpp-assignments scope remove "<name>" --user-group "<group>"` removes from `jss_user_groups` target
- [x] `bin/jamf-cli pro classic-vpp-invitations scope get/add/remove` all work (live-tested against platform-nmartin ✓)


ktn-jamf (commented):
## Summary

Nice work pulling ebooks + user-groups in and wiring VPP assignment scope through the new `ResolveByList` / `NoSubsetPut` knobs — the generator-template approach keeps the surface area small. Two things I'd want addressed before merge, plus a handful of smaller observations.

**Verdict:** Needs changes.

---

### Important

**1. `--user-group` / `--jss-user-group` / `--jss-user` are accepted as a target for every scope-enabled resource, not just VPP assignments.** `ValidateScopeCombination` in `internal/scope/scope.go` (and the matching test on line 423) widens the allowed `target` flag set unconditionally:

```go
case "computer-group", "mobile-device-group", "building", "department",
    "user-group", "jss-user-group", "jss-user":
    return nil
```

The Classic API only honours user-group / jss-user / jss-user-group as targets for VPP assignments — for policies and config profiles those belong in `limitations` (which the previous validation already enforced). After this PR, `jamf-cli pro policies scope add --target --user-group ...` will validate successfully, hit the API, and silently no-op (or end up in a section the user didn't intend). Recommend gating the new target entries on \`singularKey == "vpp_assignment"\` (and any other resource you've confirmed supports them) and updating the error message + tests to match. Worth a unit test that the old rejection still fires for \`policy\` / \`os_x_configuration_profile\`.

**2. No HTTP-orchestration coverage for the new `ResolveByList` and `NoSubsetPut` code paths.** `internal/scope/scope_test.go` exercises `replaceScopeInXML` directly (good) but `FetchScope` with `ResolveByList=true` and `PutScope` with `NoSubsetPut=true` — including `resolveNameToID` and `putFullDocument` — have no tests. These are the highest-risk additions because they encode new assumptions about Classic XML shape (depth-2 list elements, `<scope>...</scope>` byte-search splice). A small `httptest`-driven test that wires through a fake `registry.HTTPClient` and asserts the GET→GET→PUT sequence would catch regressions cheaply.

---

### Nice-to-have

- **`resolveElemName(section, flagName)` ignores its `section` arg.** It's a one-line wrapper around `flagToElemName[flagName]` — either drop the unused parameter, or actually use `section` to disambiguate (which would also be a cleaner way to handle the `--user-group` overload than relying on map collision behaviour). The doc comment above it ("Both --user-group and --jss-user-group route to jss_user_groups") describes a map-driven behaviour that's now load-bearing for VPP scope correctness — worth a test that pins it.
- **`resolveNameToID` reads up to 10 MiB then silently truncates** via `io.LimitReader`. For Classic list endpoints this is almost always fine, but if a tenant exceeds it the function will return "not found" instead of an error. Consider returning a descriptive error when the limit is hit, or at minimum log it.
- **`replaceScopeInXML` uses raw byte search for `<scope>` / `</scope>`.** The comment says Classic API XML is well-formed so this is safe, but a name/description field containing the literal text `<scope>` (escaped or otherwise) would break it. Low likelihood, but at minimum, assert that `</scope>` appears exactly once after the open tag and fail loudly if not.
- **`DetermineScopeTarget` error message lists `--jss-user-group` but not `--jss-user` as a target option** (the validation list does include it). One-liner consistency fix.

---

### Other observations

- CI is green and `internal/scope/...` tests pass locally.
- Templates touching `scopeResolveByList` and `NoSubsetPut` look right; the `make verify-generated` gate should keep that honest.
- Diff is large (1418/-36) but the bulk is the two new generated resource files — the hand-written delta is manageable.

Once the target-gating is scoped to VPP and there's at least one HTTP-orchestration test exercising `ResolveByList` + `NoSubsetPut`, I'm happy to re-look.

neilmartin83:
@ktn-jamf thanks for the review. I spent some time live-testing the full scope-acceptance matrix against a real Jamf Pro tenant (every section × every flag × every scope-enabled resource type) and the picture is more nuanced than either of us had. Pushed `a245c67` to address your feedback plus a few related issues the testing turned up.

## What the live test changed about my mental model

Cross-checked with `pro-nmartin` (direct oauth2, full permissions) and a fresh test resource for each section/type pair, then a GET after PUT to confirm what actually persisted. Three corrections vs. my original assumptions and your review:

- **`jss_user_groups` target is not VPP-only.** Server happily persists `<jss_user_groups>` as a target on policies, both config-profile types, mac-apps, mobile-apps, and ebooks (verified by GET after subset/Scope PUT). The "silently no-op" framing only fits restricted_software, which truly drops it.
- **`jss-user` works on every scope-enabled resource except restricted_software**, but only when the name resolves to a Jamf inventory user (`/JSSResource/users`), not an account user. My initial 409s were testing with `demuser` (an account-user) — once I tried a real Jamf user (`username`, id 10), it persisted everywhere.
- **The real problem with `--user-group` is the section-dependent semantic split**, not the silent no-op. Same flag value lands in `<jss_user_groups>` for target but `<user_groups>` for limitation/exclusion. Different XML location, different namespace (JSS vs LDAP), no way for the user to tell from the flag.

## Changes in `a245c67`

**Validation (your point #1):**
- Drop `--user-group` from target validation entirely. `--jss-user-group` is the explicit way to target a JSS user group; `--user-group` now only means LDAP/directory group (limitation/exclusion).
- Restricted-software target now gated to `--computer`, `--computer-group`, `--building`, `--department` (mirrors the existing limitation/exclusion gate). Previously the target list accepted everything universally.

**Newly discovered: restricted_software has no `/subset/Scope` endpoint.** Returns HTTP 404. Pre-existing in main — `scope add` on restricted-software was always broken on every tenant. Set `no_subset_put: true` in `specs/classic/resources.yaml`, so it now falls back to a full-document PUT and the existing scope.go full-doc path takes over. Verified end-to-end on the live tenant.

**Data-loss bug (this one surprised me):** `ScopeXML` had no `MobileDevices` field, no `Classes` field, no `ExclusionsXML.MobileDevices` field. So when the CLI fetched scope, unmarshal silently dropped any individual `<mobile_devices>` / `<classes>` the server returned; marshal sent empty back; the subset/Scope PUT then wiped them. Reproduced on a mobile config profile with one individual `<mobile_device>` set manually — single `scope add --mobile-device-group "All Managed iPads"` deleted it. Added the missing fields so unmarshal/marshal round-trips preserve them even though no CLI flag mutates them.

**Your point #2 (post-PUT verification):** added `VerifyItemInScope`. After every PUT, the CLI now re-fetches scope and asserts the item is (or isn't) present, with policy `limit_to_users`/`limitations.user_groups` dual-mirror handling. Catches silent server drops as a clear error:

```
$ jamf-cli pro classic-vpp-assignments scope add my-assignment --section target --building "HQ"
Error: server accepted PUT but did not persist --building "HQ" in target scope
       (resource type "vpp_assignment" does not support this scope element)
```

Replaces the old false-success path where the CLI printed "Added X" and the server had quietly dropped the field.

**New flags filling the Classic API coverage gap.** Server-side scope accepts individual `<computers>`, `<mobile_devices>`, and free-text `<users>` (inventory user names) but the CLI had no flags. Added:
- `--computer <id|name|udid>` — target/exclusion. Server resolves by any identifier.
- `--mobile-device <id|name|udid>` — target/exclusion.
- `--user <name>` — limitation/exclusion, free-text inventory username.

**Tests:** updated validation matrix tests, added round-trip preservation tests for `<mobile_devices>` and `<classes>`, added per-flag tests for the new flags. The HTTP-orchestration test you suggested for `ResolveByList`/`NoSubsetPut` I haven't added yet — happy to do that as a follow-up if you want it bundled into this PR, otherwise it'll come in a separate cleanup.

## Smaller items from your review

- **`resolveElemName` ignoring `section`** — kept the single-arg map lookup; added a doc comment on `flagToElemName` explaining why `--user-group` and `--jss-user-group` share `<user_group>` (the parent element disambiguates). With `--user-group` dropped from target, the collision is no longer load-bearing.
- **`resolveNameToID` 10MiB limit** — left as-is for this PR; happy to follow up with a clear error on truncation.
- **`replaceScopeInXML` byte search** — left as-is; the new `VerifyItemInScope` catches the bad-splice case anyway by checking the post-PUT state.
- **`DetermineScopeTarget` error message** — updated to include all valid flags including the new ones.

## Still open (out of scope for this PR)

- `classic-ebooks delete` returns HTTP 400 with the entity body. Reproduces on direct auth and platform gateway. Server bug. Will file separately.
- `platform-nmartin` (platform gateway) profile gets 403 on every `/subset/Scope` PUT regardless of resource. Either a role-config issue or a gateway endpoint filter. Will investigate separately — doesn't block this PR since `pro-nmartin` works.
- Coverage: `<ibeacons>` as limitation/exclusion still has no CLI flag. Real but low-traffic. Follow-up.

Ready for another look when you have a moment.

ktn-jamf:
> [!WARNING]
> ⚠️ **Needs minor changes** — Adds ebooks + user-groups Classic resources, VPP scope, tightens scope validation, and adds post-PUT verification.
> Blocking: (1). See collapsed sections for nice-to-haves and 2 still-open items from prior review.

### Prior findings status

| # | Finding (round 1) | Status |
|---|---|---|
| 1 | `--user-group` / `--jss-user-group` / `--jss-user` widened as targets unconditionally | ✅ **Fixed.** `--user-group` dropped from target entirely; `--jss-user-group`/`--jss-user` kept as the explicit target flags. Restricted software target now gated to `computer{,-group}`/building/department. New tests cover the regression. |
| 2 | No HTTP-orchestration coverage for `ResolveByList` / `NoSubsetPut` | 🟧 **Not addressed.** Acknowledged in author's response as a follow-up. The new XML-level tests are good, but the GET-list→GET-by-id→PUT (subset or full-doc) wire sequence is still untested. |
| 3 | `resolveElemName(section, flagName)` ignores `section` | 🟧 **Partially addressed.** Signature kept, doc comment added — but the comment is now stale (see new finding 3). |
| 4 | `resolveNameToID` 10 MiB `io.LimitReader` truncates silently | 🟧 Acknowledged, deferred to a follow-up. |
| 5 | `replaceScopeInXML` raw byte search for `<scope>` tags | 🟧 Acknowledged, deferred; risk reduced by the new `VerifyItemInScope` post-PUT check. |
| 6 | `DetermineScopeTarget` error message missing `--jss-user` | ✅ **Fixed.** Updated to list every valid flag including new ones. |

### Important

🟧 **(1) (correctness) — `internal/scope/scope.go:233`: `VerifyItemInScope` falsely reports silent-drop when user passes id or UDID for `--computer` / `--mobile-device`**

The new individual-device flags accept three identifier shapes (the help text says so verbatim):

```go
cmd.Flags().String("computer", "", "individual computer (id, name, or UDID); target/exclusion only")
cmd.Flags().String("mobile-device", "", "individual mobile device (id, name, or UDID); target/exclusion only")
```

`AddToScope` stores the raw user value in `NamedItem{Name: name}`. The server resolves the identifier on PUT and the post-PUT GET returns the canonical name in `<name>` (plus the id/udid in sibling elements). `VerifyItemInScope` then compares the user's input against `item.Name` only:

```go
for _, item := range items.Items {
    if strings.EqualFold(item.Name, itemName) {
        present = true
        break
    }
}
```

So `--computer 10` (id) or `--computer 00008101-000170490151003A` (UDID) round-trips successfully on the server but the verify step trips, the user sees `server accepted PUT but did not persist --computer "10" in target scope (resource type "policy" does not support this scope element)`, and the command exits with error even though the scope was correctly updated. Worse, the second sentence of the error message blames the resource type — actively misleading.

The matching `<id>`/`<udid>` fields are already available in `NamedItem` (verified by reading the diff in `scope_test.go:520`: `MobileDevice.ID = "18"`, `UDID = "00008101-..."`).

**Suggested fix:**
```diff
 for _, item := range items.Items {
-    if strings.EqualFold(item.Name, itemName) {
+    if strings.EqualFold(item.Name, itemName) ||
+        item.ID == itemName ||
+        strings.EqualFold(item.UDID, itemName) {
         present = true
         break
     }
 }
```

Apply the same triple-match in the `isPolicyLimitUserGroup` branch only if user groups also accept id-form input (probably not — keep that branch name-only).

**Fixed when:** `bin/jamf-cli pro policies scope add "<policy>" --computer 10` and `bin/jamf-cli pro mobile-device-configuration-profiles scope add "<profile>" --mobile-device 00008101-...` both succeed and report `Added --computer "10" to target scope of "<policy>"` without a verify error. Unit test added to `scope_test.go` that constructs a `ScopeXML` with `Computers.Items = []NamedItem{{ID: "10", Name: "Mac-Build-01", UDID: "AAA-BBB"}}` and asserts `VerifyItemInScope` returns nil for any of `"10"`, `"Mac-Build-01"`, `"aaa-bbb"` with `expectPresent=true`.

<details>

🟩 **(2) (correctness/UX, deferred from round 1) — Still no HTTP-orchestration tests for `ResolveByList` / `NoSubsetPut`**

The new XML round-trip tests at `scope_test.go:803-846` are valuable (and caught the data-loss bug, nice work), but the wire-level sequences added in this PR remain untested:

- `FetchScope` with `ResolveByList=true`: list endpoint → name match → GET by id
- `PutScope` with `NoSubsetPut=true`: full-doc GET → splice → full-doc PUT
- `VerifyItemInScope` itself: re-fetch → membership check

A single `httptest.NewServer` wired through a fake `registry.HTTPClient` covering one end-to-end add against `vpp_assignment` (ResolveByList + NoSubsetPut + Verify) would lock in all three new paths cheaply. Author acknowledged this in the PR-comment response as a follow-up — fine to defer, but worth filing a tracking issue rather than letting it drift.

**Fixed when:** a tracking issue exists, OR `internal/scope/scope_test.go` contains at least one test that drives `FetchScope`/`PutScope`/`VerifyItemInScope` through a fake `HTTPClient` and asserts the GET-list→GET-by-id→PUT-full-doc→GET sequence.

🟩 **(3) (code-quality) — `internal/scope/scope.go:644-650`: `resolveElemName` doc comment is stale**

The comment still claims `--user-group` and `--jss-user-group` collide in `jss_user_groups`:

```go
// resolveElemName returns the XML child element name for a new scope item.
// Both --user-group and --jss-user-group route to jss_user_groups whose
// children are <user_group>, so flagToElemName["user-group"] = "user_group"
// is correct for both cases.
func resolveElemName(section, flagName string) string {
    return flagToElemName[flagName]
}
```

After this PR `--user-group` is rejected as a target — it only routes to `limitations/user_groups` (LDAP) or `exclusions/user_groups`. The collision is no longer load-bearing. Update or drop the comment so the next reader doesn't add code on a wrong premise. Either drop the unused `section` parameter, or use `section` to make the dispatch explicit instead of relying on the (now-unambiguous) flag→elem map.

**Fixed when:** the doc comment matches the post-PR validation rules, or the `section` parameter is used/removed.

🟩 **(4) (UX) — `internal/scope/scope.go:245-249`: `silentDropError` hard-codes a single failure cause**

```go
return fmt.Errorf("server accepted PUT but did not persist --%s %q in %s scope (resource type %q does not support this scope element)", flagName, itemName, section, singularKey)
```

With the new free-text identifier flags (`--computer 10`, `--user alice`), the more likely failure mode is "identifier doesn't resolve to a real entity" — at which point Jamf either drops the entry or stores a stub the GET filters out. The current wording blames the resource type and points the user at "this scope element isn't supported," which sends them down the wrong debugging path.

**Suggested fix:** soften the parenthetical to list both plausible causes — e.g. `"identifier did not resolve to an existing record, or resource type %q does not support this scope element"`.

**Fixed when:** the error message no longer asserts a single cause when two are equally likely.

</details>

<details>

- ✅ **The `ScopeXML` round-trip data-loss fix is the most important change in this push.** Catching that `<mobile_devices>` / `<classes>` / `<exclusions><mobile_devices>` were being silently wiped on every subset/Scope PUT — and reproducing it against a real tenant — is exactly the kind of bug that hides for years. The unconditional-presence convention with the explanatory doc comment in `types.go:147-151` is the right durable fix.
- ✅ **The `restricted_software` `no_subset_put: true` discovery is a pre-existing bug fix the PR description didn't oversell.** Quietly mending a "scope add was always broken on every tenant" path while shipping is good citizenship.
- ✅ **`VerifyItemInScope` correctly handles the policy `limit_to_users` ↔ `limitations/user_groups` dual-mirror.** That's the one place where a naive present/not-present check would have lied; nice catch.
- ✅ **Validation matrix tests** (`TestValidateScopeCombination_*` series) now cover the rejection paths, not just the accept paths. Future regressions will fail loudly.

</details>

<details>

- Incremental diff `47bdbe8..a245c67`: 363 insertions / 45 deletions across 7 files
- Read in full: `internal/scope/scope.go`, `internal/scope/scope_test.go`, `internal/scope/types.go`, `internal/scope/commands.go`, the `restricted_software` generator change
- Did not re-review: ebooks/usergroups generator output (covered in round 1), provenance.go SHA bump (mechanical)
- Verified locally: `go test ./internal/scope/...` passes; `make verify-generated` clean
- CI: `ci` ✅ and `preflight` ✅ on commit `a245c67`

</details>

---

Generated with Claude Code

neilmartin83:
**Round 2 addressed.**

- **(1) VerifyItemInScope id/UDID false-positive** — Fixed in `c08d06c`. Added `UDID` field to `NamedItem` (captured from `<udid>` in Classic XML). `VerifyItemInScope` now triple-matches: name, numeric id, or UDID (case-insensitive). `silentDropError` message updated to list both failure causes. Test `TestVerifyItemInScope_MatchesByNameIDAndUDID` covers all three identifier forms for `--computer` and `--mobile-device`.
- **Stale `resolveElemName` comment** — Dropped.

ktn-jamf (approved):

PR #208, opened 2026-05-14
fix(platform): apply commands fail to create new resources

### LINKED ISSUE #207: blueprints apply fails to create new blueprints — errors on empty search result instead of creating
## Summary

`jamf-cli pro blueprints apply` errors when the named blueprint does not yet exist, instead of creating it. The command documentation says "Create or update a blueprint" but in practice it only updates.

## Steps to reproduce

```bash
printf '{"name":"Brand New Blueprint","scope":{"deviceGroups":["<uuid>"]},"steps":[]}' > /tmp/bp.json
jamf-cli -p <platform-profile> pro blueprints apply --from-file /tmp/bp.json --yes
```

## Observed behaviour

```
{
  "error": "general",
  "exitCode": 1,
  "message": "ResolveBlueprintIDByName(Brand New Blueprint): API request failed with status 404 Not Found (method=GET, url=/api/blueprints/v1/tenant/<tenant-id>/blueprints): no resource found with name == \"Brand New Blueprint\""
}
```

With `-vvv`, only one HTTP request is made — `GET /blueprints?search=Brand+New+Blueprint` — which returns `200 OK` with `{"results":[],"totalCount":0}`. The CLI then generates a synthetic 404 from the empty results and exits rather than proceeding to `POST` the new blueprint.

## Expected behaviour

When the search returns zero results, `apply` should fall through to create the blueprint (POST), consistent with the "create or update" contract.

## Workaround

Create new blueprints via the REST API directly:

```bash
TOKEN=$(jamf-cli -p <profile> platform auth token --field token)
curl -X POST "https://<gateway>/api/blueprints/v1/tenant/<tenant-id>/blueprints" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"name":"Brand New Blueprint","scope":{"deviceGroups":["<uuid>"]},"steps":[]}'
```

## Version

```
jamf-cli version: $(jamf-cli version 2>&1 | head -1)
```

neilmartin83:
## Summary

- `platform.IsNotFound` matched only the `ErrNotFound` sentinel via `errors.Is`; SDK `Resolve*` methods return `*APIResponseError{404}` on empty search results, which never matched
- `blueprints apply` always errored on new names — resolver returned 404, `IsNotFound` returned false, command exited instead of creating
- `platform-device-groups apply` had the inverse problem: no `IsNotFound` guard at all, so any error (500s, auth failures) fell through to a create attempt

## Fix

- `platform.IsNotFound` now also matches `*APIResponseError` with status 404 (via `jamfplatform.AsAPIError`)
- Added the missing `IsNotFound` guard to `platform-device-groups apply` to match the blueprint pattern
- Extended `TestIsNotFound` to cover SDK 404, wrapped SDK 404, SDK 500 (must not match), sentinel, nil

## Test

Live-tested on `platform-nmartin` for both commands:
- New name → `GET` (empty) → `POST` → 201 ✓
- Existing name → `GET` (found) → `PATCH`/`PUT` → 204/200 ✓

Closes #207

### DIFF
--- a/internal/commands/pro_platform_device_groups.go
+++ b/internal/commands/pro_platform_device_groups.go
@@ -67,6 +67,9 @@ func newPDGApplyCmd(cliCtx *registry.CLIContext) *cobra.Command {
 
 			dg := devicegroups.New(cliCtx.PlatformSDKClient)
 			id, resolveErr := dg.ResolveDeviceGroupIDByName(ctx, createReq.Name)
+			if resolveErr != nil && !platform.IsNotFound(resolveErr) {
+				return resolveErr
+			}
 			if resolveErr != nil {
 				// Not found — create
 				result, err := devicegroups.New(cliCtx.PlatformSDKClient).CreateDeviceGroup(ctx, &createReq)
--- a/internal/platform/resolve.go
+++ b/internal/platform/resolve.go
@@ -2,7 +2,11 @@
 
 package platform
 
-import "errors"
+import (
+	"errors"
+
+	"github.com/Jamf-Concepts/jamfplatform-go-sdk/jamfplatform"
+)
 
 // ErrNotFound is returned when a resource name cannot be resolved to an ID.
 //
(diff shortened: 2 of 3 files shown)


ktn-jamf (approved):
LGTM

PR #218, opened 2026-05-21
feat(scope): resolve individual mobile devices and computers by UDID or numeric ID

### LINKED ISSUE #217: scope add should support targeting individual mobile devices, not just groups
The scope add command for Classic API resources that support scoping (e.g. classic-mobile-config-profiles) currently only accepts group-level targets via --mobile-device-group. There is no way to scope a profile to a specific device by serial, UDID, or device ID.

**Desired behavior**

jamf-cli pro classic-mobile-config-profiles scope add "Profile Name" \
  --section target \
  --mobile-device <id-or-udid> \
  --no-input

**Why**

There are legitimate workflows where a config profile needs to be created and immediately scoped to one specific device. Pre-staging groups for every possible target is not practical when the target device is determined dynamically at runtime.

**API support**

The Classic API already supports this. The scope object on mobiledeviceconfigurationprofiles includes a mobile_devices array that accepts individual devices by id, name, udid, or wifi_mac_address:

```
<scope>
  <mobile_devices>
    <mobile_device>
      <udid>270aae10800b6e61a2ee2bbc285eb967050b5984</udid>
    </mobile_device>
  </mobile_devices>
</scope>
```

seposium:
## Summary

Closes #217

The `--mobile-device` and `--computer` flags on `scope add` / `scope remove` previously always wrapped the supplied value in a `<name>` element. The Classic API requires individual devices to be identified by `<udid>` (40-char hex) or `<id>` (integer), not by name, so passing a UDID would silently fail — the server could not match it and the scope was not updated.

- Adds `namedItemFromIdentifier()`: detects whether a value is a 40-char hex UDID, a numeric ID, or a display name, and sets the correct `NamedItem` field.
- `AddToScope()` and `removeNamedItem()` now call this helper for `--mobile-device` and `--computer` flags.
- Improves duplicate detection in both functions to check `ID` and `UDID` fields in addition to `Name`.
- All other flag types (groups, buildings, departments, etc.) are unchanged — they continue to use name-only lookup.

## Test plan

- [x] `TestNamedItemFromIdentifier_UDID` — 40-char hex detected as UDID
- [x] `TestNamedItemFromIdentifier_NumericID` — all-digit string detected as ID
- [x] `TestNamedItemFromIdentifier_Name` — display name passthrough
- [x] `TestAddToScope_MobileDeviceByUDID` — UDID lands in `<udid>` field
- [x] `TestAddToScope_MobileDeviceByNumericID` — numeric ID lands in `<id>` field
- [x] `TestAddToScope_MobileDeviceByName` — display name lands in `<name>` field
- [x] `TestAddToScope_MobileDeviceByUDID_IdempotentUDID` — adding same UDID twice is a no-op
- [x] `TestAddToScope_MobileDeviceByUDID_IdempotentCaseInsensitive` — UDID case-insensitive dedup
- [x] `TestRemoveFromScope_MobileDeviceByUDID` — remove by UDID works
- [x] `TestRemoveFromScope_MobileDeviceByNumericID` — remove by numeric ID works
- [x] `go test ./...` — all 24 packages pass

## Context

This fix unblocks the [jamf-shared-mobile-kiosk](https://github.com/Jamf-Concepts/jamf-shared-mobile-kiosk) project, which shells out to `jamf-cli` to scope a CBA certificate profile to a specific iPhone after an RFID badge tap. The kiosk knows each device's UDID from the Jamf Pro inventory record; with this fix it can pass the UDID directly rather than requiring a pre-existing mobile device group.

### DIFF
--- a/internal/scope/scope.go
+++ b/internal/scope/scope.go
@@ -11,12 +11,39 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/Jamf-Concepts/jamf-cli/internal/registry"
 )
+
+// udidRe matches a 40-character hex string — the format of Apple device UDIDs.
+var udidRe = regexp.MustCompile(`^[0-9a-fA-F]{40}$`)
+
+// numericRe matches a plain integer string (Jamf Pro Classic API numeric ID).
+var numericRe = regexp.MustCompile(`^[0-9]+$`)
+
+// namedItemFromIdentifier builds a NamedItem with the correct field populated
+// based on what the caller passed: a 40-char hex UDID, a numeric ID, or a name.
+// Used for individual device scope targets where the API accepts any of the three.
+func namedItemFromIdentifier(value string) NamedItem {
+	switch {
+	case udidRe.MatchString(value):
+		return NamedItem{UDID: value}
+	case numericRe.MatchString(value):
+		return NamedItem{ID: value}
+	default:
+		return NamedItem{Name: value}
+	}
+}
+
+// isDeviceFlag returns true for flags that target individual devices (not groups),
+// where the caller may pass a UDID or numeric ID rather than a name.
+func isDeviceFlag(flagName string) bool {
+	return flagName == "mobile-device" || flagName == "computer"
+}
 
 // FetchScope performs a GET on a Classic API resource by name and returns
 // the resource's ID and parsed scope. When res.ResolveByList is true it lists
@@ -264,7 +291,9 @@ func AddToScope(s *ScopeXML, singularKey, section, flagName, name string) bool {
 	}
 
 	for _, item := range items.Items {
-		if strings.EqualFold(item.Name, name) {
+		if strings.EqualFold(item.Name, name) ||
+			(item.ID != "" && item.ID == name) ||
+			(item.UDID != "" && strings.EqualFold(item.UDID, name)) {
 			return false
 		}
 	}
@@ -272,7 +301,13 @@ func AddToScope(s *ScopeXML, singularKey, section, flagName, name string) bool {
 	if items.ElemName == "" {
 		items.ElemName = resolveElemName(section, flagName)
 	}
-	items.Items = append(items.Items, NamedItem{Name: name})
+	var item NamedItem
+	if isDeviceFlag(flagName) {
+		item = namedItemFromIdentifier(name)
+	} else {
+		item = NamedItem{Name: name}
+	}
+	items.Items = append(items.Items, item)
 	return true
 }
 
@@ -490,7 +525,9 @@ func removeNamedItem(items *ScopeItemSlice, name string) bool {
 	var keep []NamedItem
 	found := false
 	for _, item := range items.Items {
-		if strings.EqualFold(item.Name, name) {
+		if strings.EqualFold(item.Name, name) ||
+			(item.ID != "" && item.ID == name) ||
+			(item.UDID != "" && strings.EqualFold(item.UDID, name)) {
 			found = true
 			continue
 		}
--- a/internal/scope/scope_test.go
+++ b/internal/scope/scope_test.go
@@ -483,6 +483,148 @@ func TestValidateScopeCombination_ValidExclusions(t *testing.T) {
 	}
 }
 
+// ─── namedItemFromIdentifier ─────────────────────────────────────────────────
+
+func TestNamedItemFromIdentifier_UDID(t *testing.T) {
+	item := namedItemFromIdentifier("270aae10800b6e61a2ee2bbc285eb967050b5984")
+	if item.UDID != "270aae10800b6e61a2ee2bbc285eb967050b5984" {
+		t.Errorf("UDID = %q", item.UDID)
+	}
+	if item.Name != "" || item.ID != "" {
+		t.Errorf("unexpected fields set: name=%q id=%q", item.Name, item.ID)
+	}
+}
+
+func TestNamedItemFromIdentifier_NumericID(t *testing.T) {
+	item := namedItemFromIdentifier("42")
+	if item.ID != "42" {
+		t.Errorf("ID = %q", item.ID)
+	}
+	if item.Name != "" || item.UDID != "" {
+		t.Errorf("unexpected fields set: name=%q udid=%q", item.Name, item.UDID)
+	}
+}
+
+func TestNamedItemFromIdentifier_Name(t *testing.T) {
+	item := namedItemFromIdentifier("Josh's iPhone")
+	if item.Name != "Josh's iPhone" {
+		t.Errorf("Name = %q", item.Name)
+	}
+	if item.ID != "" || item.UDID != "" {
+		t.Errorf("unexpected fields set: id=%q udid=%q", item.ID, item.UDID)
+	}
+}
+
+// ─── AddToScope: mobile-device by UDID / ID ──────────────────────────────────
+
+func TestAddToScope_MobileDeviceByUDID(t *testing.T) {
+	s := &ScopeXML{}
+	udid := "270aae10800b6e61a2ee2bbc285eb967050b5984"
+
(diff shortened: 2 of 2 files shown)


ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Resolves `--mobile-device` / `--computer` scope flags by UDID or numeric ID instead of always wrapping the value in `<name>`. Core fix is correct, tests pass, consistent with the existing `VerifyItemInScope` 3-way match pattern.

The fix unblocks issue #217 and the downstream kiosk project cleanly. No critical or important findings. A few small things worth considering as follow-ups, but nothing blocking.

<details>

🟩 (1) (correctness, low-impact) — `internal/scope/scope.go:293-299, 521-541`: Dedup/remove now match by ID/UDID for non-device flags too

The PR description states: "All other flag types (groups, buildings, departments, etc.) are unchanged — they continue to use name-only lookup." That's true for **item construction** (the `isDeviceFlag` gate at line 305 makes sure only device flags use `namedItemFromIdentifier`). But the new dedup check in `AddToScope` and `removeNamedItem` runs unconditionally — it now matches by `ID` and `UDID` for **all** flag types.

After a server roundtrip via `FetchScope`, group/building items arrive with their `ID` populated. So:
- `scope add ... --computer-group "5"` against a fetched scope containing a group named `Acme` with `ID="5"` → dedup triggers, user is told "already in scope" but the names don't match.
- `scope remove ... --building "5"` → would remove the building whose `ID` is `5`, not the one named `5`.

In practice this is extremely unlikely (group names rarely consist of pure digits, never 40-char hex), and it's consistent with the pre-existing 3-way match in `VerifyItemInScope` (`scope.go:260-262`), so the *system* is internally consistent now. But the PR description claim isn't quite right.

**Suggested fix:** Either update the PR description to acknowledge this, or gate the ID/UDID matching to device flags by passing `flagName` into `removeNamedItem` and inlining the same check in `AddToScope`'s dedup loop:

```diff
- for _, item := range items.Items {
-     if strings.EqualFold(item.Name, name) ||
-         (item.ID != "" && item.ID == name) ||
-         (item.UDID != "" && strings.EqualFold(item.UDID, name)) {
-         return false
-     }
- }
+ isDevice := isDeviceFlag(flagName)
+ for _, item := range items.Items {
+     if strings.EqualFold(item.Name, name) ||
+         (isDevice && item.ID != "" && item.ID == name) ||
+         (isDevice && item.UDID != "" && strings.EqualFold(item.UDID, name)) {
+         return false
+     }
+ }
```

**Fixed when:** Either the PR description matches actual behavior, or non-device dedup/remove stays name-only.

🟩 (2) (UX) — `internal/scope/scope.go:688-694`: `appendNamedRows` hides UDID-only items in table output

After a successful `scope add --mobile-device <UDID>`, the in-memory `*ScopeXML` passed to `OutputScope` contains the new item as `{UDID: "270aae..."}` only — `Name` is empty until the next server fetch. `appendNamedRows` filters items with empty `Name`, so `scope add --mobile-device <UDID>` with `-o table` (default) shows the rest of the scope but **not** the device the user just added. JSON/YAML output is unaffected.

A quick stderr "Added mobile-device …" message above the output (`commands.go:95-96`) means users aren't fully blind, but it's a small papercut.

**Suggested fix:** Fall back to UDID then ID when `Name` is empty:

\`\`\`diff
 func appendNamedRows(rows *[]map[string]any, section, typeName string, items []NamedItem) {
     for _, item := range items {
-        if item.Name != "" {
-            *rows = append(*rows, map[string]any{"section": section, "type": typeName, "name": item.Name})
+        display := item.Name
+        if display == "" {
+            if item.UDID != "" {
+                display = item.UDID
+            } else if item.ID != "" {
+                display = item.ID
+            }
+        }
+        if display != "" {
+            *rows = append(*rows, map[string]any{"section": section, "type": typeName, "name": display})
         }
     }
 }
\`\`\`

**Fixed when:** \`scope add --mobile-device <UDID> -o table\` shows the newly-added device in the table output, even before a fresh server fetch.

🟩 (3) (test coverage) — \`internal/scope/scope_test.go\`: No test for \`--computer\` UDID/ID resolution

Every new test exercises \`--mobile-device\`. \`--computer\` goes through the same \`isDeviceFlag\` path so it's structurally covered, but an explicit \`TestAddToScope_ComputerByUDID\` (and one for numeric ID) would document the contract for the other supported flag and protect against accidental future divergence (e.g., if someone adds \`--ipad\` as a third device flag and forgets to wire it in).

**Suggested fix:** Add at minimum:
\`\`\`go
func TestAddToScope_ComputerByUDID(t *testing.T) {
    s := &ScopeXML{}
    udid := "270aae10800b6e61a2ee2bbc285eb967050b5984"
    if !AddToScope(s, "configuration_profile", "target", "computer", udid) {
        t.Fatal("expected true")
    }
    if s.Computers.Items[0].UDID != udid {
        t.Errorf("UDID = %q", s.Computers.Items[0].UDID)
    }
}
\`\`\`

**Fixed when:** \`--computer\` UDID/ID/Name selection has at least one direct test.

🟩 (4) (input validation, edge case) — \`internal/scope/scope.go:25-26\`: \`numericRe\` accepts \`"0"\` and leading zeros

\`^[0-9]+$\` matches \`"0"\`, \`"007"\`, \`"00000000000000000000"\`. Each would land in the \`<id>\` field and be sent to the API as a likely-invalid ID. Not a security or correctness issue — the server will reject — but a clearer error would be friendlier. Not worth fixing on its own; bundle with future input-validation work if it ever comes up.

</details>

<details>

- **Files reviewed:** \`internal/scope/scope.go\` (full read for context), \`internal/scope/scope_test.go\` (diff hunks)
- **Surrounding context checked:** \`internal/scope/types.go\` (\`NamedItem\` struct), \`internal/scope/commands.go\` (\`AddToScope\`/\`RemoveFromScope\` callers, scope add/remove flow), main's \`VerifyItemInScope\` (to confirm 3-way match was pre-existing)
- **Verified locally:** \`go test ./internal/scope/...\` passes (309ms)
- **Skipped:** Specialist agents (medium-risk diff, no security/auth/perf surface), repo-wide ripple analysis (functions called only from \`internal/scope/commands.go\`)

</details>

<details>

- ✅ Help text on \`--mobile-device\` and \`--computer\` (\`scope.go:506,508\`) already says "id, name, or UDID" — matches the new behavior, no docs drift.
- ✅ Pattern matches the pre-existing 3-way match in \`VerifyItemInScope\` rather than inventing a new convention — keeps the package internally consistent.
- ✅ Idempotency case-folded for UDIDs (\`TestAddToScope_MobileDeviceByUDID_IdempotentCaseInsensitive\`) — catches the realistic mistake where one source upper-cases the UDID and another lower-cases it.
- ✅ Single-commit, narrowly scoped PR with a clear motivating use case (kiosk project) in the description.

</details>

<details>
<summary>Acceptance criteria checklist</summary>

Linked issue #217 requested:
- \`--mobile-device <id-or-udid>\` accepted on \`scope add\` for Classic mobile config profiles ✅ — \`namedItemFromIdentifier\` covers both, plumbed through \`AddToScope\` via \`isDeviceFlag("mobile-device")\`.
- API path: emit \`<udid>\` (or \`<id>\`) instead of \`<name>\` ✅ — \`NamedItem\` struct already has \`xml:"udid,omitempty"\` and \`xml:"id,omitempty"\` (\`types.go:38-42\`); empty \`Name\` will still serialize as \`<name></name>\` because \`Name\` lacks \`omitempty\`, but Jamf accepts that.
- Generalizes to \`--computer\` (not in the issue but reasonable) ✅ — \`isDeviceFlag\` includes both.

**Potential gap worth checking before/after merge:** \`NamedItem.Name\` is \`xml:"name"\` (no \`omitempty\`) — UDID-only items will marshal as \`<mobile_device><udid>...</udid><name></name></mobile_device>\`. Live-test confirms the Classic API accepts this without complaint (per the PR description), but if any future Jamf server rejects empty \`<name>\`, you'd want to add \`omitempty\` to \`NamedItem.Name\`. Not blocking now.

</details>

Generated with Claude Code

PR #225, opened 2026-05-30
fix(normalize): escape all U+0000-001F control chars in JSON string values

neilmartin83:
## Summary

- Fixes a bug where shells like `zsh` interpret `\n` in `echo` output as real newline bytes (0x0A), producing invalid JSON that the YAML fallback would silently mangle (converting newlines to spaces in script contents, etc.)
- Replaces the original three-case switch (`\n`, `\r`, `\t` only) with a full RFC 8259 range check: any byte ≤ 0x1F inside a JSON string value is escaped as `\uXXXX`
- Removes four stale `NOTE: Also used by classic_registry.go` comments from functions not referenced by `classic_registry.go`; the two accurate ones (`readApplyInput`, `extractIDString`) are kept
- Removes dead `t.Skip` guard that could never fire (literal-newline bytes are never valid JSON)
- Updates test `want` strings to match `\uXXXX` output format; adds `TestEscapeJSONStringLiterals_OtherControlChars` to cover non-common control characters

## Test plan

- [x] `make test` passes
- [x] `make generate && make verify-generated` confirms generated code is in sync
- [x] Live test: `echo '{"name":"...","scriptContents":"#!/bin/bash\necho recon","priority":"AFTER"}' | jamf-cli pro scripts create` succeeds against a real Jamf Pro instance (script created with newline preserved in `scriptContents`)

### DIFF
--- a/generator/parser/generator.go
+++ b/generator/parser/generator.go
@@ -2753,7 +2753,6 @@ func renameResourceByID(ctx context.Context, client registry.HTTPClient, updateU
 // other fields untouched. Empty value is a no-op; empty body is bootstrapped
 // to {} first. Used to apply a user-supplied --name flag to the request body
 // without requiring them to hand-build JSON.
-// NOTE: Also used by classic helpers (same generated package).
 func setBodyStringField(body []byte, field, value string) ([]byte, error) {
 	if value == "" {
 		return body, nil
@@ -2786,7 +2785,6 @@ type fileFieldSpec struct {
 // If body is empty and at least one file flag is active, a new object is
 // constructed. File values overwrite any value already present in the body;
 // companion and name fields are only filled when absent.
-// NOTE: Also used by classic helpers (same generated package).
 func injectFileFields(body []byte, specs []fileFieldSpec) ([]byte, error) {
 	active := false
 	for _, s := range specs {
@@ -2901,7 +2899,6 @@ func readApplyInput(fromFile string) ([]byte, error) {
 
 // printScaffoldOutput prints a scaffold JSON string, converting to YAML when the
 // output format requests it.
-// NOTE: Also used by classic_registry.go helpers (same generated package).
 func printScaffoldOutput(jsonStr, format string) error {
 	if format == "yaml" {
 		var v any
@@ -2917,7 +2914,6 @@ func printScaffoldOutput(jsonStr, format string) error {
 }
 
 // normalizeInputToJSON converts YAML input to JSON. JSON input is returned unchanged.
-// NOTE: Also used by classic_registry.go helpers (same generated package).
 func normalizeInputToJSON(data []byte) ([]byte, error) {
 	// Fast path: already JSON
 	if json.Valid(data) {
@@ -2942,11 +2938,10 @@ func normalizeInputToJSON(data []byte) ([]byte, error) {
 	return out, nil
 }
 
-// escapeJSONStringLiterals escapes literal control characters (newlines, carriage
-// returns, tabs) that appear inside JSON string values. Fixes input produced by
-// shells like zsh whose built-in echo interprets \n even in single-quoted strings,
-// creating invalid JSON that the YAML fallback would silently mangle.
-// NOTE: Also used by classic_registry.go helpers (same generated package).
+// escapeJSONStringLiterals escapes any U+0000–U+001F control character that
+// appears inside a JSON string value. Fixes input produced by shells like zsh
+// whose built-in echo interprets \n even in single-quoted strings, creating
+// invalid JSON that the YAML fallback would silently mangle.
 func escapeJSONStringLiterals(data []byte) []byte {
 	var buf bytes.Buffer
 	buf.Grow(len(data))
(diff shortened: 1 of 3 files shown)


ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Fixes silent data corruption when piping JSON containing literal control characters (e.g. `zsh` `echo` turning `\n` into a real `0x0A`) into generated `apply`/`create`/`update` commands. Correct, well-tested, and the generated code is verified in sync. No critical or important issues.

## Findings

No blocking issues. The fix is sound: `escapeJSONStringLiterals` only runs *after* the `json.Valid` fast path fails and *before* the YAML fallback, so it's a pure improvement for the target case — inputs that previously got mangled by YAML (newlines folded to spaces) now round-trip as valid JSON. Inputs that don't become valid JSON after escaping still fall through to YAML exactly as before.

I verified the two substantive correctness risks:
- **State machine is byte-safe.** Ranging over `[]byte` yields bytes, and the `escaped`/`inString` interplay correctly prevents `\"` from terminating a string (test `..._EscapedQuoteNotMistreatedAsStringEnd` confirms). UTF-8 continuation/lead bytes are all ≥ 0x80, so the `b <= 0x1f` guard never touches multibyte runes — no corruption risk.
- **No scope gap on Classic.** `normalizeInputToJSON` is only on the modern path (167 generated files use it); Classic commands decode XML directly via `xml.Decoder` and never hit this function, so there's nothing equivalent to fix there.

<details>

🟩 (1) (design, c=75) — `generator/parser/generator.go:2919` / `registry.go:546`: Document the intentional JSON-wins precedence

Inserting the escape→JSON attempt ahead of the YAML fallback means a *flow-style YAML* document with a multi-line **double-quoted** scalar — e.g. `{"a": "x⏎y"}` — now resolves as JSON (`x\ny`) rather than YAML (which folds the newline to a space, `x y`). This is the *correct* choice for the CLI's dominant use case (shell-echoed JSON), and the collision is near-impossible in practice (YAML users use block literals `|`, not multi-line flow double-quotes). Worth a one-line comment so a future maintainer doesn't mistake it for an accident.

**Suggested fix:**
```diff
 	// Shells like zsh interpret \n in echo output as real newlines (0x0a), producing
 	// literal control characters inside JSON string values — invalid JSON. Try
 	// escaping those characters and retry before falling back to YAML, which would
-	// silently fold them into spaces.
+	// silently fold them into spaces. This takes precedence over YAML on purpose:
+	// for the JSON-with-literal-newlines shape we want \n preserved, not folded.
 	if fixed := escapeJSONStringLiterals(data); json.Valid(fixed) {
```

**Fixed when:** The precedence ordering carries a rationale comment. (Optional — behavior is already correct.)

🟩 (2) (docs, c=100) — PR description narrates intermediate commits, not the net diff

The description says it "replaces the original three-case switch" and "removes dead `t.Skip` guard" / "updates test `want` strings." Those describe changes between commit `7fb96a7` and `773e76f` *within this branch* — they're not visible in the net diff against `main` (where `escapeJSONStringLiterals` is net-new and `file_fields_test.go` is purely additive). Not a code issue; just flagging so the squash-merge commit body isn't misleading to someone reading it later.

**Fixed when:** Description (or squash message) reflects the net change vs `main`. Optional.

</details>

<details>

- [x] Design and architecture: escape step correctly slotted between JSON fast-path and YAML fallback; edits the generator template (not generated output) per the repo's generated-code boundary; `registry.go` confirmed in sync via `make verify-generated`
- [x] Correctness: traced the `escaped`/`inString`/`"`/`<=0x1f` state machine; verified `\"` handling, newline-outside-string passthrough, multibyte UTF-8 safety, and the `\`+literal-control-char edge (falls back to YAML, not made worse)
- [x] Security: no injection/secret/authz surface; input handling only, no credential flags added (CLAUDE.md credential policy unaffected)
- [na] Performance: `fmt.Fprintf` per control char is reflection-based but control chars are rare and this isn't a hot path — not worth flagging
- [x] Test coverage: 10 new cases incl. escaped-quote, CRLF, tab, NUL/BS/FF, newline-outside-string, JSON passthrough, zsh-echo regression, YAML block literal; all pass; pre-existing table test still green
- [x] Reliability: failure path preserved — invalid-after-escape input still routes to YAML; no swallowed errors
- [x] Code quality: stale `NOTE: Also used by classic_registry.go` comments removal verified accurate via grep (0 refs for the 4 removed; 1 and 4 refs for the 2 kept)
- [x] Simplification: no redundant abstraction; function is minimal
- [na] Frontend concerns: N/A
- [x] Project rules compliance: no `.claude/rules/` dir; honored CLAUDE.md generated-code boundary (template edited, output regenerated) and credential-input policy

</details>

<details>

- **Reviewed in full:** all 3 files (+209/−8) — `generator/parser/generator.go` (template), `internal/commands/pro/generated/registry.go` (regenerated output), `file_fields_test.go` (new tests)
- **Verified:** `go test ./internal/commands/pro/generated/` (full package + targeted tests) pass; `make verify-generated` reports "Generated code is up to date"; classic input path inspected to rule out a scope gap; branch commit history (`7fb96a7`, `773e76f`) and `classic_registry.go` references checked via grep
- **Not run:** full `make test` across all packages — change is isolated to one helper + its call sites in the generated package, which was tested directly
- **Prior reviews:** none (first review)

</details>

<details>

- ✅ The test suite anticipates the genuinely tricky cases — escaped quotes not terminating strings, and control-chars *outside* strings passing through untouched — which are exactly where a hand-rolled scanner usually breaks.
- ✅ The stale-comment cleanup wasn't guesswork: the 4 removed `NOTE: Also used by classic_registry.go` annotations are verifiably false (0 references) while the 2 kept ones (`readApplyInput`, `extractIDString`) are real shared dependencies. Tightening docs in the same PR that touches the file is the right call.

</details>

_Generated by `pr-review` v1.7.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

PR #240, opened 2026-06-15
feat(hooks): add PreToolUse spec-check reminder for jamf-cli mutations

neilmartin83:
## Summary

- Adds a `PreToolUse` Bash hook to `.claude/settings.json` that fires whenever a `jamf-cli create|update|apply|patch` command is about to run
- `--scaffold` runs are explicitly excluded (they're part of the prep workflow, not a mutation)
- Non-blocking (exit 0): prints a stderr reminder with the three spec URLs inline so the agent is prompted at action time, not from skill text read 40+ turns earlier
- Blocking (exit 2) was considered and rejected: it creates a permanent loop — agent fetches spec, retries, gets blocked again indefinitely

## Why a hook, not more skill text

SKILL.md directives are read once at skill invocation. By the time the agent runs `apply`, that text is stale context. A hook fires fresh on every matching Bash call regardless of context window decay — that's the structural difference between advisory and enforced.

## Smoke test results

| Command | Expected | Result |
|---|---|---|
| `jamf-cli … apply` | warn | ✅ |
| `jamf-cli … create` | warn | ✅ |
| `jamf-cli … update` | warn | ✅ |
| `jamf-cli … patch` | warn | ✅ |
| `jamf-cli … create --scaffold` | pass | ✅ |
| `jamf-cli … list` | pass | ✅ |
| `jamf-cli … delete 1` | pass | ✅ |
| `git status` | pass | ✅ |

## Test plan

- [ ] In a session with the plugin loaded, run `/jamf-cli` then attempt `jamf-cli -p <profile> pro scripts apply` — confirm the spec-check reminder appears in the tool result before the command executes
- [ ] Confirm `jamf-cli … create --scaffold` passes through without a warning

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -9,6 +9,15 @@
             "command": "file_path=$(jq -r '.tool_input.file_path'); case \"$file_path\" in */internal/commands/generated/*) echo 'BLOCK: generated file — run make generate instead' >&2; exit 2;; esac"
           }
         ]
+      },
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "cmd=$(jq -r '.tool_input.command // empty' 2>/dev/null); [ -z \"$cmd\" ] && exit 0; echo \"$cmd\" | grep -qE 'jamf-cli' || exit 0; echo \"$cmd\" | grep -qE '\\b(create|update|apply|patch)\\b' || exit 0; echo \"$cmd\" | grep -q -- '--scaffold' && exit 0; printf '[jamf-cli] Spec check required before this mutation.\\nLoad the API spec first (WebFetch), then run --scaffold to get the exact template.\\n  Platform (blueprints/compliance-benchmarks/platform-devices/platform-device-groups/ddm-reports):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/platform/<resource>-api.json\\n  Classic API (/JSSResource/ paths):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/classic/resources.yaml\\n  Pro modern API (all other resources):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/<ResourceName>.yaml\\n' >&2; exit 0"
+          }
+        ]
       }
     ],
     "PostToolUse": [
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -9,15 +9,6 @@
             "command": "file_path=$(jq -r '.tool_input.file_path'); case \"$file_path\" in */internal/commands/generated/*) echo 'BLOCK: generated file — run make generate instead' >&2; exit 2;; esac"
           }
         ]
-      },
-      {
-        "matcher": "Bash",
-        "hooks": [
-          {
-            "type": "command",
-            "command": "cmd=$(jq -r '.tool_input.command // empty' 2>/dev/null); [ -z \"$cmd\" ] && exit 0; echo \"$cmd\" | grep -qE 'jamf-cli' || exit 0; echo \"$cmd\" | grep -qE '\\b(create|update|apply|patch)\\b' || exit 0; echo \"$cmd\" | grep -q -- '--scaffold' && exit 0; printf '[jamf-cli] Spec check required before this mutation.\\nLoad the API spec first (WebFetch), then run --scaffold to get the exact template.\\n  Platform (blueprints/compliance-benchmarks/platform-devices/platform-device-groups/ddm-reports):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/platform/<resource>-api.json\\n  Classic API (/JSSResource/ paths):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/classic/resources.yaml\\n  Pro modern API (all other resources):\\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/<ResourceName>.yaml\\n' >&2; exit 0"
-          }
-        ]
       }
     ],
     "PostToolUse": [
--- a/skills/hooks/hooks.json
+++ b/skills/hooks/hooks.json
@@ -1,0 +1,15 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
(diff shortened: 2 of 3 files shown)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Adds a `PreToolUse` Bash hook (`skills/hooks/hooks.json`) that nudges the agent to load the API spec before a `jamf-cli create|update|apply|patch`, plus a SKILL.md section on Platform device-group operations.
> Nothing breaks as-is (the hook is non-blocking and correctly placed), but **(1) is strongly recommended before merge** — it defeats the hook's purpose for 4 of 5 Platform resources. Everything else is nice-to-have.

## Rating: 4/5
- Would be a 5 by fixing the Platform spec-path mapping in (1) so the reminder points at files that actually exist, and tightening the matcher in (2).

The structural call here is good: a hook fires fresh on every Bash call, immune to context decay, where SKILL.md text goes stale 40 turns in. The JSON shape, the auto-discovered `skills/hooks/hooks.json` location, the `Bash` matcher, and the non-blocking exit-0/stderr contract are all **correct** — I verified the shape against every shipped plugin (including `concepts-audit` in this same `jamf-plugins` marketplace) and reproduced the smoke-test table exactly. The issues below are about the *content* the hook prints, not its wiring.

### 🟧 Important

**🟧 (1) (correctness, c=100) — `skills/hooks/hooks.json:9`: the Platform spec-path the hook prints 404s for 4 of the 5 resources it names**

The reminder tells the agent to fetch `specs/platform/<resource>-api.json` for `blueprints / compliance-benchmarks / platform-devices / platform-device-groups / ddm-reports`. But literal substitution only resolves for `blueprints`. The actual files in `specs/platform/` are:

| Resource named in hook | Naïve `<resource>-api.json` | Actual file | Resolves? |
|---|---|---|---|
| `blueprints` | `blueprints-api.json` | `blueprints-api.json` | ✅ |
| `compliance-benchmarks` | `compliance-benchmarks-api.json` | `jamf-compliance-benchmark-engine-api.json` | ❌ |
| `platform-devices` | `platform-devices-api.json` | `device-inventory-api.json` / `device-management-actions-api.json` | ❌ |
| `platform-device-groups` | `platform-device-groups-api.json` | `device-groups-api.json` | ❌ |
| `ddm-reports` | `ddm-reports-api.json` | `Declaration-reporting-openapi.json` | ❌ |

Within this PR's own threat model — the agent has suffered context decay and is leaning on the freshly-printed stderr — there is no way to derive `jamf-compliance-benchmark-engine-api.json` from the resource name. The hook sends it to a 404. (This template is inherited from the existing SKILL.md table at `SKILL.md:118`, whose own example `device-groups-api.json` already contradicts the `<resource>-api.json` rule — so the PR is faithfully propagating a pre-existing ambiguity into a new surface, not inventing it. Fixing SKILL.md too is the deeper cleanup; at minimum the *new* hook text shouldn't imply a clean substitution.)

**Suggested fix** — point at the directory instead of implying a filename, since the stderr is standalone and can't carry the mapping table:
```diff
-  Platform (blueprints/compliance-benchmarks/platform-devices/platform-device-groups/ddm-reports):\n    https://raw.githubusercontent.com/Jamf-Concepts/jamf-cli/main/specs/platform/<resource>-api.json\n
+  Platform (blueprints/compliance-benchmarks/platform-devices/platform-device-groups/ddm-reports):\n    Browse https://github.com/Jamf-Concepts/jamf-cli/tree/main/specs/platform to find the file (names are abbreviated, e.g. compliance-benchmarks -> jamf-compliance-benchmark-engine-api.json).\n
```

**Fixed when:** following the hook's stderr for any of the five Platform resources leads the agent to a spec file that exists (no 404), without relying on the SKILL.md table being in context.

### Nice-to-have

<details>

**🟩 (2) (correctness, c=100) — `skills/hooks/hooks.json:9`: matcher fires on incidental verb/`jamf-cli` text anywhere in the command line**

Both `grep -qE 'jamf-cli'` and `grep -qE '\b(create|update|apply|patch)\b'` match anywhere in the string, not the jamf-cli subcommand position. I reproduced spurious reminders on read-only commands:

```
WARN  jamf-cli -p x pro scripts list && git commit -m "update docs"
WARN  jamf-cli -p x pro policies list | grep create
WARN  echo "remember to jamf-cli apply later"
```

Low harm (it's a non-blocking nudge), but spurious reminders erode the signal the hook is trying to create. Robust shell-line parsing is hard, but requiring the verb to follow `jamf-cli` (e.g. `grep -qE 'jamf-cli\b[^|&;]*\b(create|update|apply|patch)\b'`) trims the common false positives.

**Fixed when:** the three commands above pass without a warning while the author's smoke-table mutations still warn.

**🟩 (3) (docs/fidelity, c=100) — PR description names the wrong file**

The body says "Adds a `PreToolUse` Bash hook to `.claude/settings.json`". The final state puts it in `skills/hooks/hooks.json` (per commit `230672b fix(hooks): move spec-check hook…`), and `.claude/settings.json` is unchanged — it still contains only the repo's Go-dev hooks (generated-file guard, gofumpt, go.mod regen). The plugin-tree placement is the *correct* one for marketplace distribution, so this is just a stale description. Update it (and tick the test-plan boxes, or note the manual verification is still pending) so the next reader isn't sent to the wrong file.

**🟩 (4) (correctness/scoping, c=75) — body-bearing mutations outside `create|update|apply|patch` aren't covered**

`jamf-cli … blueprints deploy "My BP"` and `… import-profile` pass without a reminder, and `… create --scaffold | … apply` is suppressed by the `--scaffold` substring even though the pipeline ends in a real `apply`. `deploy`/`undeploy`/`clone`/`scope` arguably don't construct a request body so excluding them is defensible — but `import-profile` does take a file. Worth a deliberate decision on which verbs the "spec check" applies to, documented next to the regex.

**🟩 (5) (maintainability, c=100) — spec-path template now lives in three places**

`SKILL.md:118` (table), `SKILL.md:127` (step list), and the new hook all carry the same `specs/platform/<resource>-api.json` template. A spec-layout change now requires three edits or they drift. The "fresh context" rationale justifies the hook holding its own copy, but a one-line comment cross-referencing the SKILL.md section would flag the coupling.

**🟩 (6) (reliability, c=75) — silent no-op when `jq` is absent**

If `jq` isn't on PATH, `cmd` comes back empty and `[ -z "$cmd" ] && exit 0` swallows the reminder — the "enforcement" silently vanishes. `jq` is a suite-wide dependency (the skills pipe through it everywhere), so this is unlikely in practice, but it's the kind of silent degradation that makes a safety hook quietly stop working. A fallback grep on the raw stdin would harden it.

</details>

<details>

- [x] Design and architecture: hook-vs-skill-text rationale is sound; placement in plugin tree (`skills/hooks/`) is correct for marketplace distribution and auto-discovery — verified against shipped plugins.
- [x] Correctness: validated JSON parses; reproduced the smoke-test table; confirmed spec-path 404s by listing `specs/platform/`; confirmed false-positive matches by executing the hook command against sample stdin.
- [x] Security: non-blocking advisory hook, no injection surface (reads `.tool_input.command`, only writes to stderr, exit 0); no secrets.
- [na] Performance: trivial grep/printf on each Bash call; negligible.
- [x] Test coverage: PR ships a manual smoke table (accurate, reproduced) but no automated test; CI (`ci.yaml`) does Go lint only — no JSON/plugin validation, so the hook file's correctness rests entirely on review.
- [x] Reliability: traced the jq-absent path (silent no-op, finding (6)); exit-0/stderr contract confirmed against docs and live plugins.
- [x] Code quality: hook one-liner is dense but readable; SKILL.md additions are clear.
- [x] Simplification: noted the 3-way template duplication (5); no dead code introduced.
- [na] Frontend concerns: none.
- [x] Project rules compliance: no `.claude/rules/` in this repo; checked root `CLAUDE.md` for the `specs/` layout convention — the hook's path template is consistent with it (and inherits its ambiguity).

</details>

<details>

- Reviewed: full diff (45 lines, 2 files) at `8f8602e` vs `origin/main`.
- Verified: hooks.json shape against 9 shipped plugin hook files; `specs/platform/` + `specs/classic/` directory contents; root `.claude/settings.json` (orphan check — clean); commit history (3 commits, including the move); CI workflow scope; SKILL.md existing "Request body policy" section for consistency.
- Executed: the hook command against 13 sample `tool_input.command` payloads to confirm warn/pass behavior.
- No specialist agents (low-risk: skill/hook config, no auth/perf/migration surface). No prior reviews on the PR.

</details>

<details>

- ✅ Correctly resisted putting the hook in `.claude/settings.json` (author-local only) and placed it in the plugin tree where the marketplace actually distributes it — and matched the exact JSON convention used by `concepts-audit` in the same marketplace.
- ✅ The "blocking (exit 2) was considered and rejected — it creates a permanent fetch→retry→block loop" reasoning in the PR body is exactly the right call; the non-blocking nudge is the correct design.

</details>

_Generated by `pr-review` v1.8.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

grahampugh (approved):

PR #255, opened 2026-06-23
feat(protect): SDK v0.7.0 network content filter download + --unsigned profile exports

neilmartin83:
## Summary

Upgrades `jamfprotect-go-sdk` v0.6.1 → v0.7.0 and surfaces its new feature, plus adds an `--unsigned` option for re-uploading downloaded profiles to Jamf Pro.

### SDK v0.7.0
The only new feature in v0.7.0 is the **`networkContentFilterProfile`** organization download.

- New command: `protect downloads network-content-filter-profile`
- `protect downloads summary` now reports `hasNetworkContentFilterProfile`

### `--unsigned` profile exports
Jamf Protect returns its config profiles (PPPC, tamper-prevention, network content filter) as **CMS/PKCS7-signed** mobileconfigs. Jamf Pro signs profiles itself on delivery and **rejects pre-signed input**, so the signed download can't be re-uploaded directly.

Added `--unsigned` to all three signed-profile downloads. It strips the CMS `signedData` envelope and writes the raw mobileconfig payload — enabling an end-to-end download → upload-to-Jamf-Pro flow.

```
jamf-cli -p prod protect downloads network-content-filter-profile --unsigned -O ncf.mobileconfig
```

Signature stripping uses **`golang.org/x/crypto/cryptobyte`** (already in the module graph, Go-team maintained) to walk the CMS structure and extract the encapsulated content — **no third-party PKCS7 dependency** added.

## Testing
- `make test` / `make lint` — green (0 lint issues)
- New unit test `TestExtractPKCS7Content` against a real signed fixture
- Smoke-tested against a live tenant: all three profiles export both signed (default) and unsigned; `plutil -lint` validates every unsigned output as a well-formed plist


ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Upgrades `jamfprotect-go-sdk` to v0.7.0, adds the `network-content-filter-profile` download, and adds `--unsigned` (CMS signature stripping) to the three signed-profile downloads so they can be re-uploaded to Jamf Pro. No critical or important issues found.

## Rating: 5/5
- Correct CMS/ASN.1 walk, real-fixture test, clean dependency hygiene. Only a docs nit keeps this from being literally nothing-to-say.

### Review coverage

The core of this PR is `extractPKCS7Content` (`protect_org.go:451-498`). I traced it against RFC 5652 and it's correct:

`ContentInfo` SEQUENCE → `contentType` OID (verified `== 1.2.840.113549.1.7.2` signedData) → `[0] EXPLICIT` wrapper → `SignedData` SEQUENCE → skip `version` INTEGER + `digestAlgorithms` SET → `encapContentInfo` SEQUENCE → skip `eContentType` OID → `[0] EXPLICIT` → `OCTET STRING` payload. Each step has a distinct, accurate error message, and the empty/detached-payload cases are handled. Parsing is via `cryptobyte` (Go-team maintained, bounds-checked) so malformed input fails gracefully rather than panicking — no memory-safety or injection surface. The "no signature verification" choice is correct and clearly documented: the goal is to strip-and-re-upload, and Jamf Pro re-signs on delivery.

Verified independently:
- `go test -run TestExtractPKCS7Content` — passes (real-fixture happy path + non-signed error path).
- `go mod tidy` — **no diff**; promoting `golang.org/x/crypto` from indirect to direct `require` is correct since it's now imported directly.
- `go build ./...` and `go vet ./internal/commands/` — clean.
- `writeProfileFile`'s stderr confirmation (`"Saved to %s (%d bytes, signature stripped)"`) matches the existing `writeBase64File` style.
- No wiring/enumeration test asserts the download subcommand set, so nothing breaks; the new subcommand inherits the `groupProtectOrg` group from its parent.

<details>

🟩 (1) (docs, c=75) — `CLAUDE.md:214`: download subcommand list is now incomplete

The curated list explicitly enumerates download subcommands but omits the new one and the new flag:

> `protect downloads` subcommands fetch installer/uninstaller/pppc-profile/tamper-prevention-profile/root-ca/csr/websocket-auth/summary files.

**Suggested fix:**
```diff
-`protect downloads` subcommands fetch installer/uninstaller/pppc-profile/tamper-prevention-profile/root-ca/csr/websocket-auth/summary files.
+`protect downloads` subcommands fetch installer/uninstaller/pppc-profile/tamper-prevention-profile/network-content-filter-profile/root-ca/csr/websocket-auth/summary files. The three signed-profile downloads (pppc-profile, tamper-prevention-profile, network-content-filter-profile) accept `--unsigned` to strip the CMS signature for re-upload to Jamf Pro.
```

**Fixed when:** the CLAUDE.md list includes `network-content-filter-profile` and mentions `--unsigned`.

🟩 (2) (reliability, c=50) — `protect_org.go:493`: DER-only assumption for `eContent`

`cryptobyte` parses DER only. If a tenant ever returns a CMS envelope whose `eContent` OCTET STRING is BER-constructed (chunked / indefinite-length, which some streaming signers emit for large payloads), `ReadASN1(&payload, OCTET_STRING)` returns `"malformed eContent payload"`. This is a *graceful* failure, not corruption, and the live smoke test confirms current Jamf Protect output is DER with small payloads — so this is purely a robustness note, not a bug. No action needed unless you see that error in the wild; if so, the fix is to reassemble constructed octet-string segments.

</details>

<details>

✅ Avoided pulling in a third-party PKCS7 library by walking the structure with the already-in-graph `cryptobyte` — the right call for a single-purpose unwrap, and it keeps the dependency surface minimal.

✅ The test asserts *behavior* (payload is valid plist XML, and is strictly smaller than the signed envelope) rather than byte-exact output, so it won't be brittle against fixture regeneration.

</details>

---
🤖 Reviewed with the Jamf Claude Code skill (pr-review v1.10.0)

PR #259, opened 2026-06-28
feat(mcp): mark destructive commands in the commands catalog

ktn-jamf:
## What

Threads the existing `jamf:destructive` cobra annotation into the `commands` catalog (`commandEntry` → `commandEntriesToMaps`), so the MCP `list_commands` tool tells agents which commands mutate or erase state **before** they run them. Sharpens the `list_commands` tool description to mention the marker and the `--yes` requirement.

## Why

`mcp serve`'s `list_commands` already serves `commands -o json`, and generated/handwritten destructive commands already carry `Annotations["jamf:destructive"]="true"` (coverage enforced by `TestDestructiveVerbCommandsAreAnnotated`). The catalog just wasn't surfacing it. An agent previously only discovered a command was destructive when it got refused for missing `--yes`.

## Design

- **Positive-only:** emits `"destructive": true` only; a non-destructive command omits the key entirely. Never emits `destructive: false`, so an unflagged command can't broadcast a false "safe" signal.
- No annotation edits — the field is read-only from existing annotations.

## Tests

- `TestCollectCommands_DestructiveFlag` — `pro computers-inventory delete` → `destructive:true`; `pro computers-inventory list` → key absent.
- `TestCommandEntriesToMaps_DestructivePositiveOnly` — positive-only behavior.
- `go test ./internal/commands/` PASS, `go build ./...` clean, `make lint` clean. Existing `TestDestructiveVerbCommandsAreAnnotated` still passes.

Reviewer note (Minor, non-blocking): the unit-test fixtures in `TestCommandEntriesToMaps_DestructivePositiveOnly` use synthetic command strings; harmless but could carry a clarifying comment.

### DIFF
--- a/internal/commands/mcp.go
+++ b/internal/commands/mcp.go
@@ -78,7 +78,9 @@ Configure it in an MCP client (example for Claude Desktop's config):
 			mcp.AddTool(server, &mcp.Tool{
 				Name: "list_commands",
 				Description: "List every available jamf-cli command with its description and " +
-					"flags. Call this first to discover what you can run, then use run_command.",
+					"flags. Commands that mutate or erase state are marked \"destructive\": true " +
+					"and require an explicit --yes. Call this first to discover what you can run, " +
+					"then use run_command.",
 			}, func(ctx context.Context, _ *mcp.CallToolRequest, _ struct{}) (*mcp.CallToolResult, any, error) {
 				return runChild(ctx, executable, serverProfile, []string{"commands", "-o", "json"}), nil, nil
 			})
--- a/internal/commands/root.go
+++ b/internal/commands/root.go
@@ -829,6 +829,7 @@ type commandEntry struct {
 	Flags       []string `json:"flags,omitempty"`
 	Product     string   `json:"product,omitempty"`
 	Group       string   `json:"group,omitempty"`
+	Destructive bool     `json:"destructive,omitempty"`
 }
 
 // newCommandsCmd creates the "commands" subcommand that outputs the full
@@ -882,6 +883,7 @@ func collectCommands(cmd *cobra.Command, prefix, product, group string) []comman
 				Description: child.Short,
 				Product:     childProduct,
 				Group:       childGroup,
+				Destructive: child.Annotations["jamf:destructive"] == "true",
 			}
 
 			// Collect aliases: for leaf commands under a top-level group
@@ -937,6 +939,9 @@ func commandEntriesToMaps(entries []commandEntry, full bool) []map[string]any {
 			m["flags"] = flags
 			m["product"] = e.Product
 			m["group"] = e.Group
+			if e.Destructive {
+				m["destructive"] = true
+			}
 		}
 		result[i] = m
 	}
--- a/internal/commands/root.go
+++ b/internal/commands/root.go
@@ -939,9 +939,10 @@ func commandEntriesToMaps(entries []commandEntry, full bool) []map[string]any {
 			m["flags"] = flags
 			m["product"] = e.Product
 			m["group"] = e.Group
-			if e.Destructive {
-				m["destructive"] = true
-			}
+			// Emit unconditionally (not just when true) so CSV and table output,
+			// which derive their columns from the first row, carry the field for
+			// every row rather than dropping it when the first row isn't destructive.
+			m["destructive"] = e.Destructive
 		}
 		result[i] = m
 	}
--- a/internal/commands/root_test.go
+++ b/internal/commands/root_test.go
@@ -1298,17 +1298,20 @@ func TestCollectCommands_DestructiveFlag(t *testing.T) {
 	}
 }
 
-func TestCommandEntriesToMaps_DestructivePositiveOnly(t *testing.T) {
+func TestCommandEntriesToMaps_Destructive(t *testing.T) {
+	// Synthetic entries — command names are arbitrary for this unit test.
 	entries := []commandEntry{
-		{Command: "pro computers delete", Description: "Delete", Destructive: true},
-		{Command: "pro computers list", Description: "List"},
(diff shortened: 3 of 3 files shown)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes (one decision)** — Threads the existing `jamf:destructive` annotation into the `commands` catalog so MCP `list_commands` marks mutating commands. The MCP/JSON path works correctly; the open item is that the new field is silently absent in CSV and table output.
> Blocking decision: (1). One nice-to-have in the collapsed section.

## Rating: 4/5
- Would be a 5 by either emitting the field unconditionally (so it's first-class across all formats) or consciously scoping it to JSON/YAML and documenting that. The stated MCP goal is already met.

---

🟧 (1) (correctness/consistency, c=100) — `internal/commands/root.go:939-941`: the `destructive` marker only surfaces in JSON/YAML; CSV and `--wide` table silently drop it.

Verified empirically against HEAD (`aa72684`):

| Format | Result |
|--------|--------|
| `commands -o json` | 157 entries marked `"destructive": true` ✓ |
| `commands -o yaml` | 157 ✓ |
| `commands -o csv` | **no `destructive` column at all** |
| `commands --wide -o table` | **no `DESTRUCTIVE` column** |

**Root cause:** CSV/table derive their column set from the *first* row (`sortedKeys(v[0])` at `internal/output/output.go:287`; the wide-table path behaves the same). Every other full-mode field (`aliases`/`flags`/`product`/`group`) is always set, so the columns stay uniform. `destructive` is the first *conditionally-present* key — emitted positive-only via `if e.Destructive { m["destructive"] = true }` — and the first command row (`completion`) isn't destructive, so the column never appears for any row.

**Why it matters:** the MCP `list_commands` path uses `-o json` and is correct, so the PR's stated goal is met. But the field is invisible in two of the catalog's output formats, which (a) surprises a human/agent using `commands -o csv` or `--wide`, and (b) is exactly the cross-format gap this repo's output-flag-matrix convention exists to prevent.

This is a **design decision, not a clear bug** — positive-only emission was a deliberate choice (avoid broadcasting a false `"destructive": false` "safe" signal). Two ways to a 5; author's call:

**Option A — make it first-class across formats** (emit the bool unconditionally):
```diff
-			if e.Destructive {
-				m["destructive"] = true
-			}
+			m["destructive"] = e.Destructive
```
Tradeoff: JSON then shows `"destructive": false` on non-destructive commands. Coverage of destructive commands is enforced by `TestDestructiveVerbCommandsAreAnnotated`, so `false` is accurate for every verb-detected command; the only residual is a truly-destructive command with a non-standard verb *and* no annotation, which would then show `false` (a positive but possibly-wrong claim) rather than being silently absent.

**Option B — keep positive-only, make the scope conscious:** leave the code as-is, document that `destructive` is a structured-output-only (JSON/YAML) field, and add a test asserting it appears in `-o json`. Accept that CSV/table don't carry it.

**Fixed when:** either the field appears consistently across json/yaml/csv/table (Option A), verified by a test asserting the CSV header includes `destructive`; **or** the JSON/YAML-only scope is documented as intentional with a test pinning the JSON behavior (Option B).

<details>

🟩 (2) (test-quality, c=100) — `internal/commands/root_test.go` `TestCommandEntriesToMaps_DestructivePositiveOnly`: the fixture strings `"pro computers delete"` / `"pro computers list"` aren't real catalog paths (the sibling integration test correctly uses `computers-inventory`, since `computers` is an alias). Harmless — this unit test feeds `commandEntriesToMaps` synthetic input, so the names are arbitrary — but a reader may expect real paths after seeing the integration test corrected.

**Suggested fix:** add a clarifying comment or use neutral names.
```diff
-		{Command: "pro computers delete", Description: "Delete", Destructive: true},
-		{Command: "pro computers list", Description: "List"},
+		// synthetic entries — command names are arbitrary for this unit test
+		{Command: "x delete", Description: "Delete", Destructive: true},
+		{Command: "x list", Description: "List"},
```

**Fixed when:** the synthetic nature is clear to a reader.

</details>

<details>

- [x] Design and architecture: reusing the enforced `jamf:destructive` annotation as the single source of truth (vs. re-deriving destructiveness) is the right call — coverage can't silently rot because `TestDestructiveVerbCommandsAreAnnotated` gates it.
- [x] Correctness: positive-only emission is correct for JSON/YAML; nil-map read of `Annotations` is safe; verified the full-mode gate (`commands -o json` → full=true) so the MCP consumer receives the field. Cross-format gap captured in (1).
- [na] Security: informational metadata only; the `--yes` enforcement already exists in `mcp.go`'s child invocation and is unchanged.
- [x] Test coverage: integration test reads the real annotation (guards against key drift) + unit test pins positive-only emission. Fixture-naming nit in (2).
- [na] Performance: no hot paths or queries changed.
- [na] Reliability: no failure paths changed.
- [x] Code quality: `json:"destructive,omitempty"` struct tag is consistent with sibling fields (the output path uses the maps, not the struct, so the tag is cosmetic but conventional).
- [x] Documentation currency: the `commands` JSON schema is auto-surfaced into the site's `commands.json`/`llms.txt` from the binary, so no manual doc update is required; the companion PR #260 documents the marker for agents.
- [na] Simplification: minimal diff, nothing redundant.
- [na] Frontend concerns: none.
- [x] Project rules compliance: no `.claude/rules/` directory; CLAUDE.md credential/generated-code policies unaffected.

</details>

<details>

Reviewed all 54 changed lines across `mcp.go`, `root.go`, `root_test.go`. Read `internal/output/output.go` (column derivation) and `internal/commands/destructive_annotations_test.go` (annotation contract) for context. Built the binary at HEAD and exercised `commands` across json/yaml/csv/table to verify finding (1). No prior reviews on this PR.

</details>

---
🤖 Reviewed with the pr-review Jamf Claude Code skill

neilmartin83 (approved):

PR #264, opened 2026-07-03
feat(security): add Jamf Security Cloud namespace (Risk, Device Lifecycle, SSE)

### LINKED ISSUE #234: Jamf Security Cloud Risk API
Would be good to get access to the RISK API within JSC to get that data in... just a thought.

neilmartin83:
## Summary

Fixes #234.

- Adds a new `security` product namespace for **Jamf Security Cloud** (Radar) — Risk API, Device Lifecycle API, and Shared Signals & Events (SSE) API.
- Commands are **generator-owned**, same as Platform: `generator/parser/security.go` + `generator/security/` hand-map the twelve total operations (too few and too irregular for the generic tag/family auto-detectors Platform's parser uses) and generate `internal/commands/security/generated/` via `make generate`. Nothing is hand-written except `security setup`.
- Auth is unique to this product: each of the three APIs is provisioned as its own "Security Integration" in the Radar portal with its own application ID/secret, and the resulting JWT is scoped to exactly one API via its `aud` claim (confirmed against real sandbox credentials during development). `internal/security.Client` tracks three independent credential pairs/token caches (Basic-login-for-JWT against `api.wandera.com`, SSE data-plane on `sse.jamf.com`) — any subset may be configured.
- `security setup` prompts for each API's credentials independently (skippable, at least one required); no URL prompt since all customers share the same global hosts — tenancy is carried in the JWT.
- Commands: `security risk list/override`, `security device-lifecycle purge` (destructive, `--yes`-gated), `security stream/status/verification/jwks/well-known`.
- Private spec drop location `specs/.security-source/` (gitignored) syncs into committed `specs/security/` via `make sync-security-specs`, mirroring the existing Platform spec pattern.
- Wired into `config validate`, `multi`, the showcase site (stat card/tab/color/labels), and CLAUDE.md.

## Test plan

- [x] `make build && make test && make lint` all pass
- [x] `make verify-site`, `make verify-site-output` pass (new "security" product fully covered)
- [x] Verified live against a Jamf Security Cloud sandbox tenant:
  - `security risk list` / `--guid` filter — correct auth, pagination unwrap, response shape
  - `security risk override`, `security device-lifecycle purge`, `security stream update` `--scaffold` output
  - "credentials not configured" and `--yes`-required error paths
  - a real `purge` call — confirmed the target device disappeared from a subsequent `risk list`
- [x] `go run /generator` idempotency check (running `make generate` twice produces no diff)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Adds a new `security` product namespace for Jamf Security Cloud (Risk, Device Lifecycle, SSE) with a bespoke per-scope JWT auth client and a generator-owned command set. Design is clean and the docs (CLAUDE.md, site) are thorough — but a few real issues on destructive and `--set` paths, one flatly-wrong Makefile comment, and the total absence of tests hold it back.
> Blocking: (1), (2), (3). Strongly recommend (4) and (5). Everything else is collapsed nice-to-haves.

## Rating: 3/5
- Would be a 5 with a local guard + informative confirmation on `purge` (1), the Makefile comment corrected (2), the `--set` clobber fixed (3), and a starter test suite for the auth client + pure helpers (4).

---

The architecture is sound: three independent scoped credential pairs / token caches, a single global host with tenancy carried in the JWT, generator-owned CRUD mirroring the Platform contract, and genuinely good doc comments. `security setup` correctly obeys the credential-input policy (interactive-only, keychain storage, `--no-input` rejected, no secret-bearing flags), `setup` correctly bypasses auth resolution, and generated commands feed `PrintRaw` so `-o json/yaml/csv/table` all work. The findings are about robustness on the destructive/auth paths, one misleading comment, and verification — not the design.

## 🟧 Important

**🟧 (1) (correctness / destructive-safety, via devil-advocate, c=75) — `internal/commands/security/generated/device_lifecycle.go:45` (source: `generator/security/template.go`): `purge` confirms blind and never requires a scoping filter**

Three things compound on an irreversible bulk-delete endpoint:
1. `ConfirmAction("purge", "device-lifecycle", yes)` runs **before** `LifecycleCustomerID` resolves the customer (line 49), so the prompt — `About to purge "device-lifecycle". Continue? [y/N]` — structurally cannot say *which customer or which devices* are about to be destroyed.
2. `security.ReadBody(bodyFile, setFlags)` returns `(nil, nil)` when neither `--file` nor `--set` is given (`internal/security/body.go:34`), and `doExpect`/`do` send the `POST` with **no body at all** (not even `{}`).
3. There is no local guard requiring `externalIds`/`guids` to be non-empty.

So `jamf-cli security device-lifecycle purge --yes` fires an unfiltered, bodyless `POST …/devices/purge/async/external`. Whether that purges nothing, errors, or **purges the entire tenant's fleet** depends on server semantics this repo can't confirm — and "we don't know if an empty request wipes everything" is itself the reason to guard it locally. Even the careful interactive user gets a confirmation that reveals nothing.

**Suggested fix** (in the template, so every future destructive Device-Lifecycle op inherits it):
```diff
-			if err := security.ConfirmAction("purge", "device-lifecycle", yes); err != nil {
-				return err
-			}
 			path := "/device-lifecycle/v1/{customerId}/devices/purge/async/external"
 			customerID, err := cliCtx.SecurityClient.LifecycleCustomerID(cmd.Context())
 			if err != nil {
 				return err
 			}
 			body, err := security.ReadBody(bodyFile, setFlags)
 			if err != nil {
 				return err
 			}
+			if body == nil {
+				return fmt.Errorf("purge requires --file or --set specifying externalIds/guids; refusing an unfiltered purge")
+			}
+			if err := security.ConfirmAction(fmt.Sprintf("purge devices for customer %q", customerID), "device-lifecycle", yes); err != nil {
+				return err
+			}
 			path = strings.Replace(path, "{customerId}", url.PathEscape(customerID), 1)
```

**Fixed when:** `purge` with no `--file`/`--set` errors locally before any network call, and the confirmation (and its `--yes` bypass) name the target customer. Please also confirm the API's empty-body semantics — if an empty filter means "purge all," raise this to critical.

---

**🟧 (2) (correctness / docs, c=100) — `Makefile:143-147`: the `sync-security-specs` comment states the exact opposite of how security works**

The comment reads: *"Unlike platform, these are NOT fed into the code generator — security commands are hand-written (`internal/commands/security_*.go`) … specs/security/ exists purely as committed reference documentation."*

Every clause is false, and the PR contradicts it in three places: the commands **are** generator-owned (`generator/parser/security.go` + `generator/security/` → `internal/commands/security/generated/`), `make generate` **does** consume `specs/security/*.json` (`generator/main.go`), and this PR's own CLAUDE.md says *"Commands are **generator-owned**, same as Platform"* and *"the twelve … operations are all generator-owned."* Only `security setup` is hand-written. A contributor reading the Makefile would conclude editing the specs does nothing and editing generated files is safe — both wrong.

**Fixed when:** the comment describes security as generator-owned (specs feed the generator; new ops need a `securityOpsByFile` entry; regenerate with `make generate`) and no longer claims the commands are hand-written.

---

**🟧 (3) (correctness, via silent-failure-hunter, c=100) — `internal/security/body.go:55-68`: `applySet` silently discards a value when a `--set` path descends through a non-object**

When a dotted `--set` key walks through a segment already holding a scalar/array (from `--file` or an earlier `--set`), `next, ok := m[segment].(map[string]any)` fails and the code replaces the existing value with a fresh empty map — no error. `security stream update --file cfg.json --set delivery.method=push`, where `cfg.json` has `"delivery": "immediate"`, silently becomes `{"method":"push"}` and drops the original. The codebase's own Pro `setNestedValue` (`internal/commands/pro/generated/registry.go:750`) returns `fmt.Errorf("cannot set nested key under non-object field %q", …)` for exactly this — security should match.

```diff
-func applySet(m map[string]any, path []string, v any) {
+func applySet(m map[string]any, path []string, v any) error {
 	for i, segment := range path {
 		if i == len(path)-1 {
 			m[segment] = v
-			return
+			return nil
 		}
-		next, ok := m[segment].(map[string]any)
-		if !ok {
-			next = map[string]any{}
-			m[segment] = next
+		existing, present := m[segment]
+		next, ok := existing.(map[string]any)
+		if !ok {
+			if present {
+				return fmt.Errorf("cannot set nested key under non-object field %q", segment)
+			}
+			next = map[string]any{}
+			m[segment] = next
 		}
 		m = next
 	}
+	return nil
 }
```

**Fixed when:** a `--set` path through an existing scalar/array field errors instead of silently replacing it.

---

**🟧 (4) (test-coverage, via test-quality-reviewer, c=100) — the new package ships with zero tests**

`internal/security/*` and `generator/security/*` add ~1500 lines of non-trivial logic with **no test files** (0.0% coverage on every new function). Peer products test the analogous surfaces: `internal/client/status_error_test.go` (Pro's `httpStatusError` — security's copy is untested), `generator/platform/generator_test.go` (sibling of `LoadResources`/`ParseSecuritySpec`), `internal/commands/pro/generated/patch_helpers_test.go` (the `--set` coercion). Highest-value, lowest-cost targets — all pure or seam-testable:
- `body.go` — `parseSetValue`, `applySet` (see (3)), `ReadBody`.
- `client.go` — `decodeJWTClaims` (2-segment token, bad base64, missing `exp`), `httpStatusError` (401/403/429 → exit codes), and an `httptest`-backed `doExpect`/`login` round-trip (the `WithHTTPClient`/`WithAPIBaseURL` seams make this cheap — same pattern as `internal/client/client_test.go`). An SSE-login round-trip would also settle (11).
- `generator/parser/security.go` — `ParseSecuritySpec`, `stripCustomerIDPathParam`, `filterParamByInAndName`.

**Fixed when:** the error mapping + JWT decode, the `body.go` pure functions, and the spec parser have unit tests, with ≥1 `httptest` round-trip through `doExpect`.

---

**🟧 (5) (correctness / generator, via silent-failure-hunter, c=75) — `generator/security/emitter.go:97-103`: a paginated op that fails array-key detection silently loses pagination *and* its query flags**

`paginate := unwrapKey != ""`, and `detectListArrayKey` returns `""` unless the 2xx response is an object with exactly one array property. In `template.go` the query-flag **declarations and registrations are both nested inside `{{- if .Paginate }}`** — so when detection fails, the op regenerates as a plain GET with *no filter flags at all* and a changed output shape, and `Generate()` emits no warning. Harmless today (`userDeviceList` is the sole array property), but CLAUDE.md documents adding endpoints via `securityOpsByFile`, so a future spec adding a second array field would quietly drop `--external-id`/`--guid` with a green build.

**Fixed when:** an `IsList` + page-param op that yields `unwrapKey == ""` fails `make generate` with a clear message, and non-paging query flags render regardless of `.Paginate`.

---

<details>
<summary>Nice-to-have suggestions (8 items)</summary>

**🟩 (6) (code-quality, c=100) — `internal/security/body.go:72,85`: `parseSetValue` coerces floats but its doc claims it "mirrors" the Pro parser, which doesn't.** `parsePatchValue` (`registry.go:769`) parses only `int64` then falls back to string — no `ParseFloat`. So `--set version=1.10` → float `1.1` here but string `"1.10"` in Pro. Drop the float branch to match, or fix the comment to state the intentional divergence.

**🟩 (7) (correctness, via devil-advocate, c=100) — `internal/commands/root.go:1361-1381`: env/profile merge can splice a mismatched id/secret pair.** `fillPair` backfills `*id` and `*secret` independently, so exporting only `JAMFSECURITY_RISK_CLIENT_ID` pulls the secret from the profile — a mismatched pair surfacing as an opaque 401. Treat each pair as atomic: if one half is from env and the other would come from the profile, error with "set both or neither."

**🟩 (8) (security, c=100) — `internal/commands/root.go` (`resolveSecurityClient`): no `http://` guard on `JAMFSECURITY_URL`/`JAMFSECURITY_SSE_URL`.** These flow into `WithAPIBaseURL`/`WithSSEBaseURL` with no scheme check, yet `login()` ships `base64(id:secret)` Basic auth there. The Pro flow warns on plaintext and setup rejects `http://` — add the same here.

**🟩 (9) (security, c=100) — `internal/security/client.go` `login()`: unbounded response-body read.** `login()` uses bare `io.ReadAll(resp.Body)` while `doExpect` caps at 100MB and `do()` at 64KB. A misbehaving login host could OOM the process. Wrap with `io.LimitReader(resp.Body, 64<<10)`.

**🟩 (10) (reliability, via silent-failure-hunter + devil-advocate) — `internal/security/client.go` auth hardening cluster:**
- *Missing `exp` (c=75):* `decodeJWTClaims` doesn't check `exp` presence — a token without it yields `expiresAt = Unix(0)`, so `ensureToken` re-logs in on **every** request; combined with per-page `ensureToken` in `risk list` that doubles volume against the 5 req/sec limit and shows up as a confusing mid-pagination 429.
- *`aud` never validated (c=100):* the package comment leans on JWT `aud` scoping, but `jwtClaims` never decodes `aud`, so a wrong-scope credential (Risk creds in the Lifecycle slot) is trusted locally and its `customer_id` used for a purge. Decode `aud` and fail fast on mismatch.
- *Empty `customer_id` (c=100):* `LifecycleCustomerID` returns it unchecked; empty makes the purge path `/device-lifecycle/v1//devices/…` (double slash). Return a clear local error.

**🟩 (11) (reliability, c=100) — `generator/security/template.go`: the "fetch everything" pagination loop has no upper bound.** `for page := 0; ; page++ { … if len(Items) < pageSize { break } }` never terminates if the server ignores `page` and keeps returning full pages — the CLI hangs while `aggregated` grows unbounded. Add a sanity `maxPages` cap that errors out.

**🟩 (12) (silent-failure, c=75) — `internal/commands/security_setup.go:23`: stdin handling in `securityCredentialPair`.** (a) `line, _ := reader.ReadString('\n')` discards the read error, so a stdin failure is indistinguishable from an intentional skip; (b) the ID is read via `bufio.Reader` while the secret is read via `term.ReadPassword` on the raw fd — buffered type-ahead/paste can desync them, and this file chains three pairs through one shared reader so a desync cascades. Interactive-only enforcement limits blast radius; surfacing the error and asserting `reader.Buffered() == 0` before the password read closes it. (The pairing exists elsewhere; this is the only place chaining three.)

**🟩 (13) (maintainability, c=75) — `Makefile` `verify-security-specs` omits the generated-code check `verify-platform-specs` has.** Its comment says "same as verify-platform-specs," but the platform target *also* regenerates and diffs `internal/commands/platform/generated/`; the security one stops at `specs/security/`, so generated code can drift undetected. (Neither `verify-*-specs` runs in CI today — only `verify-generated`, which is pro-only — so it's developer-facing, but worth aligning; optionally extend `verify-generated` to cover `security/generated/`.)

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `internal/commands/security/generated/device_lifecycle.go` | 2/5 | Unfiltered/blind `purge` — (1) |
| `internal/security/body.go` | 3/5 | `applySet` clobber (3); `parseSetValue` float/doc (6) |
| `internal/security/client.go` | 3/5 | Solid design; hardening gaps (9),(10) |
| `internal/commands/security_setup.go` | 3/5 | Policy-compliant; stdin handling (12) |
| `generator/security/emitter.go` · `template.go` | 3/5 | Silent pagination/flag drop (5); unbounded loop (11) |
| `internal/commands/root.go` | 4/5 | Clean; credential-splice (7), http guard (8) |
| `generator/parser/security.go` | 4/5 | Clean hand-mapping; untested (4) |
| `Makefile` | 3/5 | Wrong `sync-security-specs` comment (2); verify gap (13) |
| `internal/config/config.go` · `security.go` · `groups.go` · `aliases.go` · site | 5/5 | Clean, consistent wiring |

</details>

<details>

- **Reviewed in full:** all hand-written Go (`internal/security/`, `security_setup.go`, `security.go`, `resolveSecurityClient`, config/registry/multi/config-validate wiring), the whole generator (`generator/parser/security.go`, `generator/security/`, `generator/main.go`), all generated commands, Makefile, CLAUDE.md, `.gitignore`, aliases/groups, site + verify-script changes.
- **Verified:** regenerated via `go run ./generator` — committed `security/generated/` matches (idempotent, no drift); confirmed `PrintRaw` reformats per `-o`; traced `setup` auth-skip and the empty-body purge path in code.
- **Spec JSON** (~2481 lines) skimmed as data.
- **Specialists:** security, silent-failure, test-quality, devil-advocate — all findings ran through per-file Haiku confidence verification.
- **Not verified (needs API knowledge):** empty-body `purge` server semantics (1); whether SSE creds authenticate at `api.wandera.com/v1/login` (folded into (4)).

</details>

_Generated by `pr-review` v1.11.2, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Re-review of the single fix commit `2369b25` ("address PR review findings"). All three blocking findings (1)(2)(3) and both strongly-recommended ones (4)(5) are resolved, along with every nice-to-have except one residual sub-point. Verified: `go build ./...`, the new test suites (`internal/security/…`, `generator/security/…`, `generator/parser/security_test.go`) all pass, and `security/generated/` regenerates idempotently.

## Rating: 5/5
- The destructive/`--set`/pagination paths are now guarded and, more importantly, *tested*; the one open item (validate the JWT `aud` claim) is a defence-in-depth nice-to-have, not a blocker.

## Prior findings status

<details>
<summary>Prior findings status (13 findings — 12 fixed, 1 partial)</summary>

| (1) | `security/generated/device_lifecycle.go` (template) | ✅ Fixed | `purge` now refuses when `body == nil` (`--file`/`--set` required), and confirmation resolves `customerID` *first* then names it: `purge for customer %q`. Local guard blocks the unfiltered-purge case regardless of server semantics. |
| (2) | `Makefile:143` | ✅ Fixed | Comment rewritten — now correctly states security is generator-owned (specs feed the generator, new ops need a `securityOpsByFile` entry + `make generate`); only `security setup` is hand-written. |
| (3) | `internal/security/body.go` | ✅ Fixed | `applySet` returns an error and no longer clobbers: descending through an existing non-object errors `cannot set nested key under non-object field %q`, matching Pro's `setNestedValue`. Covered by `TestApplySet_ErrorsThroughNonObject` / `_ThroughArray` and `TestReadBody_SetThroughFileScalarErrors`. |
| (4) | new package | ✅ Fixed | 5 test files added. `client_test.go` covers `decodeJWTClaims` (valid/missing-exp/malformed/bad-b64/bad-json), `httpStatusError` 401/403/404/429/500 → exit codes, an `httptest` login+`DoExpect` round-trip asserting token caching, the unconfigured-scope no-network path, and empty-`customer_id`. `body_test.go`, `emitter_test.go`, `security_test.go`, `confirm_test.go` round it out. |
| (5) | `generator/security/emitter.go` + `template.go` | ✅ Fixed | Paginated op with `unwrapKey == ""` now fails `make generate` loudly (`buildTemplateResource` returns an error naming the resource/op). Query-flag decls + registrations moved *outside* `{{- if .Paginate }}`, and the non-paginated branch now builds/applies the query string. Regression tests `TestBuildTemplateResource_AmbiguousArrayKeyErrors` and `TestGenerate_NonPaginatedOpRendersQueryFlags`. |
| (6) | `internal/security/body.go` | ✅ Fixed | `parseSetValue` doc now states the intentional divergence from Pro's int64-only `parsePatchValue` (Security has no merge-patch content-type constraint), rather than falsely claiming to mirror it. |
| (7) | `internal/commands/root.go` `fillPair` | ✅ Fixed | `fillPair` takes a `scope` and treats id/secret as atomic: env-half + profile-half now errors `set both via env or neither` instead of splicing a mismatched pair. |
| (8) | `internal/commands/root.go` `resolveSecurityClient` | ✅ Fixed | Emits `WARNING: using HTTP (not HTTPS) — credentials will be sent in plaintext` when `url`/`sseURL` start with `http://`, mirroring the Pro runtime flow. |
| (9) | `internal/security/client.go` `login()` | ✅ Fixed | `io.ReadAll(io.LimitReader(resp.Body, 64<<10))` — login body read now capped. |
| (10) | `internal/security/client.go` auth cluster | ⚠️ Partial | **exp** ✅ (`decodeJWTClaims` errors on `exp == 0`, killing the re-login-every-request path) and **empty `customer_id`** ✅ (`LifecycleCustomerID` errors). **`aud` still not validated** — `jwtClaims` decodes only `customer_id`/`exp`, so a wrong-scope credential is still trusted locally. Carried forward as the sole active finding below. |
| (11) | `generator/security/template.go` | ✅ Fixed | `maxPages = 10000` backstop errors instead of looping forever when the server ignores `page`. |
| (12) | `internal/commands/security_setup.go` | ✅ Fixed | ID-read error now surfaced (`err != io.EOF`), and `reader.Buffered() != 0` is asserted before the `term.ReadPassword` raw-fd read to catch buffered type-ahead desync. |
| (13) | `Makefile` `verify-security-specs` | ✅ Fixed | Target now removes generated files, re-runs `make generate`, and diffs `internal/commands/security/generated/` — closing the drift gap vs `verify-platform-specs`. |

</details>

## Active findings

<details>

🟩 (1) (security, c=100) — `internal/security/client.go:168`: JWT `aud` claim still not decoded or validated

The package doc (`client.go:11`) leans on per-API `aud` scoping (`RISK_API`, etc.) as the isolation guarantee, but `jwtClaims` decodes only `customer_id` and `exp`. A credential dropped into the wrong slot (e.g. Risk creds configured as `JAMFSECURITY_LIFECYCLE_*`) is trusted locally, and its `customer_id` would drive a `purge`. The server's own `aud` check is the real backstop, so this is defence-in-depth, not a hole — hence still a nice-to-have.

**Suggested fix:**
```diff
 type jwtClaims struct {
 	CustomerID string `json:"customer_id"`
 	ExpiresAt  int64  `json:"exp"`
+	Audience   string `json:"aud"`
 }
```
Then have each scope's `login()` assert the decoded `aud` matches the expected API constant, failing fast on mismatch.

**Fixed when:** `decodeJWTClaims` decodes `aud` and each scope rejects a token whose audience doesn't match, with a unit test for the wrong-scope case.

</details>

<details>

- **Baseline:** prior review on `443fe24` (reachable — clean incremental diff). Single fix commit since: `2369b25`.
- **Re-reviewed:** the full incremental diff `443fe24..2369b25` — Makefile, `generator/security/{emitter,template}.go`, `generator/parser/security.go` behavior via new tests, `internal/security/{body,client}.go`, `internal/commands/root.go` `fillPair`, `security_setup.go`, and the regenerated `device_lifecycle.go`/`risk.go`/`stream.go`.
- **Verified:** `go build ./...` clean; `go test ./internal/security/... ./generator/security/... ./generator/parser/...` all pass; confirmed `security/generated/` regenerates with zero diff (idempotent). Read the current `template.go` end-to-end to confirm query flags render + apply in both the paginated (inside the fetch loop) and non-paginated branches, and that `body` still flows through both.
- **No new issues** introduced by the fix commit.
- **Not re-run:** specialist agents — the fix commit introduced no new high-risk surface.

</details>

_Generated by `pr-review` v1.11.2, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

PR #270, opened 2026-07-08
feat(pro): add "update --set" fetch-merge-put for PUT endpoints

neilmartin83:
## What

Generated modern-API `update` commands now accept repeatable `--set KEY=VALUE` flags for partial field updates — no more `get | jq | update` round-trips.

When `--set` is used, the CLI:
1. **GET**s the current resource,
2. strips read-only / server-computed fields against the PUT request schema,
3. deep-merges the caller's changes,
4. **PUT**s the whole record back.

Bare `update` (stdin / `--from-file`) is unchanged — still a full-replace PUT of the supplied payload. Get-merge-put only happens with `--set`.

```
# before
jamf-cli pro categories get 12 -o json | jq '.priority = 9' | jamf-cli pro categories update 12

# now
jamf-cli pro categories update 12 --set priority=9
jamf-cli pro categories update --name "Apps" --set priority=9   # name lookup
```

## How (all at the generator level)

- **Schema-driven filtering** — a `fieldFilter` tree built from the PUT request schema strips response-only/read-only fields the endpoint would reject (prevents 400s). Nested objects filter recursively; arrays are kept whole. Open/unmodelled schemas skip filtering so nothing is over-dropped.
- **Write-only awareness** — added `WriteOnly` to the parser. Required + write-only fields (e.g. `gsx-connection.token`) can't be read back from a GET, so `update --set` warns when they aren't supplied.
- **Optimistic version locking** — handled automatically. Each `--set` re-fetches, so the current `versionLock` (top-level *and* nested, e.g. `accountSettings.versionLock`) rides along and sequential updates never conflict.
- **Exclusions** — file-field and token-composite update ops keep full-replace-only semantics (their bespoke body injection doesn't compose with a fetched body). Classic (XML) endpoints are out of scope.

54 generated `update` commands gain `--set`. Runtime helpers (`fieldFilter`, `deepMergeJSON`, `fetchForMerge`) are emitted to `registry.go`.

## Files

- `generator/parser/{types,parser,generator}.go` — `WriteOnly`, predicates, filter-literal emitter, `--set` template wiring
- `generator/parser/generator_test.go`, `internal/commands/pro/generated/update_set_test.go` — unit tests
- `internal/commands/pro/generated/*.go` — regenerated (62 files)

## Testing

`make test` green, `make lint` clean (0 issues), `make verify-generated` in sync.

**Live-tested against a real instance** (`-p platform-nmartin`):

| Case | Result |
|---|---|
| Computer prestage — nested `--set` (`authenticationPrompt` + `accountSettings.payloadConfigured`) | ✅ both changed; top-level *and* nested versionLock incremented; 42 keys preserved |
| Mobile device prestage — scalar `--set` + sequential updates | ✅ versionLock 0→1→2→3, no conflicts; all 51 keys preserved |
| Categories — create → `--set` by ID → `--set` by `--name` → delete | ✅ zero residual; read-only `id` dropped from body |
| Singleton `client-check-in` (no ID) | ✅ no-op `--set` PUT, all keys unchanged |
| Advanced search — `--set` scalar | ✅ `criteria` / `displayFields` arrays survived untouched |
| Scripts (file-field) | ✅ correctly has **no** `--set` |

**Notes:**
- Computer prestages return HTTP 500 on *successful* update (known product issue) — verified every change via follow-up GET.
- Prestage GET→PUT can hit a server-side `accountSettings.payloadConfigured` validation. This is inherent to the endpoint — the existing `get | update` full-replace flow hits the **identical** 400. `--set` is faithful to that behavior, not worse.


ktn-jamf (commented):
> [!TIP]
> ✅ **Merge-ready** — Adds a repeatable `--set KEY=VALUE` flag to 54 generated PUT `update` commands, doing a client-side fetch → strip-read-only → deep-merge → PUT. Well-scoped, gated, live-tested, and generated code is in sync. No blocking issues.

## Rating: 4/5
- Would be a 5 with the writable-field-filter assumption in (1) documented or spot-audited, plus a command-level round-trip test (3).

The build passes, `generator/parser` + `generated` tests are green, `make verify-generated` reports in sync, and I independently confirmed `versionLock` (top-level and nested `accountSettings.versionLock`) is kept as a leaf in the emitted filter — so the optimistic-locking claim holds. The design is genuinely clean: the guard (`updateSetEnabled` requires a PUT-with-modelled-body plus a co-located GET), the exclusions (file-field / token-composite ops keep full-replace), and the untouched bare-`update` path all show careful blast-radius control.

<details>

🟩 **(1) (design, c=75)** — `generator/parser/generator.go:1268` (`writableFilterLiteral` / `fieldFilter.apply`): the filter is a **whitelist** — it keeps only fields present *and* non-`readOnly` in the PUT **request** schema and deletes everything else from the fetched body. This diverges from the existing `get | jq | update` flow, which PUTs the *entire* GET body back. If any of the 54 resources has a field that is **writable but not declared in the PUT request schema** (present only in the GET/response schema), `--set` will silently drop it on write-back — potential silent data loss where the old flow preserved it.

This is conditional on request/response schema divergence, and the live tests (42/51 keys preserved on prestages) suggest fidelity is good in practice — hence nice-to-have, not blocking. But it's worth locking down since the blast radius is 54 resources.

**Suggested fix:** either (a) add a one-line note in `writableFilterLiteral`'s doc comment stating the assumption ("request schema is assumed to enumerate all writable fields; response-only writable fields are dropped"), or (b) spot-audit a couple of resources where the GET and PUT reference different component schemas. An inverted design (blacklist: keep everything, drop only fields explicitly marked `readOnly`) would trade the data-loss risk for a 400 risk — not necessarily better, but worth a sentence explaining why the whitelist was chosen.

**Fixed when:** the whitelist assumption is documented at the emitter, or an audit confirms no enabled resource has response-only writable fields.

🟩 **(2) (usability, c=75)** — generated `update` RunE (e.g. `internal/commands/pro/generated/categories.go:387`): when `--set` is supplied, piped stdin is silently ignored (`if len(flagSet) == 0 && (stat.Mode()...)`). A user who does `echo '{...}' | jamf-cli pro categories update 1 --set priority=9` gets only the `--set` path with no indication their stdin was dropped.

**Suggested fix:** in the template, when `len(flagSet) > 0` and stdin is a non-char device (piped), emit a warning to stderr (or error) that `--set` and stdin are mutually exclusive.

**Fixed when:** combining `--set` with piped stdin warns or errors instead of silently ignoring the pipe.

🟩 **(3) (test-coverage, c=75)** — `internal/commands/pro/generated/update_set_test.go`: the end-to-end test exercises the helpers (`fieldFilter` → `buildMergePatchFromSet` → `deepMergeJSON`) directly but never drives a generated `update --set` command through its RunE with a fake `registry.HTTPClient`. The GET-path-equals-PUT-path wiring, the `len(flagSet)==0` stdin gate, and the write-only warning are only validated by live testing, not by a repeatable test.

**Suggested fix:** add one table test that runs a generated update command (e.g. categories) with a stub client asserting: (a) a GET fires at the resolved path, (b) the PUT body has read-only fields stripped and the `--set` field merged, (c) stdin is not read when `--set` is present.

**Fixed when:** a command-level test covers the fetch→filter→merge→PUT round trip against a fake client.

</details>

<details>

✅ The exclusion logic (`updateSetEnabled` bailing on `FileFields`/`UpdateTokenOp`, and requiring a co-located GET) is exactly the right instinct — it prevents `--set` from being emitted where fetch-merge-put can't compose with bespoke body injection.

✅ The write-only-required warning is a thoughtful touch: fields like `gsx-connection.token` can't be read back from a GET, and surfacing that at runtime (rather than letting the PUT fail opaquely) is real DX care.

✅ Cycle guard (`visited` by `SchemaRef`) plus a depth cap of 6 in `schemaFilterLiteral` — defensive against self-referential schemas blowing up code generation.

</details>

<details>

- Reviewed in full: `generator/parser/generator.go` (+323, core logic), `parser.go`, `types.go`, both test files, and the emitted runtime helpers in `registry.go`.
- Spot-checked generated output: `categories.go` (simple), `computer_prestages.go` + `mobile_device_prestages.go` (nested versionLock).
- Verified locally: `go build ./...`, `go test ./generator/parser/ ./internal/commands/pro/generated/`, `make verify-generated`.
- Confirmed no update-set-enabled op appends query params to the path (would otherwise leak into the `fetchForMerge` GET) — latent generator concern only, does not manifest on current specs.
- The 62 mechanically-regenerated command files were sampled, not read line-by-line — `make verify-generated` confirms they match the template.

</details>

This covers all findings — the PR is merge-ready; the above are polish.

---
<sub>🤖 Reviewed with the Jamf Claude Code skill (pr-review v1.11.2)</sub>

ktn-jamf (approved):

PR #272, opened 2026-07-08
feat(backup): add --download-packages for JCDS-hosted package binaries

### LINKED ISSUE #268: Include package downloading, not just the meta data
I love that the backup function gets the meta data from packages, but it would be great if it downloaded the actual packages as well. The https://yourServer.jamfcloud.com/api/v1/packages allows for this.

neilmartin83:
## Summary
- Closes #268.
- `backup` previously only exported package *metadata* — this adds an opt-in `--download-packages` flag that also downloads each package's binary to `packages/files/`.
- Only JCDS-hosted packages can be downloaded reliably (no line-of-sight/credentials assumption for on-prem or third-party cloud distribution points). Packages with no match in `/v1/jcds/files` are skipped with a warning, not counted as a failure — this matches the scope discussed with @mattdjerome on the issue.
- Real download failures (e.g. transient API errors) are recorded in `_failures.yaml` like any other backup failure, and participate in the existing partial-failure exit code.

## Implementation
- Reuses the existing `jcdsDownloadFile`/`jcdsFileData` helpers in `pro_jcds.go` (presigned-URL fetch + 403-retry-once logic) instead of reimplementing them.
- Extracted the `GET /v1/jcds/files` + unmarshal step (previously inlined in `jcds sync`) into a shared `jcdsListFiles` helper, now used by both `jcds sync` and the new backup path.

## Test plan
- [x] `make build && make lint && make test` all pass
- [x] New tests: happy-path download, skip-when-not-on-JCDS, download-failure handling, no-op when `packages` is excluded via `--resources`, and default-off behavior when the flag isn't passed

### DIFF
--- a/internal/commands/pro_backup.go
+++ b/internal/commands/pro_backup.go
@@ -13,9 +13,11 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 	"gopkg.in/yaml.v3"
 
 	"github.com/Jamf-Concepts/jamf-cli/internal/auth"
@@ -63,11 +65,12 @@ type backupMeta struct {
 
 func newBackupCmd(cliCtx *registry.CLIContext) *cobra.Command {
 	var (
-		outputDir   string
-		format      string
-		resources   string
-		includeIDs  bool
-		concurrency int
+		outputDir        string
+		format           string
+		resources        string
+		includeIDs       bool
+		concurrency      int
+		downloadPackages bool
 	)
 
 	cmd := &cobra.Command{
@@ -79,14 +82,21 @@ Each object is saved as an individual YAML or JSON file. Server-generated fields
 (id, timestamps) are stripped by default for clean version-control diffs.
 
 Partial failures are tolerated — objects that fail to fetch are recorded in
-_failures.yaml and the backup continues.`,
+_failures.yaml and the backup continues.
+
+--download-packages additionally downloads the package binaries backing the
+"packages" resource, saved to packages/files/. Only packages hosted on the
+Jamf Cloud Distribution Service (JCDS) can be downloaded this way — packages
+on other distribution points (on-prem file share, third-party cloud) have no
+reliable download path and are skipped with a warning.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runBackup(cmd.Context(), cliCtx, backupOptions{
-				OutputDir:   outputDir,
-				Format:      format,
-				Resources:   resources,
-				IncludeIDs:  includeIDs,
-				Concurrency: concurrency,
+				OutputDir:        outputDir,
+				Format:           format,
+				Resources:        resources,
+				IncludeIDs:       includeIDs,
+				Concurrency:      concurrency,
+				DownloadPackages: downloadPackages,
 			})
 		},
 	}
@@ -96,17 +106,19 @@ _failures.yaml and the backup continues.`,
 	cmd.Flags().StringVar(&resources, "resources", "", "comma-separated resource filter (e.g., policies,scripts)")
 	cmd.Flags().BoolVar(&includeIDs, "include-ids", false, "retain server-generated IDs in output")
 	cmd.Flags().IntVar(&concurrency, "concurrency", backupDefaultConcurrency, fmt.Sprintf("max parallel API requests (ceiling %d)", backupMaxConcurrency))
+	cmd.Flags().BoolVar(&downloadPackages, "download-packages", false, "also download package binaries hosted on JCDS to packages/files/")
 	_ = cmd.MarkFlagRequired("output")
 
 	return cmd
 }
 
 type backupOptions struct {
-	OutputDir   string
-	Format      string
-	Resources   string
-	IncludeIDs  bool
-	Concurrency int
+	OutputDir        string
+	Format           string
+	Resources        string
+	IncludeIDs       bool
+	Concurrency      int
+	DownloadPackages bool
 }
 
 func runBackup(ctx context.Context, cliCtx *registry.CLIContext, opts backupOptions) error {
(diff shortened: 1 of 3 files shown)


ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Adds an opt-in `--download-packages` flag to `pro backup` that downloads JCDS-hosted package binaries into `packages/files/`, reusing the existing JCDS presigned-URL helpers. No critical or important issues found.

## Rating: 5/5
- Clean, well-scoped, and thoroughly tested. Reuses `jcdsDownloadFile`/`jcdsListFiles` instead of reimplementing, correct partial-failure exit semantics, and five test scenarios covering the meaningful paths. The three suggestions below are optional edge-case polish — none worth holding the merge.

This is a tidy PR. The design choice to gate on JCDS presence (skip-with-warning rather than fail for on-prem/third-party packages) matches the scope agreed on issue #268, and the `jcdsListFiles` extraction is a clean, behavior-preserving refactor that de-duplicates the list logic between `jcds sync` and the new backup path.

<details>

🟩 **(1) (correctness/efficiency, c=75) — `pro_backup.go:277` & `:701`: Duplicate `filename` values cause redundant concurrent downloads and inflate the exported count**

Jamf Pro doesn't enforce uniqueness on a package record's `filename`, so two package records can point to the same underlying file. Both get appended to `packageFilenames`, both pass the JCDS-set check, and both goroutines download to the *same* `outPath`. The temp-file+rename in `jcdsStreamToFile` means there's no corruption (last writer wins), but the file is fetched twice and `downloaded++` runs twice — so the "N objects exported" summary over-counts one physical file.

**Suggested fix** — dedupe before downloading (dropping into the existing JCDS-membership loop):
```diff
- for _, name := range filenames {
+ seen := make(map[string]struct{}, len(filenames))
+ for _, name := range filenames {
+     if _, dup := seen[name]; dup {
+         continue
+     }
+     seen[name] = struct{}{}
      if _, onJCDS := jcdsSet[name]; !onJCDS {
```

**Fixed when:** a package list containing two records with the same `filename` results in exactly one download and a `downloaded` count of 1.

🟩 **(2) (code-quality, c=75) — `pro_backup.go:697`: Empty `packages/files/` directory left behind when every package is skipped**

When `--download-packages` is set and packages exist but none are on JCDS, `os.MkdirAll(filesDir, …)` runs before the skip loop, so an empty `packages/files/` directory is created even though nothing was written. (The `len(filenames) == 0` early-return already handles the no-packages case, so this only bites the all-skipped case.) Minor cosmetic clutter in the backup output.

**Suggested fix:** create the directory lazily on first successful download (e.g. guard the `MkdirAll` behind a `sync.Once` invoked inside the goroutine before writing), or skip it entirely when the JCDS-membership filter yields zero matches.

**Fixed when:** a `--download-packages` run where no package is on JCDS leaves no empty `packages/files/` directory.

🟩 **(3) (docs, c=75) — `docs/site/examples.json:76`: Consider a curated example for the new flag**

The site's curated `pro backup` examples cover `--resources`, `--format`, and `--include-ids` but not the new `--download-packages`. The flag's help text is clear and the reference command list auto-introspects flags from the binary, so this is purely discoverability polish — not required.

**Suggested fix:** add an entry under `"pro backup"` in `docs/site/examples.json`, e.g. `jamf-cli pro backup --output ./backup --resources packages --download-packages`.

**Fixed when:** the backup examples include a `--download-packages` usage (optional).

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `internal/commands/pro_backup.go` | 4/5 | New flag + `backupPackageFiles`. Correct concurrency (race-clean), correct partial-failure wiring. Minor edge cases in findings (1) and (2). |
| `internal/commands/pro_jcds.go` | 5/5 | Pure behavior-preserving extraction of `jcdsListFiles`; the `jcds sync` call site now reuses it. Verified byte-for-byte equivalent to the inlined code. |
| `internal/commands/pro_backup_test.go` | 5/5 | Five scenarios: happy path (content verified), non-JCDS skip, download-failure→PartialFailure exit, resource-filtered-out (no JCDS call), default-off (no files dir). Strong coverage. |

</details>

<details>

- **Lines reviewed:** full diff (320 additions / 30 deletions, 3 files).
- **Context read:** `runBackup` failure-aggregation and exit-code path (`pro_backup.go:124-365`), `jcdsDownloadFile`/`jcdsStreamToFile`/`jcdsFetchPresignedURL` (`pro_jcds.go:103-198`), the existing `jcds sync` concurrency loop (`pro_jcds.go:385-429`) for pattern consistency, `StripServerFields` (`pro_helpers.go:303`), and the packages resource definition (`pro_resources.go:86`).
- **Verified locally:** `go build ./...` ✓, new tests pass ✓, `-race` clean on `TestBackup_DownloadPackages*` ✓, `go vet ./internal/commands/` clean ✓.
- **Confirmed:** Go 1.26.4 (no loop-var capture hazard); packages resource is not `ListOnly` so `filename` is present in the fetched detail; `StripServerFields` does not strip `filename`; `SubDir: "packages"` matches the hardcoded `packages/files/` path.
</details>

<details>

- ✅ Reuses `jcdsDownloadFile` — inheriting its temp-file+rename and 403-retry-once logic — so a failed download never leaves a partial file on disk (verified by `TestBackup_DownloadPackages_DownloadFailure`), rather than reimplementing streaming.
- ✅ Treats "not on JCDS" as a warning, not a failure, so on-prem/third-party packages don't poison the partial-failure exit code — a deliberate, correct scoping decision.
</details>

<details>

- [x] Design and architecture: opt-in flag, reuses existing helpers, extraction shared cleanly between `jcds sync` and backup — fits the codebase's patterns
- [x] Correctness: traced failure aggregation → `_failures.yaml` → `PartialOrPropagate` exit code; verified `filename` data-flow (fetched in detail, survives strip); dedup edge case flagged (1)
- [x] Security: `filepath.Join` cleans paths; filenames are cross-checked against the server-provided JCDS set; backing up one's own instance — no new trust boundary vs. existing `jcds sync`/`download`
- [x] Performance: 4MB copy buffer inherited from helper; concurrency bounded by semaphore = clamped `--concurrency`; redundant-download inefficiency noted (1)
- [x] Test coverage: 5 scenarios incl. default-off and filtered-out negative paths; verified they pass + race-clean
- [x] Reliability: download failures recorded, don't crash the run; JCDS-list fetch failure records one failure and skips downloads gracefully
- [x] Documentation currency: new flag has clear help text; site auto-introspects flags; curated example is optional polish (3)
- [x] Project rules compliance: no credential flags added; no generated-code edits; changes are in the correct hand-written `pro_*.go` files per CLAUDE.md
- [x] Simplification: `errgroup` used only for WaitGroup semantics (returns always-nil) — mirrors existing `jcds sync`, so kept for consistency, not flagged

</details>

PR #275, opened 2026-07-09
Handle duplicate computer serial numbers (fix silent --serial resolution + surface duplicates)

### LINKED ISSUE #273: Feature Request: Add command for finding all the computers with the same serial number
If a Macs motherboard get replaced it shows up in Jamf Pro as a new computer record. You can't create a smart search to find all the computers with duplicate serial numbers. Can you please add a command to jamf-cli to do this as having duplicate serial numbers breaks a bunch of other jamf-cli commands.

neilmartin83:
Closes #273.

Investigating #273 confirmed both of the reporter's claims, and surfaced a latent data-loss bug worse than described. This PR fixes the bug and adds two ways to find duplicate serials.

## 1. Fix: `--serial`/`--name` lookups silently targeted the wrong record

Generated `get`/`update`/`patch` resolved `--name`/`--serial`/`--udid` via `resolveNameToID`, which fetched with `page-size=1` and blindly took `results[0]`. When two computer records share a serial (a replaced logic board re-enrolls as a fresh record — exactly the #273 scenario), it silently resolved to whichever record the API returned first:

- `pro computers get --serial <dup>` → returns an arbitrary one of the pair.
- `pro computers patch/update --serial <dup>` → **silently mutates the wrong device**, no error.

(`delete` and `apply` were already safe — they used the collision-detecting `resolveNameToIDForApply`.) `resolveNameToID` also built its RSQL filter from the un-escaped value, unlike the apply path.

**Fix + consolidation** (`generator/parser/generator.go`): both resolvers now share
- `lookupMatchingIDs` — fetches up to 100 candidates so collisions are visible, client-side filters endpoints that ignore RSQL, escapes the value, returns every matching ID;
- `pickMatchingID` — one match resolves; several error under `--no-input`, else prompt.

They now differ only in the no-match branch (error vs. return-for-create), so `get`/`patch`/`update`/`apply`/`delete` can't drift apart on duplicate handling. Regenerated across all resources.

## 2. Feature: surface duplicate serials

Smart groups can't find duplicates (criteria are per-record, not cross-record), so the CLI now can:

- **`pro audit`** — new hygiene check flagging that duplicates exist, with affected-record count (blank serials ignored). The signal.
- **`pro report duplicate-serials`** — lists the offending records (`serial`, `id`, `name`, `last_contact`), grouped by serial and ordered by numeric id so the stale and current records sit adjacent for cleanup. The actionable list #273 asked for. Works with `-o json/csv/yaml`.

## Tests
- `resolve_name_test.go` — collision→error, single→resolve, no-match behaviour for both resolvers, `lookupMatchingIDs` returns all duplicates.
- `pro_audit_test.go` — duplicate found (count, blanks excluded) / all-unique→nil.
- `pro_report_test.go` — grouping + numeric-id order + record detail / all-unique→empty / fetch error.

`make generate`, `go test ./...`, `make lint` (0 issues), and `make verify-generated` all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ See the [Setup Guide](https://github.com/Jamf-Concepts/jamf-cli/wiki/Setup-Guide
 - **`--field`** — Extract a single field from any response: `jamf-cli pro comp list --field id`
 - **`apply`** — Name-based upsert: creates if new, replaces if existing (with confirmation)
 - **`patch`** — JSON Merge Patch (RFC 7386): update individual fields without a full replace. Use `--set key=value` for scalar fields or pipe a merge-patch document. Accepts `--name`, `--serial`, `--udid` (resource-dependent) in place of an ID. `--scaffold` prints the patchable field template
-- **`--name` flag** — `get`, `update`, `delete`, and `patch` commands all accept `--name` (and resource-specific alternates like `--serial`, `--udid`) in place of a positional ID
+- **`--name` flag** — `get`, `update`, `delete`, and `patch` commands all accept `--name` (and resource-specific alternates like `--serial`, `--udid`) in place of a positional ID. When the value matches more than one record (e.g. two computers sharing a serial after a logic-board swap), the lookup fails with a usage error under `--no-input` (or a non-terminal stdin) and prompts interactively otherwise — it never silently picks one. Run `jamf-cli pro report duplicate-serials` to list the colliding records
 - **`--scaffold`** — Print JSON templates for create/update commands with example values
 - **Five output formats** — `table`, `json`, `csv`, `yaml`, `plain`
 - **Auto-pagination** — `--all` fetches every page; `--limit` caps results
--- a/generator/parser/generator.go
+++ b/generator/parser/generator.go
@@ -1902,13 +1902,15 @@ func new{{ $.GoName }}{{ toCamel .Name }}Cmd(ctx *registry.CLIContext) *cobra.Co
 			var resolvedPatchID string
 {{ range $.LookupFields }}
 			if flag{{ toCamel .Flag }} != "" {
-				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ .RSQLField }}", "{{ lookupIDField $ }}", flag{{ toCamel .Flag }})
+				noInput, _ := cmd.Flags().GetBool("no-input")
+				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ .RSQLField }}", "{{ lookupIDField $ }}", flag{{ toCamel .Flag }}, noInput)
 				if err != nil {
 					return fmt.Errorf("looking up --{{ .Flag }} %q: %w", flag{{ toCamel .Flag }}, err)
 				}
 				resolvedPatchID = rid
 			} else {{ end }}if flagName != "" {
-				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ $.NameField }}", "{{ lookupIDField $ }}", flagName)
+				noInput, _ := cmd.Flags().GetBool("no-input")
+				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ $.NameField }}", "{{ lookupIDField $ }}", flagName, noInput)
 				if err != nil {
 					return err
 				}
@@ -1939,7 +1941,8 @@ func new{{ $.GoName }}{{ toCamel .Name }}Cmd(ctx *registry.CLIContext) *cobra.Co
 					return fmt.Errorf("no {{ $.NameSingular }} found with --{{ .Flag }} %q", flag{{ toCamel .Flag }})
 				}
 {{- else }}
-				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ .RSQLField }}", "{{ lookupIDField $ }}", flag{{ toCamel .Flag }})
+				noInput, _ := cmd.Flags().GetBool("no-input")
+				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ .RSQLField }}", "{{ lookupIDField $ }}", flag{{ toCamel .Flag }}, noInput)
 				if err != nil {
 					return fmt.Errorf("looking up --{{ .Flag }} %q: %w", flag{{ toCamel .Flag }}, err)
 				}
@@ -1959,7 +1962,8 @@ func new{{ $.GoName }}{{ toCamel .Name }}Cmd(ctx *registry.CLIContext) *cobra.Co
 					return fmt.Errorf("no {{ $.NameSingular }} found with {{ $.NameField }} %q", flagName)
 				}
 {{- else }}
-				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ $.NameField }}", "{{ lookupIDField $ }}", flagName)
+				noInput, _ := cmd.Flags().GetBool("no-input")
+				rid, err := resolveNameToID(reqCtx, ctx.Client, "{{ nameLookupPath $ }}", "{{ $.NameField }}", "{{ lookupIDField $ }}", flagName, noInput)
 				if err != nil {
 					return err
 				}
@@ -2918,25 +2922,26 @@ func batchDeleteError(cmd *cobra.Command, succeeded, failed int, firstErr error,
 		fmt.Sprintf("%d of %d %s failed", failed, succeeded+failed, noun))
 }
 
-// resolveNameToID looks up a resource by name using a filtered list call and
-// returns its ID. This enables --name as an alternative to positional ID args
-// on get commands. The nameField parameter specifies the filter field
-// (usually "name", but some resources use "displayName"). The idField parameter
-// specifies which response property holds the identifier (usually "id", but some
-// resources use "templateId", "groupId", etc.).
-func resolveNameToID(ctx context.Context, client registry.HTTPClient, listPath, nameField, idField, name string) (string, error) {
-	filterPath := fmt.Sprintf("%s?filter=%s&page-size=1",
-		listPath, url.QueryEscape(fmt.Sprintf(` + "`" + `%s=="%s"` + "`" + `, nameField, name)))
+// lookupMatchingIDs looks up resources whose nameField equals name and returns
+// every matching resource's idField value. It fetches up to 100 candidates so
+// callers can detect duplicate-name/serial collisions (e.g. two computer records
+// sharing a serial after a logic-board swap), and re-fetches unfiltered when an
+// endpoint ignores the RSQL filter param (e.g. prestages). The name is escaped
+// to prevent RSQL injection. Returns an empty slice when nothing matches.
+func lookupMatchingIDs(ctx context.Context, client registry.HTTPClient, listPath, nameField, idField, name string) ([]string, error) {
+	escapedName := strings.ReplaceAll(name, ` + "`" + `"` + "`" + `, ` + "`" + `\"` + "`" + `)
+	filterPath := fmt.Sprintf("%s?filter=%s&page-size=100",
+		listPath, url.QueryEscape(fmt.Sprintf(` + "`" + `%s=="%s"` + "`" + `, nameField, escapedName)))
 
 	resp, err := client.Do(ctx, "GET", filterPath, nil)
 	if err != nil {
-		return "", fmt.Errorf("looking up %q: %w", name, err)
+		return nil, fmt.Errorf("looking up %q: %w", name, err)
 	}
 	defer func() { _ = resp.Body.Close() }()
 
-	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 8<<20))
 	if err != nil {
-		return "", fmt.Errorf("reading lookup response: %w", err)
+		return nil, fmt.Errorf("reading lookup response: %w", err)
 	}
 
 	// Paginated response: {"totalCount": N, "results": [{...}]}
@@ -2948,18 +2953,18 @@ func resolveNameToID(ctx context.Context, client registry.HTTPClient, listPath, 
 		// Some endpoints return plain arrays instead of paginated objects
 		var arr []json.RawMessage
 		if jsonErr := json.Unmarshal(body, &arr); jsonErr != nil {
-			return "", fmt.Errorf("parsing lookup response: %w", err)
+			return nil, fmt.Errorf("parsing lookup response: %w", err)
 		}
 		data.Results = arr
 		data.TotalCount = len(arr)
 	}
 
 	if data.TotalCount == 0 || len(data.Results) == 0 {
-		return "", fmt.Errorf("no resource found with name %q", name)
+		return nil, nil
 	}
 
 	// Client-side filter: some endpoints ignore RSQL filter params (e.g. prestages).
-	// Always verify the returned result actually matches the requested name.
+	// Always verify the returned results actually match the requested name.
 	results := filterResultsByName(data.Results, nameField, name)
 	if len(results) == 0 && data.TotalCount > len(data.Results) {
 		// RSQL was likely ignored and we only got the first page. Re-fetch all results.
@@ -2976,19 +2981,61 @@ func resolveNameToID(ctx context.Context, client registry.HTTPClient, listPath, 
 			}
 		}
 	}
-	if len(results) == 0 {
+
+	ids := make([]string, 0, len(results))
+	for _, raw := range results {
+		var obj map[string]any
+		if err := json.Unmarshal(raw, &obj); err != nil {
+			continue
+		}
+		if id := extractIDString(obj, idField); id != "" {
+			ids = append(ids, id)
+		}
+	}
+	return ids, nil
+}
+
+// pickMatchingID collapses a set of lookup candidates to a single ID. One match
+// returns immediately. Several matches (duplicate names/serials) error under
+// --no-input, otherwise prompt the user to choose — so an ambiguous lookup can
+// never silently target the wrong record.
+func pickMatchingID(ids []string, name string, noInput bool) (string, error) {
+	if len(ids) == 1 {
+		return ids[0], nil
+	}
+	if noInput {
+		return "", fmt.Errorf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", "))
+	}
+	fmt.Fprintf(os.Stderr, "Multiple resources found matching %q:\n", name)
+	for i, id := range ids {
+		fmt.Fprintf(os.Stderr, "  [%d] ID: %s\n", i+1, id)
+	}
+	fmt.Fprintf(os.Stderr, "Enter number to select (or 0 to cancel): ")
+	var choice int
+	if _, err := fmt.Fscan(os.Stdin, &choice); err != nil || choice < 1 || choice > len(ids) {
+		return "", fmt.Errorf("aborted")
+	}
+	return ids[choice-1], nil
+}
+
+// resolveNameToID looks up a resource by name (or serial/udid) and returns its
+// single matching ID. This enables --name/--serial/--udid as an alternative to
+// positional ID args on get/update/patch commands. The nameField parameter
+// specifies the filter field (usually "name", but some resources use
+// "displayName"). The idField parameter specifies which response property holds
+// the identifier (usually "id", but some resources use "templateId", "groupId",
+// etc.). Errors when nothing matches, and — unlike a naive first-match lookup —
+// errors (or prompts) when several records share the value, so duplicate serial
+// numbers can never silently target the wrong record.
+func resolveNameToID(ctx context.Context, client registry.HTTPClient, listPath, nameField, idField, name string, noInput bool) (string, error) {
+	ids, err := lookupMatchingIDs(ctx, client, listPath, nameField, idField, name)
+	if err != nil {
+		return "", err
+	}
+	if len(ids) == 0 {
 		return "", fmt.Errorf("no resource found with name %q", name)
 	}
-
-	var first map[string]any
-	if err := json.Unmarshal(results[0], &first); err != nil {
-		return "", fmt.Errorf("parsing lookup result: %w", err)
-	}
-
-	if id := extractIDString(first, idField); id != "" {
-		return id, nil
-	}
-	return "", fmt.Errorf("resource %q has no %s field", name, idField)
+	return pickMatchingID(ids, name, noInput)
 }
 
 // extractIDString extracts the named identifier field from a JSON object as a string.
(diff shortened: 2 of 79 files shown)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Fixes a real data-loss bug (duplicate computer serials silently resolving `--serial`/`--name`/`--udid` lookups to the wrong record) by consolidating the resolver logic, and adds `pro audit`'s "Duplicate serial numbers" check plus a new `pro report duplicate-serials` command.
> Blocking: (1), (2), (3), (4), (5). See collapsed section for 4 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 once the interactive prompt fails fast on non-TTY stdin (1), the serial-lookup path requests `section=HARDWARE` so its own exact-match safety net isn't a no-op (2), the new ambiguous-collision error gets a distinct exit code per this repo's own agent contract (3), the RSQL escaping handles a literal backslash correctly (4), and `lookupMatchingIDs` stops silently dropping candidates it can't extract an ID from (5).

This PR does what it says: it closes a genuinely dangerous silent-wrong-record bug in `get`/`patch`/`update --serial`, and both new duplicate-serial discovery surfaces work as described. `go build`, `go vet`, `go test ./...`, `golangci-lint run`, and `make verify-generated` all pass clean against the PR head, and mutation testing confirms the core "collision → error/prompt" invariant is real, not just asserted. The five findings below are all in the surrounding correctness/reliability details, not the headline fix — but two of them (1, 2) sit squarely in the scenario this PR exists to make safe, so they're worth closing before merge rather than as follow-ups.

🟧 (1) (reliability, c=100) — `generator/parser/generator.go:3013-3020` (mirrored at `internal/commands/pro/generated/registry.go:296-305`, and identically in all ~150 `get`/`update`/`patch`/`resolveNameToIDForApply` call sites across the regenerated resources): `pickMatchingID` blocks on `os.Stdin` with no TTY guard, and this is now reachable from commands that never used to hang

```go
fmt.Fprintf(os.Stderr, "Enter number to select (or 0 to cancel): ")
var choice int
if _, err := fmt.Fscan(os.Stdin, &choice); err != nil || choice < 1 || choice > len(ids) {
    return "", fmt.Errorf("aborted")
}
```

`--no-input` is opt-in (defaults `false` — `internal/commands/root.go:780`), and this codebase has an established convention for exactly this situation: `internal/security/confirm.go:24`, `internal/platform/confirm.go:40`, and `internal/commands/multi.go:881` all check `term.IsTerminal(int(os.Stdin.Fd()))` before blocking on stdin, failing fast with an actionable error when it isn't a terminal. `pickMatchingID` skips that check entirely.

Before this PR, `get`/`update`/`patch --serial <dup>` never prompted at all — they silently took the first match. This PR is what makes those commands capable of reaching an interactive stdin read for the first time. A script or wrapper that predates this change and omits `--no-input` (the documented CI pattern `JAMF_CLI_ARGS='--quiet --no-input'` is opt-in, not default) will now hang indefinitely on an open-but-idle stdin (common under process supervisors, some CI runners, non-pty SSH) the moment it hits a duplicate serial — turning a data-safety fix into an availability foot-gun. Mutation testing confirms this: every test in `resolve_name_test.go` calls the resolvers with `noInput=true`, so the interactive branch has zero coverage and nothing would catch a regression here.

**Suggested fix:**
```diff
+	if !term.IsTerminal(int(os.Stdin.Fd())) {
+		return "", fmt.Errorf("multiple resources found matching %q (IDs: %s); re-run with --no-input for a clean error, or target a specific ID", name, strings.Join(ids, ", "))
+	}
 	fmt.Fprintf(os.Stderr, "Multiple resources found matching %q:\n", name)
```
(needs `"golang.org/x/term"` added to the generator template/generated package imports, matching `multi.go`.)

**Fixed when:** running the collision path with stdin redirected from a non-TTY and `--no-input` not set exits promptly with an actionable error instead of blocking on `fmt.Fscan`, verified by a test that feeds `pickMatchingID` a non-TTY path.

---

🟧 (2) (correctness, via devil-advocate, c=75) — `generator/parser/generator.go:2932-2996` (mirrored `registry.go:218-282`), call site `internal/commands/pro/generated/computers_inventory.go:249`: the `--serial` lookup never requests `section=HARDWARE`, so the client-side exact-match safety net is dead code specifically for the field this PR is about

`lookupMatchingIDs` is called with `listPath = "/v3/computers-inventory"` and no `section=` query param. Confirmed from this repo's own generated flag help (`computers_inventory.go:209`): *"section of computer details, if not specified, General section data is returned."* So every candidate JSON object returned for a `--serial` lookup lacks a `hardware` key entirely.

`filterResultsByName` (`registry.go:385-398`) is the client-side re-verification the PR relies on to catch a server-side RSQL filter that doesn't actually narrow to an exact match — its own doc comment says: *"When the field is absent from the response... the result is kept — the server's RSQL filter is trusted."* Since `hardware.serialNumber` is never present in the default-section response, `resolveNestedField` always returns `found=false`, so this function unconditionally keeps every candidate for serial lookups — it can never independently confirm a `hardware.serialNumber` match. Contrast with `--name` lookups: `general.name` *is* in the default section, so `filterResultsByName` genuinely verifies those.

In the current design this doesn't manifest for the "RSQL was ignored, re-fetch unfiltered" fallback branch (that branch only triggers when `filterResultsByName` returns zero results, which never happens here since "field absent" always keeps everything) — the practical risk is narrower: it means the CLI has zero independent check on whatever the server's RSQL filter returns for `hardware.serialNumber=="X"`. If that filter is ever not a strict exact match (tokenized/analyzed indexing, case handling, a server quirk) the client would trust it blindly, with no equivalent safety net to the one `--name` lookups get. Given this PR's own premise is "duplicate serial data exists in the wild," verifying the exact-match guarantee actually holds for the one field this PR is centered on seems worth the one-line fix.

**Suggested fix:**
```diff
-	rid, err := resolveNameToID(reqCtx, ctx.Client, "/v3/computers-inventory", "hardware.serialNumber", "id", flagSerial, noInput)
+	rid, err := resolveNameToID(reqCtx, ctx.Client, "/v3/computers-inventory?section=HARDWARE", "hardware.serialNumber", "id", flagSerial, noInput)
```
(in the generator template's serial-lookup fields — `lookupMatchingIDs` already handles a `?`-containing `listPath` correctly elsewhere via its existing query-string construction — verify and apply consistently to every `--serial`/`--udid` lookup path that filters on a non-default section.)

**Fixed when:** a test drives `lookupMatchingIDs` for a `hardware.serialNumber` lookup with a synthetic response that includes the `hardware` section, and asserts `filterResultsByName` actually narrows on a mismatched value instead of unconditionally keeping it.

---

🟧 (3) (design, via devil-advocate + usability-reviewer, c=100) — `generator/parser/generator.go:3006-3008` (mirrored `registry.go:293-295`) vs. `internal/commands/agent_context.md:24-38`: the new ambiguous-collision error isn't classified in this repo's own exit-code contract for agents

```go
return "", fmt.Errorf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", "))
```

This repo ships an explicit machine-consumer contract (`agent_context.md`) stating: *"React to a non-zero exit without parsing the message"*, backed by a table mapping codes 2–7 to specific agent actions (`usage`, `not_found`, `permission_denied`, etc.). A plain `fmt.Errorf` isn't an `*exitcode.Error`, so `exitcode.CodeFrom` (`internal/exitcode/exitcode.go:82-88`) falls back to `General` (1) — "unclassified failure." An agent or script driving this CLI exactly as the contract instructs sees exit code 1 for a duplicate-serial collision, indistinguishable from a network blip or a JSON parse error, even though this is a well-defined, highly actionable condition this PR introduces as a first-class new failure mode.

**Suggested fix:**
```diff
-		return "", fmt.Errorf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", "))
+		return "", exitcode.New(exitcode.Usage, fmt.Sprintf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", ")))
```
and add a row/bullet to `agent_context.md` documenting this new failure mode and the recommended agent response (e.g. re-run against a specific ID, or query `pro report duplicate-serials`).

**Fixed when:** the collision error under `--no-input` carries a distinct, documented exit code (not the generic `General`/1 fallback), verified by a test asserting `exitcode.CodeFrom(err)` on the collision path, and `agent_context.md` describes the new failure mode.

---

🟧 (4) (security, c=75) — `generator/parser/generator.go:2932-2933` (mirrored `registry.go:219`): RSQL escaping neutralizes literal quotes but not backslashes, so a backslash adjacent to a quote can still break out of the filter clause

```go
escapedName := strings.ReplaceAll(name, `"`, `\"`)
filterPath := fmt.Sprintf("%s?filter=%s&page-size=100",
    listPath, url.QueryEscape(fmt.Sprintf(`%s=="%s"`, nameField, escapedName)))
```

This is the classic "escape the escape character last" ordering bug: `ReplaceAll` only inserts a `\` in front of existing `"` characters — it never escapes a `\` that's already in the input. For `name` ending in a bare backslash (e.g. `x\`), `escapedName` is unchanged, so the emitted fragment is `...=="x\"` — a backslash immediately followed by the template's own closing quote. If the server's RSQL grammar treats `\"` as an escaped literal quote (standard RSQL string-literal escaping), the string is left unterminated instead of closed. Mutation testing confirms the test suite can't catch this class of bug at all: `resolveNameMockClient.Do` (`resolve_name_test.go:20`) discards the `path` argument entirely, so removing the escaping line altogether still passes every existing test.

This is real progress over the pre-PR `resolveNameToID` (which had *zero* escaping), and the practical blast radius is bounded — the value comes from the CLI operator's own `--serial`/`--name`/`--udid` flag, not a remote multi-tenant input — but it's the exact class of bug the PR's own doc comment claims to have closed ("The name is escaped to prevent RSQL injection").

**Suggested fix:**
```diff
-	escapedName := strings.ReplaceAll(name, `"`, `\"`)
+	// Escape backslashes first — otherwise a name containing `\"` can
+	// neutralize the quote-escaping below.
+	escapedName := strings.ReplaceAll(name, `\`, `\\`)
+	escapedName = strings.ReplaceAll(escapedName, `"`, `\"`)
```
Apply in the generator template and regenerate. Also have `resolveNameMockClient.Do` capture its `path` argument and add a test with a name containing `\"` asserting the constructed filter is fully neutralized.

**Fixed when:** `lookupMatchingIDs` escapes `\` before `"`, and a test fails if either escaping line is removed.

---

🟧 (5) (correctness, via silent-failure-hunter, c=75) — `generator/parser/generator.go:2985-2995` (mirrored `registry.go:272-282`): `lookupMatchingIDs` silently drops a candidate whose ID can't be extracted, instead of erroring as the pre-PR code did

```go
ids := make([]string, 0, len(results))
for _, raw := range results {
    var obj map[string]any
    if err := json.Unmarshal(raw, &obj); err != nil {
        continue
    }
    if id := extractIDString(obj, idField); id != "" {
        ids = append(ids, id)
    }
}
return ids, nil
```

`extractIDString` (`registry.go:330-339`) returns `""` for anything that isn't a JSON string or `float64` — a missing/`null` id field included. Before this PR, the single-result code path explicitly errored on this (`return "", fmt.Errorf("resource %q has no %s field", ...)` — visible in this PR's own diff as removed code). Now, a candidate with an unextractable ID just vanishes from `ids` with no signal at all.

Concrete scenario: two computer records genuinely share a serial (this PR's own motivating case), but one of them — plausibly a just-re-enrolled record that hasn't completed a full sync — has a malformed/absent `id` in the response. `lookupMatchingIDs` returns `ids = ["42"]` (length 1) instead of `["42", "88"]`; `pickMatchingID` sees a single candidate and resolves silently, with zero indication a second colliding record was excluded. That's the exact silent-wrong-resolution failure mode this PR's own doc comment claims is now impossible ("errors... so duplicate serial numbers can never silently target the wrong record") — the guarantee only holds when every candidate's ID happens to be extractable.

**Suggested fix:**
```diff
 ids := make([]string, 0, len(results))
+var skipped int
 for _, raw := range results {
     var obj map[string]any
     if err := json.Unmarshal(raw, &obj); err != nil {
+        skipped++
         continue
     }
-    if id := extractIDString(obj, idField); id != "" {
-        ids = append(ids, id)
+    id := extractIDString(obj, idField)
+    if id == "" {
+        skipped++
+        continue
     }
+    ids = append(ids, id)
 }
+if skipped > 0 {
+    return ids, fmt.Errorf("%d matching record(s) could not be resolved to an ID and were excluded from collision detection", skipped)
+}
 return ids, nil
```

**Fixed when:** a candidate that matches the filter but has an unextractable ID causes `lookupMatchingIDs` to surface that fact rather than silently shrinking the candidate set, verified by a test with one well-formed and one malformed-ID candidate in the same response.

<details>

🟩 (6) (usability, c=75) — `generator/parser/generator.go:3007` and `internal/commands/pro_audit.go:436`: neither the collision error nor the audit's recommendation text points to the new `pro report duplicate-serials` command

The collision error (`"multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID"`) and the audit's `Recommendation` (`"%d serial(s) map to multiple computer records...; delete the stale record for each..."`) both tell the user duplicates exist without naming the command this PR just built specifically to show them which records those are.

**Suggested fix:** name `pro report duplicate-serials` in both messages, e.g. `"...; run 'pro report duplicate-serials' to see all affected records, or target a specific ID"`.

**Fixed when:** both messages point the user at `pro report duplicate-serials`.

🟩 (7) (usability, c=75) — `generator/parser/generator.go:2999-3020`: prompt UX regressed slightly in the consolidation

Two small, non-blocking UX points from the shared prompt: (a) `"Enter number to select (or 0 to cancel): "` treats *any* invalid input (typo, out-of-range, EOF) identically to a deliberate `0` — safe by construction (the range check gates indexing) but gives no "invalid selection, try again" feedback; (b) the pre-existing `apply`-flow prompt said `"Enter number to replace"`, correctly signaling the destructive nature of that specific flow — the new shared wording always says "select," which is accurate for `get`/`patch` but a small loss of proactive clarity for `apply` (partially mitigated by `apply`'s separate downstream "will be replaced, type yes" confirmation before the actual write).

**Suggested fix:** distinguish `choice==0` from out-of-range in the error message, and thread a `verb` parameter (`"select"` vs `"replace"`) through `pickMatchingID` from its two callers.

**Fixed when:** author's call — not required for correctness.

🟩 (8) (test-coverage, via test-quality-reviewer, c=75) — `internal/commands/pro_audit.go:412`, `internal/commands/pro_report_serials.go:68,117`: a few edge cases in the new duplicate-serial logic are unverified (mutation-confirmed)

Bounded mutation testing found three survivors with low real-world impact: removing `strings.TrimSpace(serial)` in both `checkDuplicateSerials` and `runReportDuplicateSerials` doesn't fail any existing test (no fixture has two structurally-different-but-equivalent-after-trim whitespace serials); flipping `idLess`'s non-numeric fallback comparison (`a < b` → `b < a`) doesn't fail any test (every fixture uses purely numeric IDs); and `checkDuplicateSerials` has no dedicated fetch-error test, unlike its sibling `runReportDuplicateSerials`.

**Suggested fix:** add a whitespace-collision fixture, a mixed numeric/non-numeric `idLess` case, and a `TestCheckDuplicateSerials_FetchError` mirroring the report's version.

**Fixed when:** author's call — all three are low-likelihood-in-practice given current Jamf Pro data shapes.

🟩 (9) (documentation, c=75) — no README or CHANGELOG entry documents the `--name`/`--serial`/`--udid` collision behavior change

`README.md:150` documents that `get`/`update`/`delete`/`patch` accept `--name`/`--serial`/`--udid` in place of an ID, but says nothing about the new collision handling this PR introduces (hard error under `--no-input`, interactive prompt otherwise) — and there's no `CHANGELOG.md` anywhere in the repo to record it as a behavior change. Existing scripts/wrappers that predate this change and hit a duplicate will now fail (or, per (1), hang) with nothing in the docs explaining why.

**Suggested fix:** add a line under the `--name` flag entry in `README.md:150` noting the new collision behavior and pointing to `pro report duplicate-serials`.

**Fixed when:** README states that ambiguous lookups now fail/prompt instead of silently picking a match.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/parser/generator.go` | 2/5 | Template source of the resolver consolidation; carries the root cause of (1)-(5) |
| `internal/commands/pro/generated/registry.go` | 2/5 | Regenerated output of the same template — identical issues ship here (confirmed via `make verify-generated`) |
| `internal/commands/pro/generated/resolve_name_test.go` | 3/5 | Solid collision/no-match/single-match coverage, but doesn't verify RSQL escaping (4) or exercise the interactive prompt branch (1) |
| `internal/commands/pro_audit.go` | 4/5 | Clean, well-tested new check; recommendation text doesn't point to the new report command (6) |
| `internal/commands/pro_report_serials.go` | 4/5 | Clean, well-tested new command; minor untested edge cases (8) |

</details>

<details>

- Read the full diff (75 files, 1006/-437 lines) via `git diff origin/main...origin/feat/issue-273-duplicate-serials`
- Deep-read: `generator/parser/generator.go` (the generator template — source of truth per this repo's CLAUDE.md), `internal/commands/pro/generated/registry.go` (the actual shipped resolver code), `internal/commands/pro/generated/resolve_name_test.go`, `internal/commands/pro_audit.go` + test, `internal/commands/pro_report_serials.go` + test, `internal/commands/pro_report.go`
- Spot-checked several of the 66 mechanically-regenerated `internal/commands/pro/generated/*.go` files (e.g. `computers_inventory.go`) for call-site consistency with the template
- Ran `go build ./...`, `go vet ./...`, `go test ./...`, `golangci-lint run ./...`, and `make verify-generated` directly against the PR head — all clean
- Read `internal/commands/agent_context.md`, `README.md`, `CLAUDE.md`, and codebase conventions (`internal/security/confirm.go`, `internal/platform/confirm.go`, `internal/commands/multi.go`) for context and precedent
- Read linked issue [#273](https://github.com/Jamf-Concepts/jamf-cli/issues/273) for the original feature request
- Dispatched 6 specialist agents given the diff size and exported-signature changes (security, devil's-advocate, silent-failure-hunter, test-quality [ran real bounded mutation tests], usability, fidelity-vs-issue-#273); all findings above were independently re-verified against the source before inclusion
- Did not deep-review: the Classic API generator/registry (untouched, architecturally different — no client-side collision bug class to fix there), or the 66 mechanically-identical generated files beyond spot-checks

</details>

<details>

✅ Consolidating `resolveNameToID` and `resolveNameToIDForApply` onto one `lookupMatchingIDs`/`pickMatchingID` core is the right structural call — it closes the exact gap that caused this bug (the two resolvers drifting apart on collision handling and escaping) across all ~150 call sites at once, instead of patching `resolveNameToID` in isolation and leaving the next refactor free to drift again.

✅ The PR's central claim — "duplicate serial numbers can never silently target the wrong record" — is backed by real evidence, not just tests that exercise the happy path. Bounded mutation testing confirmed both the `len(ids)==1` collapse and the `n>1` duplicate-count threshold are genuinely caught by the test suite if broken.

</details>

<details>
<summary>Acceptance criteria checklist</summary>

| # | Criterion | Status | Notes |
|---|-----------|--------|-------|
| 1 | Fix: consolidate `resolveNameToID`/`resolveNameToIDForApply` on shared `lookupMatchingIDs`/`pickMatchingID`, both error/prompt on collision + escape RSQL, regenerated across all affected resources | ✅ Met (see (2), (4), (5) for gaps in the guarantee) | `registry.go:218-326,684-693`; template updated identically; `go build`/`go vet`/`make verify-generated` all clean |
| 2 | `delete`/`apply` already safe (used collision-detecting `resolveNameToIDForApply`) | ✅ Met | Confirmed pre-PR `resolveNameToIDForApply` already had page-size=100, escaping, and collision handling; 100% of `delete`/`apply`/`erase` call sites use `ForApply`, before and after |
| 3 | Feature: `pro audit` "Duplicate serial numbers" hygiene check | ✅ Met | `pro_audit.go:391-437`; wired into `allAuditChecks()`; blank serials ignored |
| 4 | Feature: `pro report duplicate-serials` | ✅ Met | `pro_report_serials.go`; grouped by serial, numeric-ID order, `-o json/csv/yaml` |
| 5 | Tests: resolve_name_test.go, pro_audit_test.go, pro_report_test.go | ✅ Met | All claimed sub-cases present and accurate |
| 6 | `make generate`/`go test ./...`/`make lint`/`make verify-generated` all pass | ✅ Met | Independently re-run: all clean |

</details>

<details>

- [x] Design and architecture: consolidation of two drifted resolvers onto one shared core is sound; verified no other call site (Classic API, package-upload) was missed or left on the old vulnerable pattern
- [x] Correctness: traced `lookupMatchingIDs`/`pickMatchingID` collision logic in full, including the RSQL-ignored fallback branch and the section-omission gap; findings (2), (5)
- [x] Security: RSQL escaping order bug (4); reviewed stderr ID disclosure (no issue — operator already has this data via their own authenticated call) and the 8MB response cap bump (reasonable given page-size=100)
- [x] Performance: `checkDuplicateSerials`/`runReportDuplicateSerials`'s full-inventory `FetchAllPaginated` scans are consistent with every other `pro audit`/`pro report` check in this codebase (no new pattern, no cap needed beyond what already exists)
- [x] Test coverage: ran real bounded mutation testing (via test-quality-reviewer) against every new/changed function; findings (4), (8); confirmed the PR's core invariant is genuinely tested, not just asserted
- [x] Reliability: TTY-guard omission (1) traced against established codebase convention and confirmed newly-reachable via this PR for get/update/patch; exit-code classification gap (3) traced against the explicit `agent_context.md` contract
- [x] Code quality: naming and structure consistent with surrounding generated-code conventions; no doc/code mismatches beyond the doc-comment guarantee in (5)/(2) overstating what's actually verified
- [x] Simplification: no redundant abstractions; the shared helper reduces duplication versus the pre-PR two-copies-of-similar-logic
- [na] Frontend concerns: CLI-only, no UI
- [x] Documentation currency: checked README/CLAUDE.md/agent_context.md for the new commands and the behavior change; finding (9)
- [na] Project rules compliance: no `.claude/rules/` directory in this repo

</details>

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Fixes a real data-loss bug (duplicate computer serials silently resolving `--serial`/`--name`/`--udid` lookups to the wrong record) and adds `pro audit`'s "Duplicate serial numbers" check plus a new `pro report duplicate-serials` command. All 5 blocking findings from the prior review are fixed, each with a targeted regression test.

## Rating: 5/5
- All 5 blocking findings are genuinely fixed — verified independently (fresh `go build`/`go vet`/`go test ./...`/`golangci-lint`/`make verify-generated` at the new head, all clean) rather than taken on the commit message's word. Only a non-blocking UX nit ((7)) and a half-addressed nice-to-have ((6)) remain, neither worth holding this.

`0282c1f` ("fix(generator): harden duplicate-serial resolution (PR #275 review)") lands a genuinely thorough response: every important finding got a purpose-built regression test that exercises the exact scenario the finding described, not just a happy-path check. `TestPickMatchingID_NonTTYFailsFast` drives a real `os.Pipe()` to prove the TTY guard fires; `TestLookupMatchingIDs_EscapesBackslashAndQuote` decodes the actual constructed RSQL filter string rather than just checking the call succeeds; `TestFilterResultsByName_NarrowsOnHardwareSection` proves the client-side safety net can now genuinely narrow on `hardware.serialNumber` instead of trusting the server unconditionally. The `?section=HARDWARE` fix was also applied comprehensively — to every `--serial` call site in `computers_inventory.go` (get, delete, patch, upload, download, filevault-by-id, view-device-lock-pin, view-recovery-lock-password), not just the ones the original PR description called out.

## Prior findings status

| (1) | `generator/parser/generator.go` (pickMatchingID) | ✅ Fixed | `term.IsTerminal` check added; fails fast with `exitcode.Usage` on non-TTY stdin or `--no-input`, prompts only on a real terminal. Tested via `TestPickMatchingID_NonTTYFailsFast` (real `os.Pipe()`) |
| (2) | `generator/parser/generator.go`, `computers_inventory.go` | ✅ Fixed | New `LookupField.Section` + `lookupFieldPath` template helper appends `?section=HARDWARE`; applied to all 9 `--serial` call sites in `computers_inventory.go`. Tested via `TestFilterResultsByName_NarrowsOnHardwareSection` + `TestLookupMatchingIDs_SectionQueryJoin` (verifies `&` vs `?` joining) |
| (3) | `generator/parser/generator.go` (pickMatchingID) | ✅ Fixed | Collision error now `exitcode.New(exitcode.Usage, ...)`; `agent_context.md` documents the new failure mode. Tested via `exitcode.CodeFrom(err)` assertions in two new tests |
| (4) | `generator/parser/generator.go` (`lookupMatchingIDs`) | ✅ Fixed | Backslash now escaped before quote. Tested via `TestLookupMatchingIDs_EscapesBackslashAndQuote`, which decodes the actual constructed filter string and asserts the exact escaped sequence |
| (5) | `generator/parser/generator.go` (`lookupMatchingIDs`) | ✅ Fixed | Unextractable-ID candidates now increment a `skipped` counter and surface an error instead of silently shrinking the candidate set. Tested via `TestLookupMatchingIDs_UnextractableIDErrors` (one well-formed + one malformed candidate) |
| (6) | `pro_audit.go` / `registry.go` (collision error) | ⚠️ Partial | Audit `Recommendation` now names `pro report duplicate-serials` — but the collision error text in `pickMatchingID`/`resolveNameToID` itself is unchanged (`"...resolve the duplicates or target a specific ID"`, no pointer to the report command) |
| (7) | `generator/parser/generator.go` (prompt UX) | 🔴 Open | No change — "0 to cancel" is still indistinguishable from invalid input, and the prompt still says "select" for both `get`/`patch` and the `apply`/replace flow. Explicitly marked lowest-priority/author's-call in the original review; fine to leave as a follow-up |
| (8) | test coverage (TrimSpace/idLess/fetch-error) | ✅ Fixed | `TestRunReportDuplicateSerials_WhitespaceCollision`, `TestIDLess` (includes a mixed numeric/non-numeric case), and `TestCheckDuplicateSerials_FetchError` all added |
| (9) | README/CHANGELOG | ✅ Fixed | `README.md:150` now documents the collision behavior and points to `pro report duplicate-serials` |

## Active findings

🟩 (1) (usability, c=75) — `internal/commands/pro/generated/registry.go:319`: the collision error itself still doesn't point to `pro report duplicate-serials`, only the audit recommendation does

```go
return "", exitcode.New(exitcode.Usage, fmt.Sprintf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", ")))
```

This is the error a user or script actually hits when running `get`/`patch`/`update --serial <dup>` directly (not via `pro audit`), and it still doesn't name the command built specifically to show them which records collide.

**Suggested fix:**
```diff
-		return "", exitcode.New(exitcode.Usage, fmt.Sprintf("multiple resources found matching %q (IDs: %s); resolve the duplicates or target a specific ID", name, strings.Join(ids, ", ")))
+		return "", exitcode.New(exitcode.Usage, fmt.Sprintf("multiple resources found matching %q (IDs: %s); run `pro report duplicate-serials` to see all affected records, or target a specific ID", name, strings.Join(ids, ", ")))
```

**Fixed when:** the collision error text names `pro report duplicate-serials`.

🟩 (2) (usability, c=75) — `generator/parser/generator.go` (`pickMatchingID` prompt): "0 to cancel" still indistinguishable from a typo, and the prompt still says "select" for the `apply`/replace flow

Unchanged from the original review — both points were explicitly marked non-blocking/author's-call: (a) any invalid input (not just literal `0`) produces the same generic `"aborted"` error, and (b) the shared prompt wording lost the "replace" framing the pre-existing `apply`-flow prompt had (partially mitigated by `apply`'s separate downstream "will be replaced, type yes" confirmation).

**Fixed when:** author's call — not required for correctness, fine as a follow-up.

<details>

- Located the incremental fix commit `0282c1f` ("fix(generator): harden duplicate-serial resolution (PR #275 review)") via `git diff <old-head>..origin/<head>` where `<old-head>` is the merge commit (`16fda16`) my prior review's `commit_id` pointed to
- Read the full diff of `0282c1f` (11 files, +293/-35): `generator/parser/generator.go`, `generator/parser/parser.go`, `generator/parser/types.go`, `internal/commands/agent_context.md`, `internal/commands/pro/generated/computers_inventory.go`, `internal/commands/pro/generated/registry.go`, `internal/commands/pro/generated/resolve_name_test.go`, `internal/commands/pro_audit.go`, `internal/commands/pro_audit_test.go`, `internal/commands/pro_report_test.go`, `README.md`
- Verified each of the 9 prior findings against the actual code (not the commit message) with exact citations
- Independently re-ran `go build ./...`, `go vet ./...`, `go test ./internal/commands/... ./generator/...`, `golangci-lint run ./...`, and `make verify-generated` against the new head — all clean
- Checked for new comments/replies since the prior review (none — the author responded via code, not discussion) and confirmed no new issues introduced by the fix commit
- Did not re-run the full specialist-agent dispatch — the fix commit is narrowly scoped to addressing the prior review's exact findings, with no new high-risk surface introduced

</details>

PR #279, opened 2026-07-10
feat(app-installers): deployment export, history, version-update, per-computer + bulk retry (--all)

neilmartin83:
## What

Surfaces five previously-missing App Installer **deployment** endpoints. All were verified to exist on the live backend before adding (see *How verified*).

New `pro app-installer-deployments` commands:

| Command | Endpoint |
|---|---|
| `export` | `POST /v1/app-installers/deployments/export` (CSV) |
| `history` | `GET /v1/app-installers/deployments/{id}/history` |
| `add-history-note` | `POST /v1/app-installers/deployments/{id}/history` |
| `version-update` | `POST /v1/app-installers/deployments/{id}/version-update` |
| `installation-retry-by-computer-id` | `POST /v1/app-installers/deployments/{id}/computers/{computerId}/installation-retry` |
| `installation-retry --all` | `POST /v1/app-installers/deployments/computers/installation-retry` (bulk, no `{id}`) |

`installation-retry --all` makes a **single server-side call** to the collection-level endpoint (not a client-side loop); it's mutually exclusive with `<id>`/`--name`.

## Generator changes

- **`pairCollectionBulkActions`** (`parser.go`, runs before `disambiguateSameTerminalOps`): pairs a single-`{id}` x-action with a same-method collection-level sibling at the path minus `{id}`, records `BulkActionPath` on the per-`{id}` op (template emits `--all`), and drops the bulk op so it doesn't spawn a duplicate command. Guards — same method, per-`{id}` path ends in a literal verb (not `}`), exactly one path param — prevent mis-pairing CRUD ops that specs mis-tag with `x-action` (e.g. `PUT /accounts/{id}`).
- **`opHasNameLookup` now requires exactly one path param.** Multi-param actions fall to positional `ExactArgs(N)` instead of the single-id name-lookup path.

### Incidental bug fix (drives most of the diff outside app-installers)

The `opHasNameLookup` change **fixes five pre-existing broken commands** whose multi-param paths were routed through the single-id codepath, leaving one path param unsubstituted (literal `{…}`) in the request URL:

- `computers-inventory download <id> <attachmentId>`
- `notifications delete <id> <type>`
- team-viewer `close` / `resend-notification` / `sessions-status <configurationId> <sessionId>`

## How verified

Live-probed against a Jamf Pro sandbox. Existence confirmed via the method-sensitivity oracle (the modern API returns an identical generic 404 for unknown routes *and* legit not-found, but 405 on a wrong method proves a route exists; control paths returned 404). Every command routes correctly on the wire; non-2xx responses are server-state (e.g. no failed installs to retry, already-latest version).

## Spec note

`specs/AppInstallerDeployments.yaml` is a monolith **PreservedSpec**, so these hand-added paths won't be clobbered by `make sync-spec`.

## ⚠️ Overlaps with open PRs (needs rebase, no logical conflict)

- **#275** (duplicate-serials): also edits `generator/parser/generator.go` and regenerates several of the same files (`computers_inventory.go`, `app_installer_deployments.go`, `team_viewer_remote_administrations.go`, `registry.go`). The source edits are in different functions/regions and the features are orthogonal; generated-file conflicts are mechanical — whoever merges second should rebase and run `make generate`.
- **#278** (CLAUDE.md docs): both edit `CLAUDE.md` in different sections; trivial.

## Review follow-up (blocking findings 1–4 addressed)

- **(1) `--all` confirmation gate** — the collection-level bulk action now requires explicit confirmation (`--yes` or interactive prompt; errors under `--no-input` without `--yes`), matching the codebase convention for tenant-wide mutations. `flagYes` is declared/wired for non-destructive `BulkActionPath` ops.
- **(2) `flagName` coupling + guard tests** — the `--all` template's `--name` mutual-exclusion check is now gated behind `opHasNameLookup`, so a `BulkActionPath` op on a resource without name-lookup preconditions no longer emits an undefined `flagName`. Added regression tests isolating `pairCollectionBulkActions`' two guards (each fails when its specific guard is removed, verified by mutation) plus `TestGenerate_BulkActionWithoutNameLookup`.
- **(3) Multi-arg destructive confirmation** — dry-run/confirm messages now show every positional arg (`strings.Join`), so `notifications delete <id> <type>` no longer hides `<type>`.
- **(4) Binary-response example** — now renders every path-param placeholder (extracted `pathParamUsage` helper), so `computers-inventory download <id> <attachmentId>` is copy-pasteable.

Nice-to-haves (5) privilege-equivalence, (7) `export --out-file` (pre-existing, 6 resources), and (8) scope note are not addressed here — (7) warrants a separate follow-up against the shared template.

## Checks

`make build`, `make lint` (0 issues), `make test` (0 failures), `make verify-generated` all pass. New/updated tests: `TestPairCollectionBulkActions` (incl. mis-tagged-CRUD, cross-method, and the two isolated guard regressions), `TestStripParamSegments`, multi-param `TestOpHasNameLookup`, `TestGenerate_BulkActionAllFlag` (now also asserts the confirmation gate), `TestGenerate_BulkActionWithoutNameLookup`.


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Adds 5 App Installer deployment endpoints (export/history/version-update/per-computer + tenant-wide bulk retry) and fixes 5 pre-existing broken multi-param commands as a side effect of a generator correctness fix.
> Blocking: (1), (2), (3), (4). See collapsed section for 4 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 with a confirmation gate on the tenant-wide `--all` retry, the `flagName`/`opHasNameLookup` coupling hardened (or asserted) in the generator, the `notifications delete` confirmation message showing both identifying args, and the `download` example fixed for its new 2-arg signature — all small, surgical fixes.

**(1)** (design/reliability, c=100) — `internal/commands/pro/generated/app_installer_deployments.go:1074-1088`: `installation-retry --all` has no confirmation, dry-run, or cooldown despite tenant-wide blast radius

`isDestructiveAction` (`generator/parser/parser.go:2144`) only classifies `{delete, delete-multiple, erase, wipe, remove, lock, restart, shutdown, unmanage}` as destructive — `installation-retry` isn't in it, so `IsDestructive` is `false` and the generator template never emits `--yes`/`--dry-run`/`cooldown.Enforce` for this op (those are all gated behind `{{- if .IsDestructive }}`). The generated `--all` branch fires straight to `POST /v1/app-installers/deployments/computers/installation-retry` — a single call that reissues failed App Installer installs across **every deployment on the tenant** (potentially thousands of computers hitting distribution infrastructure) — with zero friction. This directly contradicts this codebase's own established convention for exactly this kind of blast radius: `internal/commands/pro_bulk.go`'s `bulk` command family states "Default behavior is a dry-run preview — no changes are made unless `--yes` is provided." A typo'd or scripted `installation-retry --all` re-triggers installs tenant-wide with no way to preview impact first.

**Suggested fix:**
```diff
 			if flagAll {
 				if len(args) > 0 || flagName != "" {
 					return fmt.Errorf("--all applies to every {{ $.NameSingular }}; do not combine it with an <id> or --name")
 				}
+				if !flagYes {
+					return fmt.Errorf("--all retries installs across every {{ $.NameSingular }}; pass --yes to confirm")
+				}
 				resp, err := ctx.Client.Do(reqCtx, "{{ .Method }}", "{{ .BulkActionPath }}", nil)
```
(requires also declaring/wiring a `flagYes` for `BulkActionPath` ops, following the same pattern as `.IsDestructive` ops.)

**Fixed when:** `--all` on a `BulkActionPath` command requires explicit confirmation (`--yes` or a dry-run preview), consistent with `pro_bulk.go`'s stated convention for tenant-wide mutations.

---

**(2)** (correctness + test-coverage, c=100, verified by reproduction) — `generator/parser/parser.go:663-721` / `generator/parser/generator.go:1719-1722`: `pairCollectionBulkActions` doesn't verify the paired op will satisfy `opHasNameLookup`'s resource-level preconditions; two of its own safety guards have no regression test

`pairCollectionBulkActions` sets `BulkActionPath` based purely on path-shape + HTTP method. The `--all` template block it enables unconditionally references `flagName`:
```go
if flagAll {
    if len(args) > 0 || flagName != "" { ... }
```
but `flagName` is only declared when `opHasNameLookup(op, r)` is true — which also requires `!r.IsSingleton`, `hasResolvableID(r)`, and `collectionPath(r.Operations) != ""`. Nothing ties these together. I constructed a resource with a `BulkActionPath` op but no `list` op (so `collectionPath` returns `""`), ran `gen.Generate`, and the emitted file failed to compile with `undefined: flagName`. This doesn't affect this PR's own spec (App Installer Deployments has a full list/get/create/update skeleton — verified via `make verify-generated`, `go build`, and `go test`, all clean), but it's a real latent trap for the next spec that reuses this pairing convention on an action-only or singleton resource.

Separately, mutation-testing `pairCollectionBulkActions`'s two mis-pairing guards found both **untested**: deleting the `strings.HasSuffix(op.Path, "}")` check (blocks mis-tagged CRUD ops like `PUT /accounts/{id}` from pairing) and deleting the `strings.Count(op.Path, "{") != 1` check (blocks multi-param actions from claiming a bulk sibling) each leave the full test suite green. `TestPairCollectionBulkActions`'s "mis-tagged CRUD" sub-test happens to pass for the unrelated reason that its methods differ, not because the `HasSuffix` guard fired.

**Suggested fix:** assert the invariant where the resource is known (in `Generate`/the per-resource pass, not `pairCollectionBulkActions` itself, which only sees `[]*Operation`):
```go
if op.BulkActionPath != "" && !opHasNameLookup(op, r) {
    return nil, fmt.Errorf("resource %s: op %s has BulkActionPath but fails opHasNameLookup preconditions", r.Name, op.Name)
}
```
Plus add a case to `TestPairCollectionBulkActions` that removes the `HasSuffix`/`Count` guards' target scenario without the accompanying method/param-count difference, so each guard is exercised for the reason it exists.

**Fixed when:** the generator either asserts this invariant with a clear error, or the template stops depending on `flagName`'s conditional declaration; both survived guards have a test that fails when that specific guard (not an incidental method/count mismatch) is removed.

---

**(3)** (correctness/usability, c=100) — `generator/parser/generator.go:1901-1913`, generated in `internal/commands/pro/generated/notifications.go:87-99`: destructive-delete confirmation text drops the second identifying argument

Before this PR's `opHasNameLookup` fix, `notifications delete` (incorrectly) went through the name-lookup path and printed `Would delete notification %q (id: %s)`. Post-fix, multi-param destructive ops fall into the generic `{{- if and .IsDestructive (not (opHasNameLookup . $)) }}` template branch, which was written for single-arg ops and only ever interpolates `args[0]`:
```go
fmt.Fprintf(os.Stderr, "Would delete resource %s\n", args[0])
...
fmt.Fprintf(os.Stderr, "⚠️  This will delete resource %s. Type 'yes' to confirm: ", args[0])
```
`notifications delete` now requires **two** args (`<id> <type>`) to uniquely identify what's being deleted, but the confirmation prompt the user reads before typing "yes" only shows the `id`, silently omitting `type`. This is a safety-relevant regression on a destructive action's own confirmation text, newly exposed because this is the first multi-param op ever routed through this branch (previously `opHasNameLookup` swallowed all `hasPathParam`-true cases, so no destructive op reached this fallback with more than one arg).

**Suggested fix:**
```diff
-				fmt.Fprintf(os.Stderr, "Would {{ .Name }}{{ if hasPathParam .Path }} resource %s{{ end }}\n"{{ if hasPathParam .Path }}, args[0]{{ end }})
+				fmt.Fprintf(os.Stderr, "Would {{ .Name }}{{ if hasPathParam .Path }} resource %s{{ end }}\n"{{ if hasPathParam .Path }}, strings.Join(args, " "){{ end }})
```
(same for the `--yes` prompt line.)

**Fixed when:** the dry-run and confirmation messages for a multi-param destructive op show every positional argument the user is about to act on, not just the first.

---

**(4)** (usability, c=100) — `internal/commands/pro/generated/computers_inventory.go:1102-1106`: `download`'s example text wasn't updated for its new 2-arg signature

`Use`/`Args` were correctly fixed to `download <id> <attachmentId>` / `cobra.ExactArgs(2)`, but the `Example` block still shows the old, broken single-`<id>` invocation:
```
Example: `  # Save to file
  jamf-cli pro computers-inventory download <id> -O output.bin

  # Pipe to stdout
  jamf-cli pro computers-inventory download <id> > output.bin`,
```
Copy-pasting either line fails with `Error: accepts 2 arg(s), received 1`. Root cause: `exampleText`'s generic `opHasBinaryResponse` branch (`generator.go:237-247`) emits a single ` <id>` placeholder whenever `hasPathParam(op.Path)` is true, regardless of path-param count — it wasn't updated alongside the `Args`/`Use` fix, and `download` is (after this PR) the only multi-param binary-response op in the codebase, so this is newly exposed here.

**Suggested fix:**
```diff
- if hasPathParam(op.Path) {
-     idArg = " <id>"
- }
+ for _, p := range pathParams(op.Parameters) {
+     idArg += fmt.Sprintf(" <%s>", p.Name)
+ }
```
**Fixed when:** the `download` example shows both `<id>` and `<attachmentId>` and running it as written succeeds.

<details>

🟩 (5) (design, c=75, via devil-advocate) — `generator/parser/parser.go:704-709`: `pairCollectionBulkActions` doesn't check that the paired ops require the same privileges

The pairing matches on path-shape + method only, never comparing `Privileges` between the per-`{id}` op and its bulk sibling before the bulk op's own `Privileges` are discarded (it's deleted from `ops`). The generated `--all` command's `jamf:privileges` annotation is emitted solely from the per-`{id}` op. In this spec both endpoints require "Update App Installers," so nothing is visibly wrong today, but a future spec where the bulk endpoint requires a broader privilege than its per-id sibling would generate a `--all` command whose help text and 403 hint advertise the wrong (narrower) privilege.

**Fixed when:** the pairing rule skips (or a test asserts) privilege equivalence between paired ops, or the generated `--all` command surfaces the bulk op's own privileges.

🟩 (6) (documentation accuracy, c=100) — PR description undercounts the generator fix's blast radius

The description lists 4 commands fixed by the `opHasNameLookup` change (`computers-inventory download`, `notifications delete`, team-viewer `close`/`resend-notification`), but the diff also fixes a 5th with the identical bug pattern: `team-viewer sessions-status` (`team_viewer_remote_administrations.go:598-624`). The fix itself is correct; the description just doesn't account for it — worth a quick note so a reader trusts the stated scope matches the actual diff.

🟩 (7) (usability, c=50, pre-existing/out-of-scope) — `internal/commands/pro/generated/app_installer_deployments.go:937-1006`: the new `export` command's documented `--out-file` example silently doesn't work

`--out-file` is the global persistent flag that only redirects `ctx.Output`'s formatter writer, but `export`'s `RunE` streams the response directly via `io.Copy(os.Stdout, resp.Body)` / `--save-to`, never touching `ctx.Output`. Following the command's own example (`export --out-file app-installer-deployments.csv`) creates/truncates the target file but writes nothing to it while the CSV silently goes to the terminal instead. This is a pre-existing generator-wide template bug (`generator.go`'s `case "export":` example) already present in 5 other resources (`buildings`, `packages`, `cloud_id_p_configurations`, `inventory_preloads`, `jamf_remote_assist_session_histories`) — this PR just adds a 6th instance by using the existing, unmodified template. Not blocking this PR; flagging for a follow-up ticket against the shared template.

🟩 (8) (scope, c=75, via scope-reviewer) — bundles a new feature with an unrelated 5-command bug fix, but the dependency is one-way, not incidental

The PR bundles the new app-installer-deployments feature with the `opHasNameLookup` fix affecting 5 pre-existing, unrelated commands. Worth noting for future PRs of this shape: it's not purely incidental bundling — the new `installation-retry-by-computer-id` command (2 path params) would hit the exact same `opHasNameLookup` defect if the fix weren't included, so the feature genuinely depends on the fix. Splitting would mean landing the fix first, then rebasing the feature onto it. Given the one-way dependency and that both halves are independently tested, shipping together here is defensible — flagging only so the pattern ("a fix required by new code ends up looking like unrelated scope") is recognized next time it appears without an author's self-disclosure.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/parser/parser.go` | 3/5 | `pairCollectionBulkActions` well-guarded against mis-tagging, but two guards untested and privilege equivalence unchecked (2, 5) |
| `generator/parser/generator.go` | 3/5 | `opHasNameLookup` multi-param fix is correct; the `--all` template's `flagName` coupling (2) and the generic destructive-fallback's single-arg confirmation message (3) are real gaps exposed by this PR's own changes |
| `internal/commands/pro/generated/app_installer_deployments.go` | 3/5 | New commands correctly wired to spec; `--all` lacks confirmation (1); inherited pre-existing `export --out-file` bug (7) |
| `internal/commands/pro/generated/computers_inventory.go` | 4/5 | Fixes a genuinely broken `download` command; stale example text (4) |
| `internal/commands/pro/generated/notifications.go` | 3/5 | Fixes a genuinely broken `delete` command; confirmation message regression (3) |
| `internal/commands/pro/generated/team_viewer_remote_administrations.go` | 4/5 | Fixes `close`/`resend-notification`/`sessions-status` correctly; PR description undercounts by one (6) |
| `generator/parser/generator_test.go` | 4/5 | Good multi-param coverage for `opHasNameLookup`; `TestGenerate_BulkActionAllFlag`'s fixture always satisfies `opHasNameLookup`'s preconditions so it can't catch (2) |
| `generator/parser/parser_test.go` | 3/5 | `TestPairCollectionBulkActions` covers method/path-shape mismatches but not the two guards mutation-tested to be silently removable (2) |
| `specs/AppInstallerDeployments.yaml` | 5/5 | Clean, matches the live-verified endpoints described in the PR |

</details>

<details>

- Read the full diff (13 files, 972+/242-) via `git diff origin/main...origin/feat/app-installer-deployment-endpoints`.
- Traced `pairCollectionBulkActions`, `stripParamSegments`, and the `opHasNameLookup` path-param-count change end-to-end against the generated `app_installer_deployments.go`, `computers_inventory.go`, `notifications.go`, and `team_viewer_remote_administrations.go` outputs.
- **Ran it, didn't just read it:** `go build ./...` (clean), `go test ./generator/... ./internal/commands/...` (all pass), `make lint` (0 issues), `make verify-generated` (regenerated from spec, zero diff against committed generated code), `make verify-site` (all products have full site support), manually exercised `--help` on the new action commands, and independently confirmed the `--out-file`/formatter bypass in `internal/commands/root.go`.
- Dispatched 4 specialist agents given this PR's high-risk classification (>500 lines changed, core generator file modified): `usability-reviewer` (found the stale example and confirmation-message regressions), `scope-reviewer` (assessed the feature+bugfix bundling), `test-quality-reviewer` (mutation-tested the new generator logic and reproduced the `flagName` compile failure by construction), and `devil-advocate` (surfaced the `--all` blast-radius gap and privilege-equivalence gap). Every specialist finding was independently re-verified against the actual source before inclusion here.
- Did not re-verify the PR's own "live-probed against a Jamf Pro sandbox" endpoint-existence claims — no sandbox access from this environment; the described verification methodology (405-vs-404 method-sensitivity oracle) is sound.

</details>

<details>

✅ The `pairCollectionBulkActions` guard against mis-tagged CRUD ops (a `PUT /accounts/{id}` that a spec incorrectly marks `x-action: true`, matching a `POST /accounts` create) is a sharp catch — the `strings.HasSuffix(op.Path, "}")` exclusion specifically prevents a subtle false-positive that would otherwise silently corrupt an unrelated resource's `update` command.

✅ The PR's own "How verified" section describes a legitimate live-verification methodology (405-vs-404 method-sensitivity oracle) for confirming undocumented endpoints exist before adding them — good practice for a hand-added spec that bypasses `make sync-spec`.

</details>

<details>

- [x] Design and architecture: `pairCollectionBulkActions` fits the existing "one pass transforms `[]*Operation` before disambiguation" style; found the missing tenant-wide safety gate (1) and privilege-equivalence gap (5)
- [x] Correctness: traced path-substitution logic in all 5 fixed multi-param commands against their specs; found the `flagName` coupling gap (2, reproduced by construction) and confirmation-message regression (3)
- [x] Security: no new auth/injection surface; the `--all` finding (1) is a blast-radius/confirmation gap, not an authz bypass — the endpoint itself still requires "Update App Installers"
- [na] Performance: no hot paths, loops, or new query patterns — single HTTP calls per command
- [x] Test coverage: ran the full suite (pass) and mutation-tested the new logic — found two untested guards (2)
- [x] Reliability: `--all` mutual-exclusivity error path is clear; the missing confirmation gate (1) is the main reliability/safety gap
- [na] Code quality: naming and structure consistent with the rest of `generator/parser/`
- [na] Simplification: no redundant abstractions introduced
- [na] Frontend concerns: N/A, this is a CLI generator
- [x] Documentation currency: `CLAUDE.md`'s "Generated Command Features" section was updated to describe the new `pairCollectionBulkActions`/`BulkActionPath`/`--all` convention
- [na] Project rules compliance: no `.claude/rules/` directory in this repo; `CLAUDE.md`'s "Generated Code Boundary" rule was respected (no hand-edits to `internal/commands/pro/generated/`)

</details>

ktn-jamf (dismissed):
> [!TIP]
> ✅ **Merge-ready** — All 4 blocking findings from the prior review are fixed and independently re-verified (build, full test suite, lint, `make verify-generated`, and manual runtime checks of the new confirmation gate and example text).

## Rating: 5/5
- Clean fixes: each blocking finding got a targeted change plus a regression test that specifically isolates the guard it protects, and the 7 other single-param resources mechanically touched by the confirmation-message fix are functionally unchanged (verified by diff).

## Prior findings status

| (1) | `app_installer_deployments.go` / `generator.go` (`--all` confirmation) | ✅ Fixed | `flagYes` + `--no-input`-aware prompt added, matching `pro_bulk.go`'s convention. Verified live: `installation-retry --all --no-input` now refuses with `"pass --yes to confirm when --no-input is set"` instead of firing. |
| (2) | `parser.go`/`generator.go` (`flagName`/`opHasNameLookup` coupling) | ✅ Fixed | The `--all` branch's `flagName != ""` check is now gated behind `{{ if opHasNameLookup . $ }}`, so a `BulkActionPath` op on a resource lacking name-lookup preconditions no longer references an undeclared `flagName`. New `TestGenerate_BulkActionWithoutNameLookup` reproduces exactly the scenario I used to trigger the compile failure and asserts no `flagName` reference; two new `TestPairCollectionBulkActions` sub-tests isolate the `HasSuffix`/`Count` guards I found untested (each now fails when its specific guard, not an incidental method/param mismatch, is removed). |
| (3) | `notifications.go` (destructive confirmation message) | ✅ Fixed | Both the dry-run and `--yes` prompt now use `strings.Join(args, " ")` instead of `args[0]`. Verified live: `notifications delete 1 2 --dry-run` now prints `"Would delete resource 1 2"`, showing both `id` and `type`. |
| (4) | `computers_inventory.go` (`download` example) | ✅ Fixed | Extracted a shared `pathParamUsage` helper (used by both `Use` and the example generator) so the example always matches the real arg count. Verified live: `--help` now shows `download <id> <attachmentId> -O output.bin`. |
| (5) | privilege-equivalence check in `pairCollectionBulkActions` | 🔴 Open | Not addressed; no reply given beyond "not addressed here" in the updated PR description. Low priority — nice-to-have, no current spec triggers it. |
| (6) | PR description undercounted the fix (4 vs actual 5 commands) | ✅ Fixed | Description now says "fixes five pre-existing broken commands" and explicitly lists `sessions-status`. |
| (7) | pre-existing `export --out-file` bug | 💬 Dismissed | Author: "pre-existing (6 resources); warrants a separate follow-up against the shared template." Reasonable — out of scope for this PR, affects other resources equally. |
| (8) | scope bundling observation (feature + unrelated bugfix) | 🔴 Open | Acknowledged only ("not addressed here"); this was a process observation for future PRs, not a code change, so nothing further to do here. |

No new comments from other reviewers appeared between my initial review and this one, and the fix commit didn't touch any file outside the generator/test/generated surface already in scope.

<details>
<summary>Scope of this re-review</summary>

- Diffed the single fix commit (`9494250`, "fix(generator): address PR #279 review (blocking findings 1-4)") against my baseline review's commit (`447fc41`).
- Re-ran `go build ./...`, `go test ./generator/... ./internal/commands/...` (all pass, exit 0 — verified directly, not through a piped/truncated check), `make lint` (0 issues), and `make verify-generated` (zero diff after regeneration).
- Manually rebuilt the binary and exercised the actual fixes: `installation-retry --all --no-input` (confirms the new refusal), `installation-retry <id> --all` (confirms mutual exclusivity still works), `notifications delete 1 2 --dry-run` (confirms both args now shown), and `computers-inventory download --help` (confirms the corrected example).
- Spot-checked all 7 other generated files mechanically touched by the finding-(3) template fix (`computer_prestage_scopes.go`, `enrollment_customization_panels.go`, `erase_device_computers.go`, `erase_device_mobiles.go`, `mobile_device_prestage_scopes.go`, `remove_computer_mdm_profiles.go`, `remove_mobile_device_mdm_profiles.go`) — all single-arg ops, so `strings.Join(args, " ")` is functionally identical to the old `args[0]`, confirmed by diff.
- Did not re-run the full specialist panel (usability/scope/test-quality/devil-advocate) since the fix is narrowly scoped to the 4 findings those agents' input fed into, and each fix was independently verifiable by direct source reading, build, and execution.

</details>

This covers all findings — the PR is merge-ready.

ktn-jamf (approved):

PR #303, opened 2026-07-22
fix(generator): warn on all write-only fields dropped by update --set (#302)

### LINKED ISSUE #302: update --set silently wipes write-only fields (e.g. prestage managed admin password) with no warning
### Summary

`update --set` silently wipes write-only fields (e.g. the managed local admin password on a computer prestage) and emits **no warning**, because the write-only-field warning only fires for fields that are simultaneously **top-level** *and* **required** in the PUT request schema.

### Reproduce

```
jamf-cli pro computer-prestages update 12 --set department="Config Management"
```

Result: the prestage's managed local admin password (`accountSettings.adminPassword`) is blanked. In the Jamf Pro UI the password field is now empty and editing the prestage forces you to re-enter it. The server often also returns a 500 while still persisting the mutation (the resulting managed-admin state is invalid).

### Root cause

`update --set` is not a partial update — it is fetch → merge → **PUT the whole record** (`internal/commands/pro/generated/*.go`, `fetchForMerge` + `deepMergeJSON`). Write-only fields (`writeOnly: true`) are never returned by the GET, so they are absent from the fetched body and get sent back blank on the PUT.

The existing guard, `writeOnlyRequiredFields` in `generator/parser/generator.go`, only warns for fields that are **both required and top-level** in the PUT schema:

```go
for name, prop := range s.Properties {
    if prop.WriteOnly && required[name] {   // top-level + required only
        out = append(out, name)
    }
}
```

For `ComputerPrestagesV3`, `adminPassword` is `writeOnly: true` but lives **nested** under `accountSettings` and is **not** in the schema's `required` list — so the warning never fires. Same for `recoveryLockPassword`.

### Expected

When `--set` is used on a PUT update and any write-only field (at any nesting level) is *not* supplied, print a warning that the value will be blanked and how to preserve it, e.g.:

```
warning: computer-prestage field "accountSettings.adminPassword" is write-only: the server never returns it, so this update will blank any existing value. Pass --set accountSettings.adminPassword=<value> to preserve it.
```

### Fix

Widen the generator guard: detect write-only fields recursively (dot-notation paths), drop the `required` gate, and check presence against the nested `--set` map at runtime. Templates in `generator/parser/generator.go`.

neilmartin83:
Fixes #302.

## Problem

`update --set` is a fetch → merge → **PUT-the-whole-record** operation, not a partial update. Write-only fields (`writeOnly: true` — passwords, secrets) are never returned by the GET, so they get dropped from the fetched body and sent back **blank** on the PUT, blanking the stored value server-side.

The old guard (`writeOnlyRequiredFields`) only warned for fields that were **both top-level AND required** in the PUT schema. Jamf secrets are almost never either, so the warning fired for essentially nothing. Reported case: `computer-prestages update <id> --set department=...` silently blanked `accountSettings.adminPassword`.

## Fix

- Replace `writeOnlyRequiredFields` with **`writeOnlyFields`** — a depth-capped, `$ref`-cycle-guarded recursive walk (mirroring `schemaFilterLiteral`) that reports **every** write-only field in dot-notation, regardless of `required`.
- Add **`hasNestedKey`** runtime helper to check whether a dot-path was supplied via `--set` (the `--set` map is nested, so a flat key lookup would always miss nested fields).
- Template warns per omitted write-only field with the exact `--set` to preserve it.

Example output now:

```
warning: computer-prestage field "accountSettings.adminPassword" is write-only: the server never returns it, so this update will blank any existing value. Pass --set accountSettings.adminPassword=<value> to preserve it.
```

## Coverage

Regeneration lit up warnings on every credential field on a PUT-update endpoint — no spurious noise:

- `computer-prestages`: `accountSettings.adminPassword`, `recoveryLockPassword`
- `distribution-points`: `httpsPassword`, `sshPassword`, `readOnlyPassword`, `readWritePassword`
- `smtp-server`: `basicAuthCredentials.password`, `googleMailCredentials.clientSecret`, `graphApiCredentials.clientSecret`
- `gsx-connection`: `gsxKeystore.keystoreBytes`, `gsxKeystore.keystorePassword`
- `sso-settings-cert`: `keystorePassword`
- `cloud-ldaps`: `server.keystore.password`
- `user-accounts`: `password`, and other `plainPassword`/`token` fields

## Not addressed / out of scope

The **500** some users see alongside this is a **separate server-side product issue** (the Jamf backend rejecting the resulting invalid managed-admin state) — not a CLI bug and not touched here. The warning is the ceiling of what the client can do: write-only means the value can't be read back, so the CLI cannot preserve it automatically — only warn. `patch` (changed-fields-only) would sidestep the full-replace, but these resources are PUT-only.

## Tests

- `TestWriteOnlyFields` (parser) — nested + top-level detection, read-only excluded, `required` gate removed.
- `TestHasNestedKey` — dot-path presence semantics.
- `TestUpdateSetWriteOnlyWarning` — drives the real `computer-prestages update --set` against a mock client: warning fires when the password is omitted, is suppressed when supplied, and the supplied password reaches the PUT body.

`make generate && make test` green, lint clean, `make build` OK.

Also adds a `docs/solutions/logic-errors/` postmortem.

### DIFF
--- a/docs/solutions/logic-errors/write-only-fields-dropped-by-update-set-2026-07-22.md
+++ b/docs/solutions/logic-errors/write-only-fields-dropped-by-update-set-2026-07-22.md
@@ -1,0 +1,131 @@
+---
+title: "update --set silently drops write-only fields; warn on every one, not just required top-level"
+date: 2026-07-22
+category: logic-errors
+module: generator/parser
+problem_type: bug
+severity: high
+applies_when:
+  - "Adding or auditing the fetch-merge-PUT 'update --set' code path in the modern API generator"
+  - "A PUT request schema contains writeOnly fields (passwords, secrets, keystores, tokens)"
+  - "A user reports a credential field blanked after an unrelated update --set change"
+tags:
+  - generator
+  - update-set
+  - write-only
+  - fetch-merge-put
+  - credentials
+  - data-loss
+  - computer-prestages
+---
+
+# update --set silently drops write-only fields
+
+## Context
+
+`update --set KEY=VALUE` on a modern API command is **not** a partial update. It
+is fetch → merge → PUT-the-whole-record (`fetchForMerge` + `deepMergeJSON` in the
+generated `registry.go`; template in `generator/parser/generator.go`
+`resourceTemplate`). The current resource is GET'd, read-only fields are stripped
+via the `fieldFilter` allowlist, the caller's `--set` changes are merged, and the
+full body is PUT back.
+
+Write-only fields (`writeOnly: true` in the spec — passwords, secrets) are
+**never returned by the GET**. So they are absent from the fetched body, absent
+from `--set` unless the caller names them explicitly, and therefore sent back
+**blank** on the PUT. The server treats the missing value as "clear it."
+
+Reported via issue #302: `jamf-cli pro computer-prestages update 12 --set
+department="Config Management"` silently blanked the managed local admin password
+(`accountSettings.adminPassword`). The Jamf UI then showed an empty password field
+and often the server returned a 500 while still persisting the mutation (the
+resulting managed-admin state — enabled, no password — is invalid).
+
+The pre-existing guard, `writeOnlyRequiredFields`, only warned for fields that
+were **both required AND top-level** in the PUT schema:
+
+```go
+for name, prop := range s.Properties {
+    if prop.WriteOnly && required[name] {   // top-level + required only
+        out = append(out, name)
+    }
+}
+```
+
+`accountSettings.adminPassword` is nested one level down and is not in the
+schema's `required` list, so the warning never fired. Same for
+`recoveryLockPassword`, distribution-point passwords, SMTP/GSX/SSO/LDAP secrets,
+and user-account passwords — every one of them slipped through silently.
+
+## Guidance
+
+**Any write-only field on a PUT-update endpoint is undetectably dropped by
+`update --set` unless the caller re-supplies it. Warn on every one — at every
+nesting level, regardless of `required`.**
+
+The warning is the only safety net possible here: the CLI physically cannot read
+the value back to preserve it, so it cannot fix the data loss — only tell the user
+it is about to happen and how to prevent it.
+
+### The detector
+
+Walk the PUT request schema recursively, depth-capped and `$ref`-cycle-guarded
+(mirror `schemaFilterLiteral`'s traversal), returning **dot-notation paths** of
+every write-only field. Do not gate on `required`:
+
+```go
+func writeOnlyFields(op *Operation, schemas map[string]*Schema) []string {
+    if op == nil || op.RequestBody == nil || op.RequestBody.Schema == nil {
+        return nil
+    }
+    var out []string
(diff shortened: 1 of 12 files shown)


ktn-jamf (approved):
## Review: warn on all write-only fields dropped by `update --set`

**🎯 Verdict: Merge-ready** · **Rating: 5/5**

Tightly-scoped, well-reasoned bug fix. Root-cause analysis is correct, the implementation mirrors an existing proven traversal, and the tests exercise the regression end-to-end rather than just the helper in isolation.

### What this does
`update --set` is a fetch → merge → **PUT-whole-record** cycle. `writeOnly` fields (passwords, secrets, keystores) are never returned by the GET, so they're dropped from the fetched body and PUT back blank — silently wiping the stored value (issue #302). The old guard (`writeOnlyRequiredFields`) only warned for fields that were **both** top-level **and** `required`, which almost no Jamf secret is — so it warned for essentially nothing. This replaces it with `writeOnlyFields` (a depth-capped, `$ref`-cycle-guarded recursive walk reporting every write-only field in dot-notation) plus a runtime `hasNestedKey` helper.

### Strengths
- ✅ **Traversal is a faithful mirror of `schemaFilterLiteral`** — identical `depth > 6` cap, identical object detection, identical path-scoped `visited[SchemaRef]` cycle guard with set-then-delete. Copying the proven walk verbatim means the warning walk and the read-only filter can't drift apart.
- ✅ **Write-only object handling is correct** — a write-only object is reported as a whole path and not descended into (`continue`).
- ✅ **Tests verify behavior, not implementation** — `TestUpdateSetWriteOnlyWarning` drives the real generated `computer-prestages update --set` against a mock, captures stderr, asserts the warning fires on omission, is suppressed on supply, **and** that the supplied password actually reaches the PUT body. That last assertion catches the "warning suppressed but value still dropped" class of regression.
- ✅ **Correct workflow for generated code** — template edited, `make generate` re-run; `hasNestedKey` emitted once in `registry.go`, not duplicated per file. `make verify-generated` confirms sync.
- ✅ Postmortem added under `docs/solutions/logic-errors/` with correct frontmatter.

### Findings

🟩 **(1) nice-to-have — write-only fields inside array-of-object items are not traversed.** `collectWriteOnlyFields` only recurses into `object` properties; arrays are leaves (as in `schemaFilterLiteral`). If a future PUT schema nests a secret inside an array element, the warning would silently miss it. No current spec does this and `--set` can't address array indices by dot-path anyway, so this is a documented-limitation note, not a defect. A one-line comment on `collectWriteOnlyFields` noting arrays are intentionally out of scope would keep the next reader from assuming full coverage.

*Non-issue, verified and dismissed:* the "will blank any existing value" wording generalizes from one confirmed field to all write-only fields — but "any existing value" hedges the no-prior-value case, it's a stderr data-loss warning where over-warning is the safe default, and it's the documented ceiling of what a client can do for unreadable fields. Fine as written.

### Verified locally
`go build ./...` ✅ · `go test ./generator/parser/ ./internal/commands/pro/generated/` ✅ · `golangci-lint run generator/parser/...` → 0 issues ✅ · `make verify-generated` → "Generated code is up to date" ✅

---
<sub>🤖 Reviewed with the pr-review skill (Claude Opus 4.8)</sub>

PR #305, opened 2026-07-24
fix(generator): parse array/object --set values by schema type (#304)

### LINKED ISSUE #304: `--set` silently stringifies array/object field values instead of parsing JSON or erroring
## Summary

`--set KEY=VALUE` silently coerces the value to a **string** for fields whose target type is an array (or object). Instead of parsing a JSON value or rejecting the input, it ships a wrong-typed payload to the API. This is silent data corruption — the command appears to accept the input but sends garbage.

Discovered while trying to set `customPackageIds` on a computer prestage.

## Version

```
jamf-cli 1.24.0
commit: 68fdc2c
```

## Reproduction

`customPackageIds` on a computer prestage is an array of strings. Every `--set` form produces the wrong type (verified via `--dry-run -vvv`, inspecting the outgoing request body):

| `--set` input | Serialized in request body | Resulting JSON type |
|---|---|---|
| `customPackageIds=["295"]` | `"customPackageIds":"[\"295\"]"` | string (not array) |
| `customPackageIds.0=295` | `"customPackageIds":{"0":295}` | object (not array) |
| `customPackageIds=295` | `"customPackageIds":295` | number (not array) |

```bash
jamf-cli pro computer-prestages update <id> --set 'customPackageIds=["295"]' --dry-run -vvv
# outgoing body contains:  "customPackageIds":"[\"295\"]"   <-- a string, not an array
```

## Expected

One of:

1. **Parse** the value as JSON when the target field is a non-scalar (so `customPackageIds=["295"]` becomes a real array `["295"]`), **or**
2. **Reject** with a clear error when the target field is an array/object, e.g.
   `error: field "customPackageIds" is an array; set it by piping a full JSON document via stdin`

Either is fine. Silent stringification is the problem.

## Actual

The value is stored as a string/object/number and sent to the API as-is. No error, no warning. On this prestage endpoint the malformed payload is either rejected downstream or (worse) silently accepted with the wrong shape.

## Working alternative (for reference)

Piping a full JSON document via stdin sets the array correctly:

```bash
jamf-cli pro computer-prestages get <id> -o json \
  | jq '.customPackageIds = ["295"]' \
  | jamf-cli pro computer-prestages update <id>
# outgoing body:  "customPackageIds":["295"]   <-- correct
```

Verified end-to-end against a test instance: the array persists (re-fetch confirms `customPackageIds: ["295"]`).

## Suggested fix

In the `--set` merge logic, look up the target field's type from the resource schema:

- If the target is a scalar → current behavior.
- If the target is an array/object → attempt a JSON parse of the value, and if that fails, error out with the stdin hint above rather than coercing to a string.

At minimum, never coerce a value into a type that doesn't match the target field. Failing loudly beats sending malformed data.

## Docs note

The `update --help` "Available fields" list only shows scalar fields, so array fields like `customPackageIds` aren't advertised as settable via `--set` at all. Worth documenting explicitly that array/object fields must go through the JSON-stdin path.

neilmartin83:
Fixes #304.

## Problem

`--set KEY=VALUE` silently coerced every non-scalar value to a **string**, shipping a wrong-typed payload with no error or warning — silent data corruption. Verified against outgoing request bodies:

| `--set` input | Old serialization | JSON type |
|---|---|---|
| `customPackageIds=["295"]` | `"customPackageIds":"[\"295\"]"` | string, not array |
| `customPackageIds.0=295` | `"customPackageIds":{"0":295}` | object, not array |
| `customPackageIds=295` | `"customPackageIds":295` | number, not array |

## Fix

The generator already knows every field's type from the request schema. It now emits a per-op `map[string]string` of dot-path → JSON kind into each `--set` call site, and the runtime parses each value against its target kind:

- **array / object** → JSON-decode and verify the kind; reject a mismatch with a stdin/JSON hint
- **integer / number / boolean** → `strconv` coerce; reject non-parseable input
- **string** → keep the literal text (also fixes numeric-looking string IDs like `enrollmentSiteId=-1` being sent as a number)
- **unmodelled field** → best-effort JSON parse of `[`/`{`-prefixed values, else legacy inference (no regression)

`checkSetParentKind` rejects dot-notation into an array/scalar parent (`foo.0=x`). `null` still yields JSON null. `deepMergeJSON` already **replaces** arrays (only objects merge key-by-key), so a parsed array correctly overwrites the fetched value — parsing is a genuine feature win, not just a safety fix.

Both `update --set` (fetch-merge-PUT) and `patch --set` (merge-patch) are fixed. `--set` help now advertises the settable array/object fields and notes they take JSON values (addresses the issue's docs note).

## Behavior now

```
$ jamf-cli pro computer-prestages update 1 --set 'customPackageIds=["295"]' --dry-run -vvv
  "customPackageIds":["295"]                       # real array

$ ... --set 'customPackageIds=295'
  error: field "customPackageIds" is an array but the value is not a JSON array; ...

$ ... --set 'customPackageIds.0=295'
  error: field "customPackageIds" is an array; set the whole array with a JSON value ...
```

## Testing

- New table test `TestBuildMergePatchFromSet_Typed` (arrays, objects, scalar coercion, mismatch errors, dot-index guard, null, unmodelled fallback).
- Existing `--set` tests updated for the new signature; full suite + lint + `verify-generated` green.
- **End-to-end against a live tenant** (create→set→re-fetch→delete), value persists with the correct JSON type:
  - `api-roles.privileges` — array of strings
  - `advanced-mobile-device-searches.criteria` — array of objects
  - `computer-prestages.customPackageIds` — array (this endpoint returns HTTP 500 while still persisting — a server bug, unrelated to this change)
  - `buildings.zipPostalCode` — numeric-looking string kept as a string

## Notes

- `internal/platform/body.go` and `internal/security/body.go` already JSON-parse `[`/`{`/`"`-prefixed `--set` values (they lack schema access), so they don't exhibit the reported bug as badly; left as-is.
- Solutions doc added: `docs/solutions/logic-errors/set-array-object-stringification-2026-07-24.md`.


ktn-jamf (approved):
**Verdict: Merge-ready** ✅ — a clean, well-scoped fix for a real silent-data-corruption bug. All findings are optional follow-ups.

## Rating: 5/5

This is exactly how a generator-level correctness fix should look. The source of truth (the request schema) is already known to the generator, so the fix threads a per-op `dot-path → JSON kind` map into each `--set` call site and parses values against it. I verified the whole chain rather than trusting the description:

- **Correctness traced through edge cases** — array/object JSON decode + kind-check, scalar coercion, the `null` short-circuit, empty-array clear (`field=[]`), trailing-garbage rejection (`parseJSONSetValue`), and the dotted-parent guard (`checkSetParentKind`) at both top level *and* one level of nesting (`clientCert.data.0=x` is correctly rejected).
- **Both call sites fixed** — `patch --set` and `update --set` (fetch-merge-PUT) each receive their operation's schema map; no call site is missing the argument. `deepMergeJSON` replaces arrays wholesale, so a parsed array correctly overwrites the fetched value.
- **Emitted maps match the issue's examples** — `customPackageIds: array`, `privileges: array`, `criteria: array`, `enrollmentSiteId: string`, `zipPostalCode: string`.
- **`make verify-generated` is green** (template ↔ generated in sync) and the `--set` tests pass.

The error messages are a highlight — every rejection states the expected shape *and* the stdin escape hatch. The live-tenant create→set→re-fetch→delete verification across four field shapes is the right diligence for a data-typing change.

---

**🟩 (1) consistency — `--set` typing now diverges across products.** `internal/platform/body.go:73` and `internal/security/body.go` keep a prefix-heuristic `parseSetValue` (JSON-decode only on `[`/`{`/`"` prefix, no schema, no mismatch errors). After this PR, `pro … update --set customPackageIds=295` errors loudly, but the platform/security equivalent silently coerces. The PR scopes this out (no schema access there) — reasonable, but the CLI now teaches two `--set` contracts. *Fix: a follow-up tracking issue, or a one-line note in help/glossary. No code change here.*

**🟩 (2) test-coverage — three branches this PR adds are untested.** `TestBuildMergePatchFromSet_Typed` covers array success/rejections, object success, and scalar coercion, but not: the **object-mismatch** branch (`skipSetupItems=["a"]`), the **scalar-parent guard** (`enrollmentSiteId.foo=1`), or `parseJSONSetValue`'s **trailing-garbage** rejection (`customPackageIds=[1,2] junk`). These are the branches most likely to regress silently on a future template edit. *Fix: three more table cases asserting each returns an error.*

**🟩 (3) documented-limitation — schema map is two levels deep.** `setFieldTypeMap` records top-level + one level of object nesting (matching `flattenSchemaToScalarFields`). A `string`-typed field ≥2 levels deep that looks numeric can still fall to legacy inference and be coerced to a number — the same bug class, just deeper. Within #304's scope and already acknowledged in comments. *Fix: a one-line "Limitations" note in the solutions doc so the next auditor knows where the guarantee ends.*

<details>

- [x] Design/architecture: schema-driven typing threaded through the existing generate-time literal-emission pattern; sound and consistent.
- [x] Correctness: traced decode/coercion/null/`[]`/trailing-garbage/parent-guard; both call sites + array-replace confirmed.
- [x] Security: `--set` values flow only into `json.Marshal` — no injection/auth/secret surface.
- [na] Performance: map-building is generate-time; runtime is trivial lookups.
- [x] Test coverage: solid happy paths + array rejections; three error branches untested (finding 2).
- [x] Reliability: all mismatches error with actionable hints incl. stdin fallback; nothing swallowed.
- [x] Code quality: clear helpers, excellent errors, matches template style.
- [x] Simplification: reuses `parsePatchValue` for the unmodelled fallback; no redundant abstraction.
- [na] Frontend: CLI only.
- [x] Documentation currency: `--set` help updated; solutions doc correctly placed under `docs/solutions/logic-errors/`.
- [x] Project rules: generated-code boundary respected (template edited, `verify-generated` green); no credential-flag changes.

</details>

This covers all findings — none block merge; addressing them is optional cleanup.

---
🤖 Principal-engineer review via [Claude Code](https://claude.com/claude-code) `/pr-review`

PR #316, opened 2026-08-04
fix(blueprints): look up configuration profiles by ID, preserve CR line breaks

### LINKED ISSUE #315: Converting configuration profile by ID
### Product namespace

pro (Jamf Pro / Platform API)

### What problem does this solve?

I would highly appreciate the option to use the ID of a configuration profile instead of the name when converting to a Blueprint. That’s because we have duplicate names in our setup.

Probably my request goes further than only configuration profiles and the command to convert them. I didn’t use jamf-cli much yet, but I think an option to use an ID or display name is always good to have.

### Proposed solution

_No response_

### Alternatives considered

_No response_

neilmartin83:
Closes #315.

Two changes to Classic configuration profile handling. The first is the reported feature; the second is a defect found while probing the same code path.

## 1. ID lookup for profile conversion (#315)

Jamf Pro allows duplicate profile display names — per site, and across the computer/mobile endpoints — so name-only lookup could silently convert the wrong profile. The reporter has duplicates and no way to disambiguate.

- `blueprints import-profile [<id>]` takes the Classic ID positionally, with `--name` for display-name lookup. Same identifier convention as the generated Pro/Platform commands (positional `<id>` + `--name`, rejecting both together).
- `blueprints components configuration-profile` gains `--id`, mutually exclusive with `--from-file`/`--name`.
- **Not a breaking change:** a non-numeric positional argument is still treated as a display name, so every documented invocation keeps working.

Name resolution **reuses the generated resolver** rather than adding a parallel one: `resolveClassicNameToIDForApply` is exposed as `ResolveClassicNameToID` from the classic registry template (one wrapper, no call-site churn). Duplicates therefore behave exactly as they do in `pro classic-macos-config-profiles`: an interactive pick, or under `--no-input` an error listing the colliding IDs.

Two side benefits of resolving through the list endpoint instead of `/JSSResource/.../name/{name}`: names containing `/` (rejected outright by that path) and `+` (dropped by the platform gateway proxy) now resolve, and a name that only exists under the other profile type errors with a `--type` hint:

```
no computer configuration profile named "Certificate - AGNI - 2029-07-09" — a mobile profile matches; re-run with --type mobile
```

## 2. Carriage-return line breaks survive the wire (ports jamfplatform-go-sdk#48)

`minimizeClassicPlistSourceEscaping` decoded `&#13;` to a literal CR. XML 1.0 §2.11 line-end normalisation turns that into LF in transit, and Jamf Pro deletes literal LF from verbatim-stored payload types — merging the words either side. Net effect: a profile exported from Jamf Pro's own UI, which writes `&#13;` for login-window banner breaks, could not be re-uploaded with its line breaks intact.

- CR references (`&#13;`, `&#xD;`, zero-padded and case variants) are left undecoded.
- `escapeClassicAmpPreservingCRRefs` leaves the opening `&` of a CR reference bare so the server's single decode yields a real CR; escaping it stores the reference as literal text (the PI-827 extra layer, i.e. a device displays `&#13;`).
- **The exemption is CR-only and must not be generalised** — `&#38;` left bare decodes to a bare `&` and the write 409s. `TestNormalizePayloads_OnlyCRReferencesLeftBare` pins that boundary.

Read-back verification is adjusted to match — the half of terraform-provider-jamfplatform#309 that applies here. Without it, this fix would make every correctly stored line break look unfaithful:

- CR/LF folded before comparison (a stored CR reads back as LF from MCX/mobile fragments, and as a raw CR from verbatim types). `U+2028`/`U+2029`/`U+0085` still compared strictly — they round-trip byte-exact, so drift detection is preserved.
- `DiffPayloadValuesDetailed` classifies each divergence — line breaks/tabs deleted, PI-827 entity layer, non-BMP replacement, value absent, unexplained — so the warning carries the remedy that applies rather than blaming PI-827 for every class:

```
warning: the server stored 1 payload value(s) differently than submitted:
  - PayloadContent[0].LoginwindowText: line breaks/tabs removed — Jamf Pro deletes literal line
    feeds and tabs here; write a line break as the character reference &#13; (the form Jamf Pro's
    own UI writes) or move the value into an Application & Custom Settings payload, which keeps them
```

Report caps at 3 findings: a server-injected `PayloadContent` entry shifts array indices and turns one defect into a column of noise.

Both apply to the generated profile commands only — `create`, `update`, and both `apply` branches on `classic-macos-config-profiles` and `classic-mobile-config-profiles` (4 sites each). Deliberately not extended to `mobiledeviceapplications --appconfig-file`, which is spec-compliant and would be corrupted by the escape.

## Testing

- `make test` clean, `make lint` 0 issues, `make generate` run (template change → regenerated `classic_registry.go`).
- New unit tests: profile ID/name resolution (numeric arg skips the lookup, non-numeric resolves as a name, duplicate IDs listed, wrong-type hint, flag conflicts), CR reference table + `classicCRRefLen`, wire-form CR bareness and its boundary, CR/LF read-back tolerance, and one case per fidelity class.
- Live-verified against a Jamf Pro 11.30.2 tenant via platform gateway auth:
  - `import-profile 7293`, `import-profile --name "…"`, `import-profile "…"` (positional name) all created blueprints; every error path checked.
  - `classic-macos-config-profiles create --mobileconfig-file` with `&#13;&#13;` in `LoginwindowText` → stored payload parses to **CR=2, LF=0**, words intact.
  - Literal-LF control → the line-break class reported by path, as shown above.
  - All test profiles and blueprints deleted; tenant left clean.

Wire model recorded in `docs/solutions/conventions/classic-profile-payload-ampersand-escaping-2026-07-30.md` (whitespace clause: LF/TAB deleted, CR the only survivor, `U+2028`/`U+2029`/`U+0085` byte-exact, non-BMP unstorable).


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Adds ID-based lookup for Classic configuration profiles in `blueprints import-profile` / `components configuration-profile`, and stops the send path from destroying `&#13;` line breaks.
> Blocking: (1), (2), (3). See collapsed section for 4 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 once the duplicate-name prompt/error text stops giving apply-path advice in a blueprints command, the `components configuration-profile` help matches what the code actually does, and CLAUDE.md records the new hybrid positional convention.

The CR-preservation half of this PR is genuinely strong work: I traced `minimizeClassicPlistSourceEscaping` → `escapeClassicAmpPreservingCRRefs` → `classicProfilePayloadFromBody` byte-by-byte through the literal-text (`&amp;#13;`), mixed (`&#13;&#38;`), and round-trip cases and could not construct a corruption or a masked-regression path. The `normalizeLineEndings` fold does not hide deletions — only CR↔LF equivalence — so drift detection survives. `make verify-generated` reports the template and `classic_registry.go` in sync, `go build ./...` is clean, and `go test ./internal/commands/... ./internal/profileconvert/...` passes.

The findings below are all on the ID-lookup half, and all of them are user-facing text or convention debt rather than logic defects.

---

🟧 (1) (usability, c=100) — `generator/classic/generator.go:1523` and `:1531` (rendered at `internal/commands/pro/generated/classic_registry.go:185`, `:193`): the duplicate-name resolver gives apply-path advice to blueprints users

This PR's whole reason for existing (issue #315) is duplicate display names, and `resolveClassicProfileID` now routes every blueprints name lookup through `resolveClassicNameToIDForApply`. That resolver's collision messages were written for `classic-*-config-profiles apply`/`update` and are wrong in the blueprints context:

- Under `--no-input`: `multiple resources found with name %q (IDs: %s); resolve duplicates or use update with a specific ID`. There is no `update` in `pro blueprints import-profile`. The correct remedy — the one this PR just built — is "re-run with the profile ID" or `--id`.
- Interactively: `Enter number to replace (or 0 to cancel): `. `import-profile` creates a blueprint; nothing is replaced. Telling a user picking between two profiles that they are about to "replace" one is actively alarming.

So the exact moment the feature should teach the user how to disambiguate is the moment it prints advice that does not apply. The root cause is design: `ResolveClassicNameToID` hard-codes both the remedy sentence and the prompt verb, giving the new hand-written caller no way to phrase its own.

**Suggested fix** — make the resolver's wording caller-supplied. Minimal version, in the template:

```diff
-func ResolveClassicNameToID(ctx context.Context, client registry.HTTPClient, apiPath, wrapperKey, name string, noInput bool) (string, error) {
-	return resolveClassicNameToIDForApply(ctx, client, apiPath, wrapperKey, name, noInput)
+// ClassicNameCollisionError carries the colliding IDs so callers can phrase
+// the remedy that applies to their own command.
+type ClassicNameCollisionError struct {
+	Name string
+	IDs  []string
+}
+
+func (e *ClassicNameCollisionError) Error() string {
+	return fmt.Sprintf("multiple resources found with name %q (IDs: %s)", e.Name, strings.Join(e.IDs, ", "))
+}
+
+func ResolveClassicNameToID(ctx context.Context, client registry.HTTPClient, apiPath, wrapperKey, name, verb string, noInput bool) (string, error) {
+	return resolveClassicNameToIDForApply(ctx, client, apiPath, wrapperKey, name, verb, noInput)
 }
```

…threading `verb` into the prompt (`"Enter number to %s (or 0 to cancel): "`, `"replace"` for apply, `"import"` / `"use"` for blueprints) and returning `*ClassicNameCollisionError` instead of the pre-phrased string. `findClassicProfileByName` then wraps it:

```go
var collision *generated.ClassicNameCollisionError
if errors.As(err, &collision) {
    return "", fmt.Errorf("%w; re-run with a specific ID, e.g. `import-profile %s`", collision, collision.IDs[0])
}
```

If you'd rather not touch the template's error shape, the cheaper alternative is to keep `ResolveClassicNameToID` as-is but have `findClassicProfileByName` do its own list+match (duplicating ~20 lines) — I'd take the typed error over that, since the PR's stated goal was explicitly *not* to add a parallel resolver.

**Fixed when:** `import-profile --name "<dup>" --no-input` names the profile ID as the remedy (not `update`), the interactive prompt does not say "replace" for a non-replacing command, and a test asserts the blueprints-path collision message contains the ID-based remedy. `make generate && make test` clean.

---

🟧 (2) (documentation, c=100) — `internal/commands/pro_blueprints.go:833`: `components configuration-profile` help contradicts the code

```
Jamf Pro allows duplicate profile display names; --name errors and lists the
matching IDs when more than one profile shares the name. Use --id to pick one.
```

`findClassicProfileByName` (line 965) passes `noInput || !allowPrompt`, and `downloadClassicProfile` reaches it with `allowPrompt = true`. So on a TTY without `--no-input` the command **prompts** interactively; it only errors under `--no-input`. The sibling `import-profile` help (line 1198-1202) describes this correctly — this one doesn't. Stale docs are worse than missing docs: a user reading this expects a non-interactive failure and gets a blocking prompt instead, which matters in a wrapper script that inherits a TTY.

**Suggested fix:**

```diff
-Jamf Pro allows duplicate profile display names; --name errors and lists the
-matching IDs when more than one profile shares the name. Use --id to pick one.
+Jamf Pro allows duplicate profile display names: when a name matches more than
+one profile you are prompted to pick one, or with --no-input the command errors
+listing the matching IDs. Use --id to skip the lookup entirely.
```

**Fixed when:** The `components configuration-profile` long help describes prompt-on-TTY / error-under-`--no-input`, matching `import-profile`'s wording and the actual `noInput || !allowPrompt` behaviour.

---

🟧 (3) (design/documentation, c=100) — `CLAUDE.md:242`: `import-profile` introduces a third identifier convention that the repo's own convention doc doesn't record

CLAUDE.md states the Platform rule flatly, and enumerates its one exception:

> Platform takes a positional **`<id>`** plus a `--name` flag (`delete <uuid>` or `delete --name "My Blueprint"` — **passing a name positionally is sent as an ID and 404s**). This holds for both the generated commands … and the hand-written ones (`resolveBlueprintID` in `pro_blueprints.go`) … Exception: `blueprints clone <source-name> <new-name>` takes names positionally.

`resolveClassicProfileID` (`pro_blueprints.go:976-1006`) is neither: it sniffs the positional with `isClassicID()` and dispatches to ID *or* name. That's a deliberate and defensible backwards-compat choice, but it's a **new precedent** — the next contributor adding a dual-identifier hand-written command now has three patterns to choose from and a CLAUDE.md that describes only two. Left unrecorded, either the sniffing spreads by copy-paste or the documented rule quietly becomes false.

It also has a real behavioural corner the PR body's "Not a breaking change" claim doesn't mention: a profile whose display name is entirely numeric (`"2024"`, `"8021X"` is fine, `"8021"` is not) is now resolved as an ID. `--name` is the escape hatch, but nobody will know to reach for it until the 404 arrives.

**Suggested fix** — amend the CLAUDE.md paragraph:

```diff
-Exception: `blueprints clone <source-name> <new-name>` takes names positionally.
+Exceptions: `blueprints clone <source-name> <new-name>` takes names positionally;
+`blueprints import-profile [<id>]` sniffs its positional (`isClassicID` in
+`pro_blueprints.go`) — an all-digit argument is a Classic ID, anything else a
+display name — to keep pre-#316 name-based invocations working. Consequence: a
+profile whose display name is all digits must be looked up with `--name`. Do not
+copy the sniffing into new commands without a comparable compat reason.
```

…and add the same caveat to `import-profile`'s long help, next to the existing "A non-numeric positional argument is treated as a display name" sentence.

**Fixed when:** CLAUDE.md's Platform identifier-convention paragraph names `import-profile` as an explicit exception with its rationale and the all-digit-name caveat, and `import-profile --help` warns that an all-digit display name needs `--name`.

<details>

🟩 (4) (usability, c=100) — `internal/commands/pro_blueprints.go:899`: `--id` silently falls back to a name lookup on non-numeric input

`--id` is passed as `downloadClassicProfile(cmd, cliCtx, profileID, profileName, profileType)` — i.e. into the `arg` slot. `resolveClassicProfileID` sees `nameFlag == ""` and a non-numeric `arg`, so `--id "My Restrictions"` performs a *name* lookup and, on failure, reports `no computer configuration profile named "My Restrictions" found`. A user who explicitly typed `--id` gets an error about names.

**Suggested fix:**

```diff
 			if profileName != "" || profileID != "" {
+				if profileID != "" && !isClassicID(profileID) {
+					return fmt.Errorf("--id must be a numeric Classic API ID, got %q (use --name for a display name)", profileID)
+				}
 				// Download from Jamf Pro classic API
 				data, err = downloadClassicProfile(cmd, cliCtx, profileID, profileName, profileType)
```

**Fixed when:** `components configuration-profile --id abc` errors naming `--id` and pointing at `--name`, with a unit test covering it.

---

🟩 (5) (simplification, c=100) — `internal/profileconvert/classic_verify.go:85`, `:105`: two dead API surfaces introduced by the refactor

After the switch to `DiffPayloadValuesDetailed`, `DiffPayloadValues` has no production caller (`grep` finds only its own definition and tests), and `PayloadDiff.String()` is never called anywhere. Both are exported from an `internal/` package, so there's no external-compat reason to keep them. They'll read as live API to the next person.

**Suggested fix:** delete `PayloadDiff.String()`, and either delete `DiffPayloadValues` (porting the two tests that use it onto `DiffPayloadValuesDetailed`) or keep it with a comment saying it exists only as the paths-only convenience view.

**Fixed when:** Every exported symbol in `classic_verify.go` has a caller, or a comment explaining why it doesn't.

---

🟩 (6) (test-coverage, c=100) — `generator/classic/generator.go:2009-2026`: the new report cap is untested

`maxReported = 3` plus the `… and %d more` tail is new output-shaping logic with no test — `grep` for `verifyClassicProfileStored`, `maxReported`, or `and %d more` across `*_test.go` finds only a comment reference. The classification classes each got a case in `TestDiffPayloadValuesDetailed_Classification`; the truncation didn't. A regression here (off-by-one on the tail count, or capping before the total is captured) fails silently on stderr.

**Suggested fix:** extract the formatting into a pure `formatStoredPayloadWarning(diffs []profileconvert.PayloadDiff) string` in the template and assert 2 / 3 / 5 diffs produce no tail, no tail, and `… and 2 more` respectively.

**Fixed when:** A test in `internal/commands/pro/generated/` covers 3-and-under (no tail) and over-3 (correct remainder count).

---

🟩 (7) (performance, c=75) — `internal/commands/pro_blueprints.go:1001`: the wrong-`--type` hint costs a second full-collection GET on every miss

A name that matches nothing now fetches the entire *other* profile collection before erroring. On a tenant with thousands of profiles that's two full list payloads to produce a "not found". The hint is genuinely useful, so this is a fine trade — worth noting only because `--no-input` batch callers looping over names will pay it per miss. If it ever bites, gate the speculative lookup behind an interactive check or a `--no-hints` read.

**Fixed when:** N/A — noted for awareness; no change required.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/classic/generator.go` | 3/5 | CR-preservation logic is correct and tightly scoped, and the "must not be generalised" boundary is pinned by a test — but the resolver export ships apply-path wording to a non-apply caller (finding (1)), and the new report cap is untested (finding (6)) |
| `internal/commands/pro_blueprints.go` | 3/5 | Clean decomposition (`resolveClassicProfileID` / `fetchClassicProfile` / `classicProfileLabel`) with good error text — held down by the help/behaviour contradiction (2), the undocumented hybrid positional (3), and the `--id` fallback (4) |
| `internal/profileconvert/classic_verify.go` | 4/5 | `normalizeLineEndings` and `stringDiffReason` are correct and don't mask deletions; two dead exports left behind (finding (5)) |
| `internal/commands/pro/generated/classic_registry.go` | 5/5 | Generated output, confirmed in sync via `make verify-generated` — no hand-edit |
| `internal/commands/pro_platform_test.go` | 4/5 | `classicRouteMock` is a good addition; covers numeric-skip, name-resolve, duplicates, wrong-type, and both flag-conflict paths. Missing: `--id` non-numeric (finding (4)) |
| `internal/profileconvert/classic_verify_test.go` | 5/5 | One case per classification class plus the CR-tolerance and U+2028-strictness boundary — exactly the right pair |
| `internal/commands/pro/generated/classic_payload_escape_test.go` | 5/5 | `TestNormalizePayloads_OnlyCRReferencesLeftBare` simulating the server's single decode is the test that makes the CR exemption safe to own |

</details>

<details>

- [x] Design and architecture: reviewed the decision to reuse `resolveClassicNameToIDForApply` via a thin export rather than a parallel resolver (right call, but the error/prompt wording leaks the apply context — finding (1)); assessed the numeric-sniffing positional against the repo's stated Platform identifier convention at `CLAUDE.md:242` (finding (3))
- [x] Correctness: traced `minimizeClassicPlistSourceEscaping` → `escapeClassicAmpPreservingCRRefs` → server single-decode → `classicProfilePayloadFromBody` for `&#13;`, `&#xD;`, `&#013;`, `&amp;#13;` (literal-text case), `&#38;`/`&#x26;` (the 409 boundary), and `&#13;&#10;` mixed; confirmed `classicCRRefLen` and `minimize`'s `r == '\r'` branch agree on every form incl. the `semi > 12` bound, hex `x`/`X`, zero-padding, and negative/unparseable refs. Verified `normalizeLineEndings` folds CR↔LF only and still detects deletion (`"a\rb"` vs `"ab"` → line-break class). Checked `resolveClassicProfileID`'s four-way switch for all arg/flag combinations.
- [x] Security: no credential surface touched — no new flags/env for passwords, tokens, or client secrets, so the CLAUDE.md credential-input policy is unaffected. Confirmed leaving `&` bare inside CDATA cannot produce malformed XML (CDATA content is unparsed) and cannot smuggle a bare `&` into the stored value (`classicCRRefLen` accepts only `n == '\r'`). `url.PathEscape` retained on the ID path segment.
- [x] Performance: name lookup moved from a single `/name/{name}` GET to a full-collection list, plus a second full list on the wrong-`--type` hint path (finding (7)). Classic list responses carry id+name only and are unpaginated; acceptable, noted.
- [x] Test coverage: 5 new resolver tests, 3 new escaping tests, 7 new verify tests. Ran `go test ./internal/commands/... ./internal/profileconvert/...` — all pass. Gaps: `--id` non-numeric (4), report cap (6).
- [x] Reliability: confirmed `client.Do` errors on HTTP ≥ 400 (`internal/client/client.go:158`), so a 403 on the list endpoint propagates rather than silently reading as "no match"; the speculative other-type lookup correctly swallows only *its own* error (`otherErr != nil` → generic not-found), which is the intended degradation.
- [x] Code quality: comments are unusually good — each non-obvious decision (CR exemption, why not generalise, why list beats `/name/`, why `allowPrompt=false` for the speculative lookup) carries its rationale. Doc/code mismatch found at `pro_blueprints.go:833` (finding (2)). Verified the cited `docs/solutions/conventions/classic-api-name-path-encoding` doc actually exists.
- [x] Simplification: two dead exports left by the refactor (finding (5)); no redundant abstraction otherwise — `classicProfileCollection`/`otherProfileType`/`classicProfileLabel` each earn their keep across ≥2 call sites.
- [na] Frontend concerns: no UI, no site (`docs/site/`) changes; no new product namespace, so `make verify-site` is unaffected.
- [x] Documentation currency: `docs/solutions/conventions/classic-profile-payload-ampersand-escaping-2026-07-30.md` was updated thoroughly for the wire model (clause 4) and both new functions — genuinely exemplary. The gap is CLAUDE.md's identifier-convention paragraph (finding (3)), which now describes a rule the PR just added an exception to. Consolidated into one finding per §8.
- [na] Cross-repo contracts: no message/event field, enum, queue name, OpenAPI path, storage key, or config var another repo reads is added or renamed. The PR body references `jamfplatform-go-sdk#48` and `terraform-provider-jamfplatform#309` as *ports of the same wire model*, not as a shared contract this repo publishes — the wire behaviour is Jamf Pro's, and it's pinned locally by `TestNormalizePayloads_OnlyCRReferencesLeftBare` and `TestNormalizePayloads_CRReferencesReachServerBare` rather than by agreement with those repos.
- [x] Project rules compliance: no `.claude/rules/` directory in this repo; checked against CLAUDE.md instead — **Generated Code Boundary** respected (template edited at `generator/classic/generator.go`, `classic_registry.go` regenerated; `make verify-generated` reports "Generated code is up to date"; the edited `classic_payload_escape_test.go` is a hand-written test living in the generated dir, not generator output). **Credential Input Policy** unaffected. **Platform Command Contract** respected — `import-profile` is business logic (multi-call orchestration), correctly hand-written.

</details>

<details>

- 832 additions / 113 deletions across 8 files, reviewed in full (no sampling) at head `f7cbf0e`.
- Related files read for context beyond the diff: `internal/commands/pro/generated/classic_registry.go` (the `resolveClassicNameToIDForApply` body and `classicProfilePayloadFromBody`, to check what the new export actually inherits), `internal/commands/pro/generated/classic_macos_config_profiles.go` (the 4 `normalize`/`verify` call sites, to confirm ordering), `internal/client/client.go` (non-2xx error behaviour), `CLAUDE.md:242` (identifier convention), `docs/solutions/conventions/` (both cited convention docs).
- Verification run locally: `go build ./...` (clean), `make verify-generated` (in sync — 172 modern + 57 classic resource files), `go test ./internal/commands/... ./internal/profileconvert/...` (all pass). `make lint` could not run — `golangci-lint` is not installed in this environment; the PR body reports 0 issues and I have no reason to doubt it, but I did not confirm it.
- **Specialist agents were not fanned out.** Rationale is a property of the input, not of what I found: the diff has no authentication, authorisation, credential, network-boundary, concurrency, or persistence surface (it is string transformation plus CLI argument parsing), and its two halves are small enough to trace exhaustively by hand — which I did, enumerating the escaping state machine's inputs rather than sampling them. A security or performance specialist would have had no in-scope surface to analyse.
- No unresolved assumptions. The one class of claim I could not verify from this repo — Jamf Pro 11.30.2's actual server-side whitespace behaviour — is inherently external; it is documented in `docs/solutions/conventions/classic-profile-payload-ampersand-escaping-2026-07-30.md` clause 4 and the PR body reports live-tenant verification. I treated the server model as given and reviewed only whether the code implements it consistently, which it does.

</details>

<details>

No prior reviews, inline review comments, or issue comments on this PR — `get_reviews`, `get_review_comments`, and `get_comments` all returned empty. First review.

</details>

<details>

✅ **`TestNormalizePayloads_OnlyCRReferencesLeftBare` tests the boundary, not the feature.** It asserts `&#13;` survives bare *and* that `&#38;`/`&#x26;` stay escaped, then simulates the server's single decode and asserts the reconstructed source. That last assertion is what makes the CR exemption safe to own long-term: it pins the invariant ("one decode reproduces the minimised source") rather than the current output string, so a future contributor who tries to generalise the exemption gets a failure that explains itself. Most PRs would have stopped at the first half.

✅ **Deriving the `--type` hint from a real lookup rather than guessing.** `resolveClassicProfileID` only suggests `--type mobile` after confirming a mobile profile with that name exists — and passes `allowPrompt=false` so the speculative call can never stop to ask about a profile the user didn't name. That second detail is the kind of thing that's invisible when right and infuriating when wrong.

✅ **The convention doc was updated with the wire model, not just the code change.** Clause 4 records what the server *does* (LF/TAB deleted, CR the sole survivor, U+2028/29/85 byte-exact, non-BMP unstorable) separately from what this PR does about it. That's the difference between a doc that ages into a lie and one the next contributor can reason from.

</details>

_Generated by `pr-review` v1.18.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Adds ID-based lookup for Classic configuration profiles in `blueprints import-profile` / `components configuration-profile` (issue #315's duplicate-display-name problem), and fixes the send path so `&#13;` line breaks survive a round trip instead of being destroyed.

## Rating: 5/5

Every blocking and nice-to-have finding from the prior automated review (commit `f7cbf0e`) was addressed in the follow-up commit `dfd3068`, and I independently re-verified each fix in the code rather than trusting the commit message — no new issues surfaced.

---

### What I checked

**CR-preservation path** (`generator/classic/generator.go` → regenerated into `classic_registry.go`): traced `minimizeClassicPlistSourceEscaping` → `escapeClassicAmpPreservingCRRefs` → the server's single decode → `classicProfilePayloadFromBody` for `&#13;`, `&#xD;`, `&#013;`, `&amp;#13;` (literal-text case), and the `&#38;`/`&#x26;` 409 boundary. `classicCRRefLen` and the `r == '\r'` branch in `minimizeClassicPlistSourceEscaping` agree on every reference form, including the `semi > 12` bound, hex case variants, zero-padding, and unparseable refs. `TestNormalizePayloads_OnlyCRReferencesLeftBare` pins the invariant ("one server decode reproduces the minimised source") rather than a snapshot string, so a future attempt to generalize the exemption fails with an explanation. `normalizeLineEndings` in `classic_verify.go` folds CR↔LF only — verified `"a\rb"` vs `"ab"` still reports as a line-break-class diff, so it doesn't mask real deletions.

**ID/name resolution** (`pro_blueprints.go`): re-read `resolveClassicProfileID` and `findClassicProfileByName` post-fix. Confirmed:
- The collision error/prompt now takes a caller-supplied `verb` via the exported `ResolveClassicNameToID` + typed `ClassicNameCollisionError`, so `import-profile`'s duplicate-name error says "re-run with one of these IDs" instead of the generated apply path's "use update"/"replace" wording — the exact bug the prior review flagged is gone.
- `components configuration-profile --id abc` now errors immediately (`isClassicID` check before dispatch) instead of silently falling back to a confusing name-not-found error; covered by `TestBlueprintsComponentsConfigProfile_NonNumericIDErrors`.
- `components configuration-profile`'s long help now matches the actual prompt-on-TTY / error-under-`--no-input` behavior.
- CLAUDE.md's Platform identifier-convention paragraph now documents `import-profile`'s hybrid positional (`isClassicID` sniffing) as a named exception with the all-digit-display-name caveat, so the next contributor adding a dual-identifier command has a precedent to follow rather than a silent gap.

**Dead code / test gaps**: `DiffPayloadValues` and `PayloadDiff.String()` (the two unused exports the prior review flagged) are gone — `grep` confirms no remaining references. `formatStoredPayloadWarning` was extracted as a pure function with dedicated tests for the under/at/over-cap truncation boundary (`maxReportedPayloadDiffs = 3`).

**Verification run**: `go build ./...` clean, `go test ./internal/commands/... ./internal/profileconvert/... ./generator/...` all pass, `make lint` reports 0 issues, `make verify-generated` confirms all 172 modern + 57 classic files are in sync with their templates (no hand-edits to generated output).

---

<details>

- [x] Design and architecture: reused `resolveClassicNameToIDForApply` via a parameterized export rather than a parallel resolver — now with caller-supplied wording, so the design correctly serves both the generated apply paths and the hand-written blueprints command.
- [x] Correctness: re-traced the full CR-reference escaping/decoding state machine and the four-way `resolveClassicProfileID` switch (numeric arg, name arg, `--name` flag, both/neither) end to end.
- [x] Security: no credential surface touched (CLAUDE.md credential policy unaffected); bare `&` inside CDATA cannot smuggle malformed XML or an unintended character (`classicCRRefLen` only accepts refs decoding to `\r`); `url.PathEscape` retained on the ID path segment.
- [x] Performance: name lookup uses a full-collection list instead of the old `/name/{name}` GET (unavoidable — that endpoint can't disambiguate duplicates); the wrong-`--type` hint costs one extra list call only on a miss, previously flagged as an acceptable, non-blocking tradeoff and left as-is.
- [x] Test coverage: verified the new/updated tests actually exercise the fixed behavior — collision wording (`TestResolveClassicProfileID_DuplicateNamesListIDs` asserts no "update" wording and the ID-based remedy), `--id` validation, and the truncation boundary tests all pass against the current code.
- [x] Reliability: `client.Do` propagates non-2xx errors rather than reading them as "no match"; the speculative other-type lookup swallows only its own error, which is the intended degradation.
- [x] Code quality: comments consistently carry the "why" for non-obvious decisions (CR exemption, why not generalize, why list beats `/name/`). No further doc/code mismatches found.
- [x] Simplification: no redundant abstractions; the two dead exports flagged previously are removed.
- [na] Frontend concerns: no UI or site changes.
- [x] Documentation currency: convention doc (`docs/solutions/conventions/classic-profile-payload-ampersand-escaping-2026-07-30.md`) and CLAUDE.md both updated to match the shipped behavior.
- [na] Cross-repo contracts: no shared wire field, enum, or config key added/renamed — the PR body's references to `jamfplatform-go-sdk#48`/`terraform-provider-jamfplatform#309` describe the same external (Jamf Pro server) wire behavior, pinned locally by tests, not a contract this repo publishes.
- [x] Project rules compliance: Generated Code Boundary respected (template edited, `classic_registry.go` regenerated, `make verify-generated` clean); Credential Input Policy unaffected; Platform Command Contract respected (`import-profile` is hand-written business logic, correctly not generated).

</details>

<details>

- 43 files changed (1125 additions / 244 deletions net across both commits on this branch), reviewed in full at head `16a8b83`.
- This review incorporates and re-verifies the fixes from the prior automated review (`f7cbf0e`) rather than re-flagging them.
- Verification run locally: `go build ./...`, `go test ./internal/commands/... ./internal/profileconvert/... ./generator/...` (all pass), `make lint` (0 issues), `make verify-generated` (in sync).
- Specialist agents were not fanned out: the diff has no auth, credential, network-boundary, or concurrency surface — it is string transformation plus CLI argument parsing, small enough to trace exhaustively by hand.

</details>

<details>

One prior review (`f7cbf0e0186f3054b4f8b863edf383d47eebb448`, still reachable in the current head — not stale) raised 3 blocking findings and 4 nice-to-haves, rated 3/5. All 7 were addressed in the follow-up commit `dfd3068` ("fix(blueprints): address PR #316 review — collision wording, help text, docs"), independently re-verified above rather than taken on faith. No inline review comments or issue comments existed on this PR.

</details>



_Generated by `pr-review` v1.22.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

PR #321, opened 2026-08-14
feat(specs): sync Jamf Pro API specs to 11.31.0

neilmartin83:
Syncs `specs/` to Jamf Pro **11.31.0**, sourced from the consolidated `/api/schema/` document on a live 11.31.0 instance (`11.31.0-t1785774933693`) and split via `make sync-spec`. 550 paths / 820 operations.

## New endpoints

Three genuinely new operations, all verified against the live instance rather than just `--help`:

| Endpoint | Command | Notes |
|---|---|---|
| `GET /v2/environment-type` | `pro environment-type list` | New resource. Reports the cloud-services environment (staging/production/sandbox) this instance talks to. Returned `{"environment":"production"}`. |
| `GET /v2/smtp-server/allowed-auth-types` | `pro smtp-server allowed-auth-types` | Returned `NONE, BASIC, GRAPH_API, GOOGLE_MAIL`. Pairs with a new upstream note on `SmtpServerV2.authenticationType` that NONE/BASIC may be unavailable per-instance. |
| `GET /v1/pki/digicert/trust-lifecycle-manager/{id}/privilege-check` | `pro digi-cert-settings privilege-check <id>` | 204 when the DigiCert account holds all deployment permissions, 403 listing the missing ones otherwise. Path confirmed live (400 `NOT_FOUND` — no DigiCert config on the test instance). |

The monolith's other four additions — `/devices/extensionAttributes`, `/user`, `/user/updateSession`, `/v1/devices/{id}/groups` — are already covered by `DroppedTags` and need no action.

## Upstream privilege rename

`Read/Update Activation Code` → `Read/Update License Information`, flowing into the `jamf:privileges` annotations on `activation-codes` and `slasas`. No privilege names are hardcoded anywhere else in the CLI, so nothing else changes.

## Handwritten command impact: none

No handwritten command touches activation-code, smtp-server, digicert, or environment-type. `environment-type` correctly stays out of the `backup`/`diff` curated allowlist in `pro_resources.go` — it is read-only instance info, not config.

## Judgement calls in this PR

**Named the new resource `environment-type`, not the auto-pluralized `environment-types`.** The endpoint reports one value rather than offering a choice of types, so the plural reads as "list the available environment types". It is GET-only with no PUT, so `detectSingleton` cannot catch it — hence a `resourceNameOverrides` entry. Same shape as `startup-status`. This is the last moment where the rename is free.

**Stopped the monolith splitter churning `info.title`.** An existing spec file's title was flipping between filename-derived and tag-derived forms depending on whether that particular sync happened to bring the file a new endpoint. Only genuinely new files now take a tag-derived title. Removes 2 junk lines from this diff and stops the flip-flop on every future sync. (`ComputersInventory.yaml` and `PatchSoftwareTitleConfigurations.yaml` still move here — a one-time correction of titles frozen wrong by the 11.30.0 sync.)

**Fixed version tracking on the monorepo sync route.** Neither `make sync-specs` nor `.github/workflows/sync-specs.yaml` wrote `specs/.spec-version`; the workflow wrote `specs/VERSION`, which nothing reads and which was stale at 11.26.0. An Actions-driven sync would therefore have shipped a binary reporting the wrong `specProVersion`. Both now write `.spec-version`, `JAMF_PRO_VERSION` is required on `sync-specs` to match `sync-spec`, and the dead `specs/VERSION` file is removed.

**Rewrote `docs/sync-specs.md`.** It still described the pre-split layout (`specs/VERSION`, `internal/commands/generated/`, `commandGroupMap`, `TestApplyGroups_AllCommandsGrouped`) and had no coverage of the monolith route at all. `CLAUDE.md` updated to match.

## Verification

- `make test` — pass
- `make lint` — 0 issues
- `make verify-generated` — up to date
- `make verify-site` — all products supported
- New commands exercised against a live 11.31.0 instance

## Review follow-ups (round 1)

Addressed in `9878316` — see [the review reply](https://github.com/Jamf-Concepts/jamf-cli/pull/321#issuecomment-5295864225) for detail:

- **`privilege-check` now reports both documented outcomes.** New `documentedStatusResults` generator knob + `registry.WithAllowedStatuses` let a documented non-2xx render through the formatter. Jamf overloads 403 on this endpoint (it also means "token not authorized"), so `BAD_PERMISSIONS` bodies keep the normal exit 5 and API-role hint.
- **`.github/workflows/sync-specs.yaml` calls `make sync-specs`** instead of re-implementing it, so the Action gets the `DeviceExtensionAttribute` rename and the duplicate-basename guard. Verifying that also fixed two defects in the target: `find … -exec cp {} specs/ +` is a hard error on BSD find (macOS), and the duplicate-basename guard could never fire because it checked *after* `cp` had already collapsed duplicates.
- **`generator/monolith` has tests** covering the `info.title` guard.
- **`environment-type` is a singleton** → `pro environment-type get`, not `list`. The `resourceNameOverrides` entry is removed as redundant.
- **`.spec-version` is stamped after `make generate`**, not before.

## Consequence of the version bump

Bumping `specs/.spec-version` to 11.31.0 means `checkTenantVersion` warns on every 11.30.x tenant until it upgrades. Working as designed, but user-visible.

## Previously-known issue, now fixed upstream

The dead `--name` on ~17 generated Pro resources (`pro digi-cert-settings privilege-check --name foo` → HTTP 405) shipped as its own PR, #322, which has since merged. `main` is merged into this branch, so `privilege-check` here is `<id>`-only and that gap is closed rather than deferred.


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Syncs `specs/` to Jamf Pro 11.31.0 (3 new commands, a privilege rename), fixes spec-version tracking, and stabilises monolith `info.title`.
> Blocking: (1), (2), (3). See collapsed section for 5 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 with the `privilege-check` response handling in (1), the workflow/Makefile divergence in (2), and a test for the new splitter guard in (3).
- Coverage: `security-reviewer` and `silent-failure-hunter` have never run on this PR — the rating reflects the dimensions that were searched, not the whole diff.

### Findings

🟧 **(1)** (correctness, via devil-advocate+usability-reviewer, c=100) — `internal/commands/pro/generated/digi_cert_settings.go` (`newDigiCertSettingsPrivilegeCheckCmd`): the new `privilege-check` command cannot report either of its two documented outcomes

`specs/DigiCertSettings.yaml` documents 204 = all permissions present, 403 = body lists the missing permission names. Neither reaches the user. On 204 the body is empty, so `ctx.Output.PrintResponse` → `cliOutput.PrintResponse` (`internal/commands/root.go:78-84`) prints nothing. On 403 `client.Do` (`internal/client/client.go:157-164`) intercepts every status ≥ 400 before returning, and `httpStatusError` (`client.go:417-420`) maps 403 to `exitcode.PermissionDenied` with the hint *"the authenticated account lacks the required API privileges; check its API role"* — which is about the CLI's own token, not the DigiCert account's Trust Lifecycle Manager permissions. The missing-permission list ends up inside an error string rather than through `-o json`/`-o yaml`.

**Failure scenario:** an operator runs `jamf-cli pro digi-cert-settings privilege-check 12` against an under-permissioned DigiCert account → non-zero exit with a "check its API role" hint that blames their own token; against a passing account the command prints nothing at all. Reachable from the PR as shipped.

**Suggested fix:** this is generated code — the change belongs in the template. Add a spec-driven opt-out so an operation whose 403/204 are *documented results* bypasses `httpStatusError` and renders the body through the formatter. Minimal shape in `generator/parser/generator.go` (`resourceTemplate`), gated on the operation declaring a 204 with no content and a 403 with a schema:

```diff
- resp, err := ctx.Client.Do(reqCtx, "GET", path, nil)
- if err != nil {
-     return err
- }
+ resp, err := ctx.Client.DoAllowStatus(reqCtx, "GET", path, nil, http.StatusForbidden)
+ if err != nil {
+     return err
+ }
+ if resp.StatusCode == http.StatusNoContent {
+     return ctx.Output.PrintRaw([]byte(`{"result":"ok","missingPermissions":[]}`))
+ }
```

**Fixed when:** `privilege-check` on a passing account emits a machine-readable success object, and on a failing account emits the missing-permission list through the normal formatter without the misleading API-role hint.

---

🟧 **(2)** (correctness/infrastructure, via devil-advocate, c=100) — `.github/workflows/sync-specs.yaml` (the "Sync specs and regenerate" step): the Action re-implements `make sync-specs` inline and is missing two of its correctness steps

The workflow does `rm -f specs/*.yaml` plus the same two `find … -exec cp` calls, then goes straight to writing `.spec-version` and `make generate`. `Makefile:77-87` runs two more steps in between that the workflow has neither of: the `DeviceExtensionAttribute.yaml` → `MobileDeviceExtensionAttribute.yaml` rename (which exists because upstream's `MobileDeviceExtensionAttribute.yaml` holds only the legacy preview endpoint and would otherwise shadow the canonical `/v1/mobile-device-extension-attributes` CRUD), and the duplicate-basename hard fail. `DroppedTags` in `generator/monolith/overrides.go:25-30` compensates only on the monolith route, not this one. The PR edits exactly this step and the rewritten `docs/sync-specs.md:13-14` now asserts *"Both routes end in `make generate` and both write the version to `specs/.spec-version`"*, presenting the Action as the automated form of Option A.

**Failure scenario:** someone runs `gh workflow run sync-specs.yaml` against a `jamf/jss` ref where `DeviceExtensionAttribute.yaml` exists → the Action-authored PR carries the preview-only spec under the canonical filename, or a duplicate basename silently last-write-wins, where the same sync run locally would have renamed it or aborted. Reachable from the PR as shipped.

**Suggested fix:** have the workflow call the target instead of re-implementing it, so there is one copy of the logic:

```diff
-         run: |
-           rm -f specs/*.yaml
-           find jss -path '*/swagger_docs/uapi/*.yaml' \
-             -not -path '*/uapi/hiddenapi/*' -not -path '*/uapi/common/*' \
-             -exec cp {} specs/ +
-           find jss -path '*/swagger_docs/uapi/common/*.yaml' \
-             -exec cp {} specs/ +
-           echo "$JAMF_PRO_VERSION" > specs/.spec-version
-           make generate
-           go fmt ./internal/commands/pro/generated/...
+         run: |
+           mkdir -p sync-root && mv jss sync-root/jamf-pro-server
+           make sync-specs JAMF_SERVER_PATH=sync-root JAMF_PRO_VERSION="$JAMF_PRO_VERSION"
+           go fmt ./internal/commands/pro/generated/...
```

**Fixed when:** an Action-triggered sync against a ref containing `DeviceExtensionAttribute.yaml` or a duplicate basename produces the same result (rename applied, or hard failure) as `make sync-specs` locally.

---

🟧 **(3)** (test coverage, via test-quality-reviewer, c=75) — `generator/monolith/split.go:82-117`: the PR's only behavioural change to the splitter has no test, and a mutation survives the whole suite

`generator/monolith` contains no `*_test.go` file, and `monolith.Split` is called only from `generator/main.go`. Replacing the new guard `if !existingFiles[filename]` with `if true` — i.e. reinstating the exact `info.title` churn the guard was added to prevent — was applied in the review worktree: `go build ./...` succeeded and `go test ./generator/...` passed in full. Nothing detects it.

**Failure scenario:** a later refactor or revert of the guard reintroduces title flapping on every `make sync-spec` where an existing spec file picks up one new path, and no test fails. Reachable from the PR as shipped.

**Suggested fix:** add `generator/monolith/split_test.go` with two cases — (a) seed `specsDir` with an existing file whose `info.title` is filename-derived, feed a monolith adding one new path under that file's tag, assert the title is unchanged; (b) a genuinely new file, assert its title is tag-derived.

**Fixed when:** a test in `generator/monolith` fails when the `existingFiles[filename]` guard is removed.

<details>

🟩 **(4)** (usability, via usability-reviewer, c=100) — `internal/commands/pro/generated/digi_cert_settings.go`: the new `privilege-check --name` flag can only ever return HTTP 405.

`resolveNameToID(..., "/v1/pki/digicert/trust-lifecycle-manager", ...)` GETs a collection path that `specs/DigiCertSettings.yaml:141` defines only a `POST` for. The PR description discloses this as a pre-existing `opHasNameLookup`/`collectionPath` defect across ~17 resources, deferred to its own PR — this command just adds one more instance. Deferral is reasonable; the disclosure is appreciated.

*Promote to 🟧 when:* the `opHasNameLookup` follow-up does not land before the next release.

🟩 **(5)** (usability, via devil-advocate+usability-reviewer, c=75) — `internal/commands/pro/generated/environment_type.go:29-36`: the `resourceNameOverrides` entry fixed the noun but not the verb.

`ApplyNameOverrides` (`generator/parser/parser.go:41-51`) sets only `Name`/`NameSingular`/`GoName`, never `IsSingleton` — which is what drives the list→get rename. So the command still reads `Use: "list"` with `# List all environment-type` and `--field id`, and `specs/EnvironmentType.yaml` has exactly one property (`environment`) and no `id`, so that example prints nothing. `startup_status.go` has the identical shape, so this is a template/`detectSingleton` gap, not a one-off — worth a follow-up that teaches `detectSingleton` about GET-only settings resources.

🟩 **(6)** (reliability, c=75) — `Makefile:63` (and `:128`): `.spec-version` is stamped before the work it certifies.

`@echo "$(JAMF_PRO_VERSION)" > specs/.spec-version` runs before `rm -f specs/*.yaml`, the `cp` calls, and `$(MAKE) generate`. Each recipe line is its own shell with no rollback, so a mid-run failure leaves the stamp claiming the new version while `specs/` and the generated package hold the old set — and `SPEC_PRO_VERSION`/`LDFLAGS` (`Makefile:7-8`) bake that into `jamf-cli version` on the next `make build`. Make fails loudly, so a developer has to ignore the error to be bitten.

```diff
- @echo "$(JAMF_PRO_VERSION)" > specs/.spec-version
  @rm -f specs/*.yaml
  ...
  @$(MAKE) generate
+ @echo "$(JAMF_PRO_VERSION)" > specs/.spec-version
```

🟩 **(7)** (test coverage, via test-quality-reviewer, c=75) — `generator/parser/parser_test.go` (`TestApplyNameOverrides`): the new `"environment-types"` → `"environment-type"` entry isn't directly asserted.

The fixture only feeds `"computers-inventories"`. It is caught transitively by `TestApplyProGroups_AllCommandsGrouped`, but a one-line fixture addition would catch a typo in either the key or the value directly.

🟩 **(8)** (documentation, c=75) — `docs/sync-specs.md` ("After either route"): the checklist doesn't mention privilege-name renames.

This sync silently changed every `jamf:privileges` annotation on `activation-codes` and `slasas` (`Read/Update Activation Code` → `Read/Update License Information`). Per `CLAUDE.md` those strings are surfaced verbatim in `commands -o json` and appended to the 403 hint, so a downstream consumer can break on a routine sync. Add a step alongside "Group any new commands" and "Sanity-check auto-derived resource names": scan `git diff -- specs/` for `x-required-privileges` changes and call them out in the PR body.

</details>

<details>

`scope-reviewer` recommends **splitting into 2 PRs**. The diff bundles maintenance to the spec-sync *pipeline* with the spec *content* sync, and the two have different reviewers and risk profiles. It also removes a confound: `specs/.spec-version` is bumped to `11.31.0` using the new Makefile mechanism this same PR introduces, so reviewing that one-line bump currently requires first understanding the mechanism change.

**PR 1 — Spec-sync pipeline maintenance** (merge first, no dependencies): `Makefile` (require `JAMF_PRO_VERSION`), `.github/workflows/sync-specs.yaml`, delete `specs/VERSION`, `generator/monolith/split.go` (the title-churn guard), `docs/sync-specs.md`, `CLAUDE.md`. Mergeable when both targets fail fast without `JAMF_PRO_VERSION`, the splitter's title behaviour is confirmed on a fixture, and docs match. Touches no generated code, so `make verify-generated` is a no-op. Findings (2) and (3) land here.

**PR 2 — Spec sync to Jamf Pro 11.31.0** (depends on PR 1): `specs/.spec-version` bump, the three new/changed specs plus their generated commands, the `ActivationCode`/`Slasa` privilege rename, `ComputersInventory` and `PatchSoftwareTitleConfigurations` title corrections, and the regenerated `provenance`/`registry`/`backup_registry`/`smoke_registry` bookkeeping. `generator/parser/parser.go` (`resourceNameOverrides`) and `internal/commands/groups.go` stay here, not in PR 1 — the override exists only because this sync introduced `EnvironmentType`, and reviewing it in isolation would leave "override for what resource?". Findings (1), (4), (5) land here.

Coordination: the two touch disjoint files except `specs/.spec-version`, `docs/sync-specs.md`, and `CLAUDE.md`, where PR 2's edits are additive on PR 1's rewrite — the rebase should be mechanical. If PR 1 lands first, re-run the sync to regenerate PR 2 rather than hand-patching.

Given the PR is already verified end-to-end against a live 11.31.0 instance and each half is individually small, shipping as one is defensible — treat this as a preference, not a blocker.
</details>

<details>

- [x] Design/architecture: the `resourceNameOverrides` vs `detectSingleton` choice is finding (5); generated-vs-handwritten boundary respected (no edits under `internal/commands/pro/generated/` outside regeneration)
- [x] Correctness: traced `privilege-check`'s 204/403 through `client.Do` → `httpStatusError` → `PrintResponse` — finding (1); verified `split.go`'s title assignment (line 181 vs 182-183) makes existing files permanently filename-derived, so the `ComputersInventory`/`PatchSoftwareTitleConfigurations` title moves are a genuine one-time correction, not a new flip-flop
- [x] Cross-repo/contract: `x-required-privileges` renames flow only into `jamf:privileges` annotations; no privilege string is hardcoded elsewhere in the CLI (grepped). `specs/VERSION` has no remaining reader (grepped `specs/VERSION` across Go, YAML, Makefile, Markdown — zero hits)
- [x] Reliability: `.spec-version` stamp ordering — finding (6); confirmed `SPEC_PRO_VERSION` is an immediate `:=` read at Makefile parse time and that the sub-`make generate` re-reads it, so the sub-make path is sound
- [x] Test coverage: mutation-tested the new splitter guard — finding (3); confirmed `TestApplyProGroups_AllCommandsGrouped` covers the `groups.go` entry and passes
- [x] Infrastructure/CI: `.github/workflows/sync-specs.yaml` — finding (2)
- [x] Documentation currency: `docs/sync-specs.md` rewrite read in full and checked against the Makefile targets it describes — accurate except finding (8); `CLAUDE.md` command lines match the new required vars
- [x] Usability/DX: new command surface — findings (1), (4), (5). Requiring `JAMF_PRO_VERSION` on `make sync-specs` is a breaking change to a maintainer-only target, documented in both `CLAUDE.md` and `docs/sync-specs.md` — acceptable
- [x] Project rules: `CLAUDE.md` read (no `.claude/rules/` directory present). Generated-code boundary, generator-template-ownership, and site-verification rules all respected
- [na] Security: no auth, credential, dependency-manifest, or secret-handling change; the workflow's `permissions:` and `secrets:` blocks are untouched
- [na] Performance: no query, loop, or hot-path change
- [na] Frontend: no UI or site changes
- [?] Runtime version warning: bumping `.spec-version` to 11.31.0 means `checkTenantVersion` (`internal/commands/root.go:263-304`) will now warn on every 11.30.x tenant. This is the mechanism working as designed, but it is a user-visible consequence of the bump that nothing in the PR body mentions
</details>

<details>

Full diff (25 files, +443/−70) at head `dda5e6f`, three-dot against `origin/main`. Beyond the diff: `internal/client/client.go`, `internal/commands/root.go`, `internal/commands/version.go`, `internal/commands/pro_resources.go`, `generator/monolith/split.go` in full, `generator/monolith/overrides.go`, `generator/parser/parser.go`, `internal/commands/pro/generated/startup_status.go`, and `docs/sync-specs.md` in full.

Provenance SHA256 entries and the four regenerated registry files were spot-checked for consistency with the spec changes rather than line-by-line — they are generated bookkeeping and `make verify-generated` is the real gate (the author reports it clean; CI enforces it).

Dispatched this round: `usability-reviewer`, `devil-advocate`, `test-quality-reviewer`, `scope-reviewer`. Never-run lanes: `security-reviewer` (eligible — the diff touches a CI workflow and build config — not dispatched; no credential, dependency, or auth surface changed on inspection), `silent-failure-hunter` (eligible — `split.go`'s new guard is a filter — not dispatched; the orchestrator pass traced that guard directly instead), `performance-reviewer` (inapplicable — no query, loop, or hot-path file changed), `fidelity-reviewer` (inapplicable — no linked issue and no Jira-style key in the title, body, or branch name, so there are no acceptance criteria to check against).

One finding was dropped by the confidence pass: a non-determinism risk in `newFileTag` when two distinct new tags route into the same new filename scored `c=50` and is unreachable today (`TagFilenameOverrides` is empty).
</details>

<details>

No prior reviews, review comments, or issue comments on this PR.
</details>

<details>

✅ The `info.title` stabilisation in `split.go` is the right fix in the right place — verified that existing buckets now take the filename-derived title unconditionally (`split.go:181`), so the churn is genuinely eliminated rather than suppressed for one sync, and the `ComputersInventory`/`PatchSoftwareTitleConfigurations` moves in this diff are the last of it.

✅ The PR body's "Known issue found here, deliberately not fixed" section is exactly the right way to handle the dead `--name` flag: root cause identified (`opHasNameLookup` gating on `collectionPath() != ""`), blast radius measured (~17 resources, `apply` affected the same way), and scope-deferred with a reason. Reviewing this took minutes instead of hours as a result.
</details>

neilmartin83:
@ktn-jamf — all three blocking findings and four of the five nice-to-haves addressed in `9878316` (pushed). Finding (1)'s suggested fix turned out to be insufficient once tested live; details below.

## Blocking

**(1) 🟧 `privilege-check` cannot report either documented outcome** — fixed, but not as suggested.

The mechanism: a new generator knob `documentedStatusResults` (`generator/parser/parser.go`) marks operations whose non-2xx status is a documented result. The template carries those statuses through a new `registry.WithAllowedStatuses` context override — the same pattern as the existing `WithAccept`/`WithContentType` overrides — so `client.Do` returns the response with its body instead of mapping it to an exit-code error. No `HTTPClient` interface change, so none of the ~16 test doubles implementing it had to move. The 204 now synthesizes `{"result":"ok","detail":…}` from the spec's own 204 description.

It's an explicit allowlist rather than a spec-shape rule (your suggestion gated on "declares a 204 with no content and a 403 with a schema"): nearly every Jamf operation documents a 403 as boilerplate, and 204-only success is the shape of every `delete`, so that gate would have made a large number of commands swallow genuine permission failures. Same reasoning as `TagFilenameOverrides` and `resourceNameFieldOverrides` — the specs can't express this, so the map does.

**The part your finding didn't anticipate:** Jamf overloads 403 on this endpoint. Running the naive version against a tenant whose token lacks `Read DigiCert Settings`:

```
$ jamf-cli pro digi-cert-settings privilege-check 12 -o json
{"httpStatus":403,"errors":[{"code":"BAD_PERMISSIONS","description":"The given token was not authorized to access the requested resource."}]}
{"error":"general","exitCode":1,"message":"DigiCert account is missing one or more required permissions."}
```

That's the *token* being unauthorized, reported as a DigiCert-account problem, with the exit code dropped from 5 to 1 — one misleading message traded for another, in the more common case. So `renderDocumentedStatus` classifies the body: a `BAD_PERMISSIONS` sentinel is re-raised through the newly exported `client.StatusError` (so the error text and hint stay in one place, not duplicated), and anything else renders through the formatter with the spec's wording and a non-zero exit. Same command after the fix:

```
EXIT=5
{"error":"permission_denied","exitCode":5,
 "hint":"the authenticated account lacks the required API privileges; check its API role. Required privilege(s): Read DigiCert Settings", …}
```

Caveat worth stating plainly: the *result* branch is covered by unit tests only. The test tenant has no DigiCert Trust Lifecycle Manager configuration, so I could not observe a real "missing DigiCert permissions" 403 to confirm it doesn't also carry `BAD_PERMISSIONS`. If it does, the branch is unreachable and the discriminator needs revisiting — flagging rather than claiming otherwise.

**(2) 🟧 workflow re-implements `make sync-specs`** — fixed. The workflow calls the target now. Rather than the `mv jss sync-root/jamf-pro-server` dance, `JAMF_SERVER_ROOT` became an overridable variable (documented in the Makefile), so the Action's flat checkout points at it directly.

Verifying that path end to end in a throwaway worktree surfaced two defects in the target the workflow was about to start depending on:

- **`find … -exec cp {} specs/ +` is a hard error on BSD find**, i.e. on macOS, where maintainers actually run this: the batching form requires `{}` to be the final argument. `make sync-specs` fails with `find: -exec: no terminating ";" or "+"` before copying anything. Switched to `-exec … ';'`. Pre-existing — this PR used the monolith route, which is presumably why it went unnoticed.
- **The duplicate-basename guard could never fire.** It ran `ls specs/*.yaml | … | uniq -d` *after* the copy, by which point two upstream modules shipping the same filename have already collapsed into one file. Confirmed by seeding two modules with the same basename: the guard passed silently. It now validates the source tree before anything is copied, which also means a rejected sync leaves the committed `specs/` untouched rather than emptied. Your finding said the local run "would have renamed it or aborted" — it would have renamed, but never aborted.

Both verified in the worktree: the rename step fires (`Renamed DeviceExtensionAttribute.yaml → MobileDeviceExtensionAttribute.yaml`), a duplicate basename now aborts with `specs/` unmodified, and the happy path completes.

**(3) 🟧 splitter guard untested** — added `generator/monolith/split_test.go` with your two cases plus a third (a new path joins its existing tag owner rather than fragmenting, and is still reported in `warnings`). Your mutation is caught:

```
$ perl -pi -e 's/if !existingFiles\[filename\] \{/if true {/' generator/monolith/split.go
--- FAIL: TestSplit_ExistingFileKeepsFilenameDerivedTitle
    title = "Jamf Pro API - buildings", want "Jamf Pro API - Buildings"
```

## Nice-to-haves

**(5) 🟩 the override fixed the noun but not the verb** — fixed for the new resource. New `readOnlySingletonPaths` allowlist makes a GET-only, no-`{id}` path a singleton, so it's `pro environment-type get` with a `-o yaml` example instead of `list --field id`. Kept as an allowlist rather than teaching `detectSingleton` the general rule, because the general rule renames the existing `startup-status`/`ddm-status`/`apns-client-push-status` commands from `list` to `get` — a user-visible break that deserves its own PR. Follow-up territory, as you said.

Knock-on: singletons skip pluralization, so the `resourceNameOverrides` entry for `environment-types` is now dead code and is removed — which **supersedes (7)**, since there's no longer an entry to assert. Added a `detectSingleton` table case for the read-only shape instead, plus a `TestApplyDocumentedStatusResults` unit test for (1)'s wiring.

**(6) 🟩 `.spec-version` stamped before the work it certifies** — fixed in both targets; the stamp is now the last line after `make generate`. Confirmed live: the failed sync run left `.spec-version` at its old value.

**(8) 🟩 checklist missing privilege renames** — added, with the `git diff -- specs/ | grep x-required-privileges` command, plus a new step for spotting endpoints that document a non-2xx as a result. `CLAUDE.md`'s "Where to Make Changes" table gained rows for both new knobs.

## Not taken

**PR decomposition** — keeping it as one. The pipeline changes and the spec content are now more entangled than at review time, not less: the `find`/dupe-guard fixes were found *by* verifying the sync route this PR edits, and finding (1)'s generator knob exists only because 11.31.0 introduced the endpoint. Splitting now means re-running the sync and re-doing the live verification for no reviewer benefit.

**The `[?]` note on version warnings** — fair point, and it wasn't in the body. Adding to the PR description: bumping `.spec-version` to 11.31.0 means `checkTenantVersion` warns on every 11.30.x tenant until they upgrade. That's the mechanism working as designed, but it is the user-visible consequence of the bump.

## Verification

`go test ./...` pass, `make lint` 0 issues, `make verify-generated` up to date, `make verify-site` all products supported. Live against 11.31.0 (`tfproviderdev` via the platform gateway): `privilege-check` on a token-privilege 403 → exit 5 with the API-role hint and the required privilege named; `environment-type get` → correct verb, help, and example.

ktn-jamf (commented):
## Round 2 review — re-review of the round-1 fixes

Scope: `9878316` "fix: address PR #321 review — privilege-check output, sync parity, splitter test", plus the `5604b67` merge of `main`. Head reviewed: `5604b67`. The round-1 review is at [#pullrequestreview-4938491454](https://github.com/Jamf-Concepts/jamf-cli/pull/321#pullrequestreview-4938491454).

**Verdict: ✅ All three round-1 findings are fixed. Rating 4/5.** One new low-severity item in the workflow, plus one design question worth a decision.

### Round-1 findings

| # | Round-1 finding | Status |
|---|---|---|
| (1) | `privilege-check` reported neither documented outcome | ✅ Fixed |
| (2) | Workflow re-implemented `make sync-specs` and dropped two guards | ✅ Fixed |
| (3) | Splitter guard change had no tests | ✅ Fixed |

**(1)** `newDigiCertSettingsPrivilegeCheckCmd` now opts the request out of the client's error mapping with `registry.WithAllowedStatuses(reqCtx, 403)`. A 204 synthesizes a result payload, so a passing check prints something in every output format. A 403 renders the body through `renderDocumentedStatus` and exits non-zero, so a script can branch on it. A 403 that is really a token-authorization failure keeps exit code 5 and its usual hint. Four tests cover the paths, including `TestPrivilegeCheck_TokenFailureStaysPermissionDenied`.

**(2)** The workflow now calls `make sync-specs JAMF_SERVER_ROOT=jss` ([sync-specs.yaml:59](.github/workflows/sync-specs.yaml:59)) instead of re-implementing the copy loop, so it inherits the `MobileDeviceExtensionAttribute` drop, the `DeviceExtensionAttribute.yaml` rename, and the duplicate-basename hard-fail. The comment above the step records why the target is called rather than reproduced. This is the right shape of fix — the docs claim of equivalence is now true by construction rather than by parallel maintenance.

**(3)** `TestSplit_ExistingFileKeepsFilenameDerivedTitle` and `TestSplit_NewFileTakesTagDerivedTitle` now cover the guard. Mutation-verified: replacing `if !existingFiles[filename]` with `if true` fails the first test with a precise message about the `/v1/buildings/export` path flipping the title. The round-1 complaint was that an inverted guard survived the whole package; it no longer does.

Round-1 nice-to-have (6) is also fixed: `specs/.spec-version` is now stamped after `generate` ([Makefile:105-108](Makefile:105)), with a comment explaining that a mid-run failure must not leave a stamp certifying specs that were never produced. I did not individually re-verify each of the other round-1 🟩 items.

### Verification

- `make test` — full suite passes.
- `make verify-generated` — generated code up to date (174 modern, 57 classic resource files).
- Mutation check on finding (3), described above. The worktree was left clean.

### New items

**🟩 (a) `JAMF_PRO_VERSION` reaches a shell through two substitution passes.** The workflow passes the unvalidated `workflow_dispatch` string into `make`, which substitutes `$(JAMF_PRO_VERSION)` textually into double-quoted recipe lines such as `@echo "$(JAMF_PRO_VERSION)" > specs/.spec-version` ([Makefile:108](Makefile:108)). The shell then evaluates the result. I confirmed the mechanism in an isolated Makefile: a value of `1.2.3"; touch INJECTED; echo "` runs `touch`. The env-var indirection at [sync-specs.yaml:28](.github/workflows/sync-specs.yaml:28) correctly blocks GitHub Actions `${{ }}` injection, but it does nothing about Make's own pass.

Severity is low rather than high, for a reason worth stating: `workflow_dispatch` requires repository write access, and an actor with write access can generally run arbitrary code in CI by pushing a branch with a modified workflow. The injection adds little in that model. It does matter where branch protection stops a write-access user from pushing to the default branch while still allowing dispatch — there it reaches `contents: write`, `pull-requests: write`, and `secrets.SERVER_REPO_TOKEN` without review. Treat it as cheap defense in depth: validate against `^[0-9]+\.[0-9]+\.[0-9]+$` in the workflow, and reject a non-conforming value in the Makefile before any recipe line uses it.

**🟩 (b) Actions are pinned to mutable major tags.** `actions/checkout@v7`, `actions/setup-go@v7`, and `peter-evans/create-pull-request@v8` run in a job holding write permissions and `SERVER_REPO_TOKEN`. Pin to full commit SHAs.

**🟩 (c) Design question: the 403 classifier fails open.** `isTokenAuthorizationError` ([registry.go](internal/commands/pro/generated/registry.go)) treats a 403 as the endpoint's documented answer unless the body carries an `errors[].code == "BAD_PERMISSIONS"` sentinel. An empty body, an HTML proxy page, or a renamed error code therefore prints as "DigiCert account is missing one or more required permissions" at exit 1, instead of the auth error and exit 5.

This is a deliberate choice, not an oversight: `TestIsTokenAuthorizationError` covers the empty-body and non-JSON cases explicitly and asserts `false` for both, and a comment states the rule. So the question is only whether the default is the one you want. Failing closed — unparseable or `errors`-free bodies classified as token failures — would keep exit 5 correct for infrastructure 403s at the cost of misreporting a documented result that arrives in an unexpected envelope. Either is defensible; the current one is at least tested and documented.

### Correction to an earlier note

An intermediate lane result claimed no test exercised a non-`BAD_PERMISSIONS` or non-JSON 403 body. That is wrong — `documented_status_test.go:160-161` covers both. Item (c) above is framed as a design question rather than a coverage gap because of it.

ktn-jamf (approved):

PR #322, opened 2026-08-14
fix(generator): only generate --name/apply when the collection serves GET

neilmartin83:
Follow-up to the known issue called out in #321. Independent of it — this branch is off `main` and touches no specs.

## The bug

`--name` (and `--serial`/`--udid`, and all of `apply`) resolve an identifier exactly one way: GET the resource's collection, RSQL-filter it, extract the ID, act on `/{id}`. The three gates that decide whether to emit them — `opHasNameLookup`, `patchHasLookup`, `shouldGenerateApply` — all tested `collectionPath() != ""`.

`collectionPath` answers a *different* question — "what is this resource's collection URL" — and to answer it falls back to the create-POST path, and then to the `{id}` path minus its last segment. Neither implies the server will answer a GET there.

So any resource whose modern API is `POST /collection` + `GET /collection/{id}` got a `--name` flag that could only ever fail. Wire-confirmed against a live 11.31.0 instance:

```
pro dock-items get --name Foo              → looking up "Foo": request failed (HTTP 405)
pro venafis get --name Foo                 → looking up "Foo": request failed (HTTP 405)
pro cloud-ldaps get --name Foo             → looking up "Foo": request failed (HTTP 405)
pro adcs-settings get --name Foo           → looking up "Foo": request failed (HTTP 405)
pro icons get --name Foo                   → looking up "Foo": request failed (HTTP 405)
pro certificate-authorities get --name Foo → resource not found (HTTP 404)
pro computer-groups get --name Foo         → resource not found (HTTP 404)
```

The 405s are the create-POST fallback (collection exists, POST only). The 404s are the derived-parent fallbacks, where the parent path does not exist server-side at all.

`apply` was worse than a dead flag — name-based upsert *is* its contract, so on those resources it could never do the "update if it exists" half.

## The fix

Split the question in two. `collectionPath` keeps its meaning and its fallbacks; `listPathFromOps` and `dedupeOperations` legitimately want the structural answer. A new `nameResolutionPath(r)` answers "which URL can we GET and RSQL-filter": take `collectionPath`'s candidate, then require that some operation on the resource actually performs a GET against that exact path.

An explicit `resourceNameLookupPathOverrides` entry is hand-picked and wire-verified, so it wins outright without the check — that is how `mobile-devices` → `/v2/mobile-devices/detail` keeps working.

All three gates and both template path helpers (`nameLookupPath`, `lookupFieldPath`) now use it.

## Blast radius

40 dead lookup sites and 7 unusable `apply` commands removed across 16 resources; 1634 → 1627 commands. Affected resources keep full ID-based CRUD, and `get`/`update`/`delete` tighten from `MaximumNArgs(1)` to `ExactArgs(1)`, so omitting the ID is now a clean usage error instead of "provide an `<id>` argument, --name".

No registry changed — `backup`/`diff` and the smoke tests key off paths, not lookups.

### ⚠️ Breaking

`apply` is removed from `adcs-settings`, `cloud-azures`, `cloud-ldaps`, `digi-cert-settings`, `dock-items`, `team-viewer-remote-administrations`, `venafis`. It never worked on them — its name-to-ID step was the 405 above — so anything scripted against it was already failing. Use `create` plus `update`/`patch <id>`.

## Why not make the lookup work instead

For some of these a sibling endpoint might plausibly serve the lookup, and `resourceNameLookupPathOverrides` exists for exactly that — per-resource, deliberate, verified against the wire. What the generator must not do is infer a lookup endpoint from a POST route and ship a flag that cannot work. Removing the false affordance is the right default; adding overrides is follow-up work, one resource at a time.

## Verification

Positive path confirmed intact against a live instance, not just unit tests — created a throwaway category, then:

- `categories get --name` resolved it
- `categories apply` correctly **updated** it (id 5884, priority 9 → 3) rather than creating a duplicate
- `categories delete --name` removed it

Negative lookups on kept resources (`scripts`, `buildings`, `mobile-devices`, `computers-inventory --serial`) return a clean "no resource found", i.e. the GET+RSQL round trip still happens — no 405s.

`make test`, `make lint` (0 issues), `make verify-generated`, `make verify-site` all pass.

## Tests

- `TestNameResolutionPath` — new. Each case asserts **both** `collectionPath` and `nameResolutionPath`, so a future change that collapses the two back together fails here rather than silently resurrecting the dead flags. Mutation-checked: reverting the gates to `collectionPath` makes the new cases fail.
- `TestOpHasNameLookup` — added POST-only-collection, derived-parent-with-no-GET, and override-wins cases.
- `TestShouldGenerateApply` — added a POST-only-collection case.
- Four pre-existing tests (`TestGenerate_ApplyWithDisplayName`, `TestNeedsFmt_WithApply`, `TestNeedsURL_WithApply`, `TestShouldGenerateApply`) built fixtures with `create`+`update` and no list GET, then asserted apply was generated — a resource shape that cannot support apply. Each fixture gained the list GET it was implicitly assuming.

Postmortem: `docs/solutions/logic-errors/name-lookup-requires-get-serving-collection-2026-08-14.md`


ktn-jamf (commented):
> [!TIP]
> ✅ **Merge-ready** — Stops the generator emitting `--name`/`--serial`/`--udid` and `apply` for resources whose collection does not answer GET. No critical or important issues found.
> See collapsed section for 3 nice-to-have suggestions.

## Rating: 5/5
- Clean — the fix is correct, the regenerated output matches (`make verify-generated` clean, full `go test ./...` green in a fresh worktree at `dccb4e1`), and the 3 suggestions below are optional test/UX follow-ups that don't hold the merge.

### Findings

<details>

🟩 **(1)** (test-quality, via test-quality-reviewer, c=75) — `generator/parser/generator.go:1267` (`patchHasLookup`), `generator/parser/generator_test.go:2381` (`TestPatchHasLookup`): the third gate is the one gate no test distinguishes.

`opHasNameLookup` and `shouldGenerateApply` each gained a POST-only-collection case that fails when the gate is reverted to `collectionPath`. `patchHasLookup` got the same code change and no case: reverting line 1267 to `collectionPath(r.Operations) == ""` leaves all four `TestPatchHasLookup` subtests **and the whole `generator/parser` package** passing (mutation-confirmed). Its existing "no list path" fixture has only a `patch` op, so `collectionPath` is already `""` there — it cannot discriminate the two functions. This matters concretely: `venafis` (`POST /v1/pki/venafi` + `PATCH /v1/pki/venafi/{id}`) is a live resource whose `patch` lookup removal depends on exactly this gate.

**Suggested fix:** add a `TestPatchHasLookup` case mirroring `TestOpHasNameLookup`'s new one — `create` POST `/v1/things` + `patch` `/v1/things/{id}`, no collection GET, `want: false`.

**Fixed when:** a test fails if `patchHasLookup`'s gate is reverted to `collectionPath(r.Operations) == ""`.

**Promote to 🟧 when:** a spec sync reintroduces a POST-only-collection resource with PATCH and the dead flag ships unnoticed.

🟩 **(2)** (test-quality, via test-quality-reviewer, c=75) — `generator/parser/generator.go:342-356` (`nameLookupPath`, `lookupFieldPath`): both helpers were rewired by this PR and neither has any coverage.

No test calls either helper, and no `Generate()`-based test asserts the URL literal they emit into `resolveNameToID`/`resolveNameToIDForApply` (`TestGenerate_GetNameFlag` asserts only that `resolveNameToID` appears, never its path argument). No fixture anywhere sets `LookupFields`, so `lookupFieldPath`'s `?section=` branch — the `--serial`/`--udid` path used by `computers-inventory` and `mobile-devices` — is entirely unexercised.

**Suggested fix:** one `Generate()` test with a `LookupFields` entry carrying a `Section`, asserting the emitted `resolveNameToID(..., "<path>?section=<Section>", ...)` string; a companion case with a `NameLookupPath` override asserting the override value (not `collectionPath`'s) reaches the generated call.

**Fixed when:** a test fails if `lookupFieldPath`'s base swaps back to the pre-PR fallback or the `?section=` append is dropped.

🟩 **(3)** (usability, via usability-reviewer, c=75) — `internal/commands/pro/generated/dock_items.go:25`: the removal is invisible to CLI users.

A user who scripted `pro dock-items apply` now gets `unknown command "apply"`; `--name` gets `unknown flag`. `pro classic-dock-items` still supports both and is the working alternative, and the other six apply-losing resources have no Classic sibling at all — but the explanation lives only in `docs/solutions/`, which CLI users never read, and `Long` is the generic "Manage dock-items in Jamf Pro." (Errors are at least clean: `root.go`'s "did you mean" hint correctly stays silent, verified by running the built binary.)

**Suggested fix:** in the resource template, when `collectionPath(r) != ""` but `nameResolutionPath(r) == ""`, append one line to the parent command's `Long`: "This resource has no listable collection — supply the numeric ID (`--name` is unavailable)."

**Fixed when:** `jamf-cli pro dock-items --help` states the ID-only constraint.

</details>

<details>

**Recommendation: deliver as a single PR** (via `scope-reviewer`).

The core change is one logic fix — `nameResolutionPath()` replacing `collectionPath()` in three gates — with a single done-state. The 16 regenerated files under `internal/commands/pro/generated/` are the mechanical output of that change and are CI-pinned to it by `make verify-generated`, so splitting them breaks CI on both halves. Tests, the CLAUDE.md paragraph, and the postmortem are directly tied artifacts and too small to justify a separate review pass. No independently mergeable sub-concern exists.
</details>

<details>

- [x] Design and architecture: read `collectionPath`'s four-branch resolution (`generator.go:827-856`) against the new `nameResolutionPath`; the "exact candidate" strictness is deliberate and correct — see Scope of review for the alternative I tested and rejected
- [x] Correctness: traced all three gates and both template helpers; confirmed every `nameLookupPath`/`lookupFieldPath` template site (`generator.go:1986,1993,2161,2168,2191,2200,2212,2221`) sits inside a matching gate, so no empty path can reach the wire; spot-checked regenerated `classic_ldaps.go`, `computer_groups.go`, `icons.go`, `dock_items.go`
- [na] Security: no auth, credential, dependency, or config surface touched; credential-input policy in CLAUDE.md unaffected
- [na] Performance: no hot paths, queries, or loops changed; the new GET scan is over a handful of ops at generate time
- [x] Test coverage: `TestNameResolutionPath` asserts both functions per case, which is the right pin; gaps are findings (1) and (2); the four pre-existing fixtures gained a list GET, which is a realism fix (a `create`+`update`-only resource genuinely cannot upsert), not a weakened assertion
- [x] Reliability: the new guard fails closed; no error paths added or swallowed; `ExactArgs(1)` is an equivalent guard for the removed `provide an <id> argument, --name` error
- [x] Code quality: generated `var ()` empty blocks and the lost blank line before `return cmd` are pre-existing template artifacts (152 occurrences on `main`, 175 here), so not flagged per the repo's own convention
- [na] Simplification: no redundant abstraction introduced; `collectionPath` correctly keeps its structural callers (`dedupeOperations`, `listPathFromOps`)
- [x] Documentation currency: CLAUDE.md gained an accurate paragraph on the new invariant (`CLAUDE.md:200`) and a postmortem lands in `docs/solutions/logic-errors/`; `docs/site` needs no edit (commands.json is generated at deploy)
- [na] Cross-repo contracts: no wire key, enum, endpoint, or shared config name changed; the specs under `specs/` are untouched
- [x] Project rules compliance: checked CLAUDE.md's generated-code boundary (only templates edited, generated files regenerated, `make verify-generated` clean) and its build/test commands; no `.claude/rules/` directory exists

</details>

<details>

Full diff (20 files, +514/−3222) at head `dccb4e1` in a dedicated worktree, plus `generator/parser/parser.go` (`resourceNameLookupPathOverrides`), the affected specs under `specs/`, `registry.go`/`backup_registry.go`/`smoke_registry.go`, and the regenerated command files. Ran `go build ./...`, `go test ./...` (all pass), `make verify-generated` (up to date). Lanes dispatched: `usability-reviewer`, `test-quality-reviewer`, `devil-advocate`, `silent-failure-hunter`, `scope-reviewer`. Never run: `security-reviewer`, `performance-reviewer` (no lane — no auth/dep/config or hot-path files changed), `fidelity-reviewer` (no ticket or acceptance criteria; #321 is a sibling PR, and this branch is independent of it).

**Alternative examined and rejected.** `devil-advocate` proposed falling back to "any path-param-free GET on the resource" when `collectionPath`'s candidate has no GET, citing `computer-groups`. I checked it and the claim does not hold: the `/v1/computer-groups` list GET is **not** in the same `Resource` as `/v3/computer-groups/smart-group-membership/{id}` (the generated `computer_groups.go` carries only the `get`; `registry.go:68-69` registers `NewComputerGroupsCmd` twice, a pre-existing filename collision between two same-named resources, which is why `computer-groups list` does not exist). Worse, that fallback would resurrect dead flags elsewhere: `CertificateAuthority.yaml` has a param-free `GET /v1/pki/certificate-authority/active` and `ManagedSoftwareUpdates.yaml` has `GET .../available-updates` — neither is an RSQL-filterable collection of the `{id}` resource. The PR's stricter rule is the better default.

**Considered, below the reporting threshold** (each verified real but scored c=50 — minor or latent, listed so the author can pick them up if they wish): `generator.go:2970` — the `apply` body embeds `{{ listPathFromOps .Operations }}` (an alias for `collectionPath`, `generator.go:303-305`) while `shouldGenerateApply` now gates on `nameResolutionPath`, so a future resource carrying both a `NameLookupPath` override and an apply-eligible shape would get an `apply` that resolves against the un-overridden path; switching it to `{{ nameLookupPath . }}` closes it. Also: the `NameLookupPath` override bypasses the new GET-existence check with no test tying `mobile-devices` → `/v2/mobile-devices/detail` to a real GET op; and suppression is silent, where a one-line stderr note (as at `generator.go:920-926`) would make a future spec regression visible in `make generate` output.
</details>

<details>

No prior reviews or comments on this PR.
</details>

<details>

✅ `TestNameResolutionPath` asserts **both** `collectionPath` and `nameResolutionPath` per case, so a future change that collapses the two back together fails at the pin rather than silently resurrecting the dead flags — this is the right shape for a test guarding a deliberate divergence.

✅ The fix is wire-verified in both directions (positive path re-confirmed on a live 11.31.0 instance: `categories apply` updated rather than duplicated; negative lookups on kept resources still round-trip), and the root cause is written up as a postmortem instead of only a commit message.
</details>


<sub>🤖 Generated by the `pr-review:review` skill · reviewed head `dccb4e1`</sub>

ktn-jamf (approved):

PR #324, opened 2026-08-17
fix(build): validate JAMF_PRO_VERSION before a shell reads it

ktn-jamf:
Follow-up to the review of #321. That review found that `JAMF_PRO_VERSION` reaches a shell through two substitution passes. This PR closes the hole.

## The problem

`make` substitutes `$(JAMF_PRO_VERSION)` textually into the recipe, and the shell then evaluates the result. The sync targets wrote the value with `echo "$(JAMF_PRO_VERSION)"`, so a value that contains a double quote closed the string and the rest ran as a command.

The old presence test had the same weakness:

```make
@if [ -z "$(JAMF_PRO_VERSION)" ]; then
```

The value was already interpolated at that point, so the guard could not protect itself. Adding a format check in the same style would not have helped.

`.github/workflows/sync-specs.yaml` passed an unvalidated `workflow_dispatch` string into `make`. The `env:` indirection blocks GitHub Actions `${{ }}` injection, but it does nothing about the second pass inside `make`.

Severity is low. `workflow_dispatch` requires repository write access, and such an actor can usually already run arbitrary code in continuous integration. It matters where branch protection stops a write-access user from pushing to the default branch but still allows a dispatch, because the job holds `contents: write`, `pull-requests: write`, and `SERVER_REPO_TOKEN`.

## The fix

- Export `JAMF_PRO_VERSION` and read it in recipes as `$$JAMF_PRO_VERSION`. An environment reference is not substituted, so the shell receives one value.
- Add `check_jamf_pro_version`, which requires three dot-separated numbers. It tests the environment copy, so a hostile value cannot escape the test itself.
- Call the guard from both `sync-specs` and `sync-spec`, before any destructive step. Both targets write `specs/.spec-version`; the review named only the first.
- Validate the workflow input in its own step, before the checkouts, so a bad value fails the run early with a clear message.

## Verification

| Value | Result |
|---|---|
| `1.2.3"; touch INJECTED; echo "` | rejected by both targets, no file created |
| `11.31.0`, `9.0.1` | accepted |
| `11.31`, `11.31.0-beta`, empty | rejected |

`make test` passes.

## Note for reviewers

The pattern is strict: `^[0-9]+\.[0-9]+\.[0-9]+$`. A version with a suffix, such as `11.31.0-beta`, is rejected. Every usage example in the Makefile is a three-part version, so this matches current practice. Loosen the pattern if a suffixed version ever needs to be stamped.

Two further follow-ups from the #321 review are **not** in this PR:
- The three actions in this workflow are pinned to mutable major tags rather than commit SHAs.
- `isTokenAuthorizationError` fails open on an unparseable 403 body. That is a deliberate, tested design choice and needs a decision rather than a fix.


neilmartin83 (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Validates `JAMF_PRO_VERSION` before make/shell substitution and adds an early workflow-input check.
> Blocking: (1). See collapsed section for 3 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 with the whole-value anchor in (1) — the quote-injection case the PR targets is genuinely closed, but the same input still escapes the pattern via a newline.
- Coverage: `security-reviewer` and `devil-advocate` have never run on this PR — the rating reflects the dimensions that were searched, not the whole diff.

### Findings

🟧 **(1)** (security, c=100) — `.github/workflows/sync-specs.yaml:38` and `Makefile:68`: the pattern is anchored per **line**, not per value

`grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+$'` matches if *any* line of the input matches, so a multi-line value passes both guards. `printf '%s'` preserves the embedded newline rather than collapsing it, so the guard's own stated contract ("three dot-separated numbers") does not hold. Verified against this branch: `make sync-spec JAMF_PRO_VERSION=$'1.2.3\nbogus'` clears the version guard and proceeds to the next check. The quote case in the PR description is correctly rejected — this is a different escape from the same input.

**Failure scenario:** a dispatcher submits `jamf_pro_version` = `11.31.0\n";:>PWNED;echo"` → both guards accept it → `specs/.spec-version` gets two lines → `SPEC_PRO_VERSION := $(shell cat specs/.spec-version | tr -d '[:space:]')` (`Makefile:7`) splices the payload into `LDFLAGS`, and the next `make build` / `make install` runs it in a shell. I reproduced this chain end to end with a whitespace-free payload — the file was created. Reachable from the PR as shipped, and the poisoned `.spec-version` is committed by the `create-pull-request` step.

**Suggested fix** (both copies — the duplicated regex is the cost of the two-place guard):

```diff
+@case "$$JAMF_PRO_VERSION" in *[!0-9.]*) \
+	echo "Error: JAMF_PRO_VERSION must be three dot-separated numbers, for example 11.31.0."; \
+	echo "$(1)"; \
+	exit 1;; \
+esac
 @printf '%s' "$$JAMF_PRO_VERSION" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+$$' || { \
```

`*[!0-9.]*` rejects a newline, quote, semicolon, or space before the shape check runs; the `grep` then only has to confirm the three-part form. Mirror it in the workflow step (`case "$JAMF_PRO_VERSION" in *[!0-9.]*) … esac`). If (2) is taken and a suffix is allowed, widen the class to `*[!0-9A-Za-z.-]*` — a newline is still outside it.

**Fixed when:** `JAMF_PRO_VERSION=$'11.31.0\nbogus'` is rejected by `sync-specs`, `sync-spec`, and the workflow step, and `specs/.spec-version` can only ever receive a single line.

<details>

🟩 **(2)** (usability, c=100) — `Makefile:68`: the strict pattern rejects the build-suffixed versions the docs tell operators to read

`docs/sync-specs.md:71` instructs the operator to confirm the version via `/api/v1/jamf-pro-version`, which returns build-suffixed values on non-GA instances — and `compareProVersions` is explicitly built for them (`internal/commands/root.go:311`, asserted at `internal/commands/version_compat_test.go:154` with `11.28.0-t1234`). Either allow an optional suffix (`^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.]+)?$`) or state the accepted format in `docs/sync-specs.md` next to that step so the hard stop isn't a surprise mid-sync.

**Promote to 🟧 when:** a release flow needs a suffixed version stamped verbatim into `specs/.spec-version`.

🟩 **(3)** (security, c=100) — `Makefile:7-8`: `.spec-version` still reaches a shell by textual substitution

`SPEC_PRO_VERSION` is read from a file, stripped of whitespace, and interpolated into `LDFLAGS`, which `go build`/`go install` hand to a shell. Pre-existing, not introduced here, but it is the third substitution pass in the chain this PR set out to close and it is what gives (1) its teeth. Sanitizing at the read site is one edit: `... | tr -cd '0-9A-Za-z.-'`.

**Promote to 🟧 when:** `.spec-version` can be written by anything other than the two now-guarded targets.

🟩 **(4)** (correctness, c=75) — `Makefile:158`, `Makefile:210`: `$(JAMF_MONOLITH_SPEC)` is still substituted unquoted

Same textual-substitution pattern the version just stopped using, in the target this PR edits — `--monolith $(JAMF_MONOLITH_SPEC)` word-splits, so a path containing a space breaks the sync with a confusing generator error. Operator-supplied only (no workflow input feeds it), so this is robustness rather than a trust boundary. Quote it, or pass it through the environment as the version now is.

</details>

<details>

- [x] Design and architecture: the `define` + `$(call)` canned-recipe shape is the right factoring for a guard needed by two targets; `export` + `$$VAR` is the correct fix for the double-substitution problem, and I confirmed the mechanism works (command-line, environment, and unset cases all behave) by running an equivalent Makefile
- [x] Correctness: executed both guards against this branch — the quote payload is rejected by `sync-specs` and `sync-spec` with no file created, `11.31.0-beta`/`11.31`/empty are rejected, `11.31.0`/`9.0.1` accepted; the multi-line case is finding (1). Verified the guard precedes the first destructive step (`rm -f specs/*.yaml`, `Makefile:106`), so the PR's ordering claim holds. `$(call)` argument containing `(e.g. …)` parses correctly — parens balance, no comma
- [x] Security: traced the full input path from `workflow_dispatch` to a shell. Job-level `env:` correctly blocks `${{ }}` injection; the validate step is first and `exit 1` gates every later step, including the raw `${{ inputs.jamf_pro_version }}` uses in `create-pull-request` (`:125-127`). The residual escape is (1), its amplifier (3). `SERVER_REPO_REF` reaches a shell only as an env expansion (`:95`) — no second pass, not injectable
- [na] Performance: build-time guard, two `grep` invocations
- [x] Test coverage: no automated coverage for either guard, and the repo has no precedent for testing Makefile targets (no test references `make`, CI runs `make test` only) — so this is a gap in the repo, not in the PR. Verification was manual per the PR body; I re-ran it independently rather than take it on trust
- [x] Reliability: guard failures `exit 1` and abort the target; the `|| { … exit 1; }` group does not swallow the grep's status; workflow step fails the run before any checkout
- [x] Code quality: comments explain *why* (two substitution passes, environment reference not re-substituted) rather than what — matches the density of the surrounding Makefile. Minor asymmetry: the `JAMF_SERVER_ROOT` usage block at `Makefile:80` still prints the old two-line form while the new guard prints one; not worth a finding
- [x] Simplification: the regex now lives in two places with two different messages. Duplication is justified here (fail-early in CI + local guard) and the comment says so — but (1) must be applied to both copies or they drift
- [x] Documentation currency: `CLAUDE.md:113-114,302-308` and `docs/sync-specs.md:48,55,74` all use three-part versions, so the new constraint breaks no documented example. The newly-enforced format itself is undocumented — folded into (2) rather than raised separately
- [na] Cross-repo contracts: no wire key, enum, endpoint, or shared config name changes; `jamf/jss` is consumed by ref only, unchanged
- [x] Project rules compliance: no `.claude/rules/` directory; checked `CLAUDE.md` — the generated-code boundary and credential-input policy are untouched by this diff

</details>

<details>

Full diff (2 files, +40/−16) at head `d8cdf20`, plus `Makefile:1-20` and `:195-225` for the `LDFLAGS`/`generate` consumers, `.github/workflows/sync-specs.yaml` in full, `internal/commands/root.go:257-340`, `internal/commands/version_compat_test.go:144-165`, and `docs/sync-specs.md`.

Findings (1), (3) and the guard behaviour in the coverage notes were verified by **executing** the constructs — the real Makefile targets for the guards, an equivalent Makefile for the `LDFLAGS` chain — not by reading them.

No specialists dispatched: this session's standing instruction is not to spawn subagents unless asked, and the input suited direct verification — 56 lines of make/YAML, small enough to run end to end, which is stronger evidence than a static lane pass. Never-run lanes: `security-reviewer` and `devil-advocate` (eligible — CI workflow + build-file changes, high-risk tier — skipped this round, and named beside the rating); `test-quality-reviewer` (no test files changed and no harness exists for make targets); `performance-reviewer`, `usability-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `fidelity-reviewer` (inapplicable — no hot paths, no CLI surface, 56 lines needs no decomposition, no error-handling code, no linked ticket).

</details>

<details>

✅ `export` + `$$JAMF_PRO_VERSION` is the correct fix, not a cosmetic one — it removes the substitution pass instead of trying to escape it, and the comment at `Makefile:52-56` explains precisely why a stricter `$(…)`-based test could never have protected itself.

✅ The PR fixed `sync-spec` as well as `sync-specs`. The #321 review named only the first; both write `specs/.spec-version`, and catching the sibling is the difference between closing a hole and moving it.

</details>


<sub>🤖 Generated by the `pr-review:review` skill v1.30.0 · reviewed head `d8cdf20`</sub>

ktn-jamf:
All four findings are addressed at `7b27577`. Thank you for executing the guards instead of reading them — finding (1) was correct, and the amplification chain you traced is what made it blocking rather than cosmetic.

**🟧 (1) per-line anchor — fixed.** A character-class check now runs before the regex in both guards. `*[!0-9A-Za-z.-]*` rejects every character outside the accepted set, which is what excludes a newline; the regex then only confirms the shape. I confirmed your chain first: `tr -d '[:space:]'` at the read site *removes* the newline, so the two lines weld into one valid payload inside `LDFLAGS`. Now `JAMF_PRO_VERSION=$'11.31.0\n\";:>PWNED;echo\"'` is rejected by `sync-specs` and by `sync-spec`, and no file is created.

**🟩 (2) build suffix — taken.** The guard now accepts an optional suffix, so `11.31.0-t1785774933693` passes. Your evidence decided this: `docs/sync-specs.md` sends the operator to `/api/v1/jamf-pro-version`, and `compareProVersions` already handles that form. The strict pattern I shipped would have blocked a real sync. The accepted format is now documented next to that step.

**🟩 (3) `.spec-version` reaching a shell — taken.** The read site uses `tr -cd '0-9A-Za-z.-'`. A poisoned file now reaches `LDFLAGS` inert. This is worth keeping even though the two targets guard their writes, because the file is also editable by hand and by a rebase.

**🟩 (4) unquoted `JAMF_MONOLITH_SPEC` — taken.** Quoted at both remaining sites.

Verification, all re-run on the merged tree that now carries the dependabot bump:

| Value | Result |
|---|---|
| `11.31.0\n";:>PWNED;echo"` | rejected by both targets, no file |
| `1.2.3"; touch INJECTED; echo "` | rejected, no file |
| poisoned `.spec-version` | reaches `LDFLAGS` inert, nothing runs |
| `11.31.0`, `11.31.0-t1785774933693` | accepted |
| `11.31`, `11.31.`, empty | rejected |

`make test`, `make verify-generated` and `make build` pass, and the build still stamps `specProVersion=11.31.0`.

One note on your coverage remark: there is still no automated test for either guard, so every check above is manual. You are right that the repo has no harness for `make` targets, so I did not invent one here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

neilmartin83 (approved):
> [!TIP]
> ✅ **Merge-ready** — Validates `JAMF_PRO_VERSION` before make/shell substitution and sanitizes `.spec-version` at the read site. All four prior findings fixed, each re-verified by execution.
> See collapsed section for 2 nice-to-have suggestions.

## Rating: 5/5
- Clean — the two nice-to-have suggestions below are optional and don't hold the merge.
- Coverage: `security-reviewer` and `devil-advocate` have never run on this PR — the rating reflects the dimensions that were searched, not the whole diff.

## Prior findings status

| (1) `c=100` | `Makefile:83`, `sync-specs.yaml:42` | ✅ Fixed | Character-class check now precedes the regex in both guards. Re-ran the chain: `$'11.31.0\nbogus'` and `$'11.31.0\n";:>PWNED;echo"'` rejected by `sync-specs` **and** `sync-spec`, no file created, `rm -f specs/*.yaml` never reached |
| (2) `c=100` | `Makefile:89`, `docs/sync-specs.md:58` | ✅ Fixed | Suffix allowed and documented. `11.31.0-t1785774933693` and `11.31.0-beta.1` accepted; `11.31`, `11.31.`, `11.31.0-`, `-11.31.0`, empty still rejected |
| (3) `c=100` | `Makefile:7-13` | ✅ Fixed | `tr -cd '0-9A-Za-z.-'` at the read site. Poisoned the file with `11.31.0";touch POISONED;echo "` — `LDFLAGS` came out as `specProVersion=11.31.0touchPOISONEDecho`, nothing ran |
| (4) `c=75` | `Makefile:180`, `Makefile:232` | ✅ Fixed | Both quoted. `make sync-spec JAMF_MONOLITH_SPEC="/tmp/spaced dir/mono.json"` reaches the generator as one argument |

### Findings

<details>

🟩 **(1)** (security, c=100) — `.github/workflows/sync-specs.yaml:80`: passing the version as a make *command-line* variable adds a substitution pass the guard cannot see

`JAMF_PRO_VERSION` is already in the job's `env:` block (`:28`), so the trailing `JAMF_PRO_VERSION="$JAMF_PRO_VERSION"` is redundant — and make expands command-line variables recursively *before* any recipe runs, so a make function in the value executes ahead of `check_jamf_pro_version`. Verified on this branch: `make sync-specs JAMF_SERVER_ROOT=/tmp JAMF_PRO_VERSION='$(shell touch PWNED && echo 11.31.0)'` created the file, and the guard then passed on the expansion result. Not reachable in CI as shipped — the validate step rejects `$`, `(`, `)` first — but that makes it the sole barrier, and narrows the comment at `Makefile:65-68` ("a hostile value cannot escape the test itself") to the environment path only.

**Suggested fix:**

```diff
-          make sync-specs JAMF_SERVER_ROOT=jss JAMF_PRO_VERSION="$JAMF_PRO_VERSION"
+          # JAMF_PRO_VERSION comes from the job env, which make imports as a
+          # variable without the extra expansion pass an assignment here adds.
+          make sync-specs JAMF_SERVER_ROOT=jss
```

Verified the env-only form works: the guard receives the value verbatim and rejects a literal `$(shell …)`.

**Promote to 🟧 when:** the validate step is removed or weakened, or another caller passes an untrusted value on a `make` command line.

🟩 **(2)** (correctness, c=100) — `Makefile:13`: the `|| echo "unknown"` fallback is dead

`||` binds to the pipeline's last command, so `tr` — which succeeds on empty input — decides it. With `specs/.spec-version` absent, `SPEC_PRO_VERSION` is empty rather than `unknown` and the binary reports `specProVersion: ""`. Pre-existing (the old `tr -d '[:space:]'` behaved the same) and low-impact since the file is committed, but the line is one this PR rewrote. Fix: `$(shell tr -cd '0-9A-Za-z.-' < specs/.spec-version 2>/dev/null || echo "unknown")`.

</details>

<details>

- [x] Correctness: executed both guards against this branch across 12 values (multi-line, quote-injection, whitespace, empty, short, trailing-`-`, leading-`-`, plain, two suffix forms) — every rejection lands with the right message and `rc=2`, every accept proceeds. Confirmed the version guard still precedes the first destructive step in `sync-specs` (`rm -f specs/*.yaml`, `Makefile:128`) and precedes `generate` in `sync-spec`. Note the server-path check runs *before* the version guard in `sync-specs`, so a bad path masks a bad version — ordering is harmless, both abort
- [x] Security: re-traced the input path from `workflow_dispatch` to a shell. The two guards are now genuine mirrors (same character class, same regex), and the character class is what closes prior finding (1). Sanitization at the read site closes the amplifier — verified inert with a poisoned file. Residual gap is (1): make's own expansion of a command-line variable, blocked in CI by the validate step
- [x] Design and architecture: the two-check shape (character class, then regex) is the right factoring — the class states the alphabet, the regex states the form, and the comment at `Makefile:70-76` explains why one check alone cannot work. The workflow copy carries the same reasoning with a "keep the two in step" note
- [na] Performance: build-time guard; one `case` and one `grep` per target
- [x] Test coverage: still no automated coverage for either guard, and the repo still has no harness for `make` targets — the author says so explicitly rather than claiming otherwise. Unchanged repo-level gap, not a regression in this PR. `make test` passes (no `FAIL` lines), `make build` succeeds and stamps `specProVersion=11.31.0`
- [x] Reliability: both guards `exit 1`; the `|| { … }` group preserves the grep's status; `esac` terminates cleanly inside the `define` continuation
- [x] Code quality: comments explain the mechanism (per-line anchoring, the write-site/read-site split) rather than the syntax. Minor: the new `case` block's error message names the rule ("no whitespace or punctuation other than . and -") while the regex message names the shape — two messages for one contract, but each is accurate for the check that emits it
- [x] Simplification: the guard is duplicated in two places by necessity (fail-early in CI + local). This round kept both copies in step — same class, same regex — which is what prior finding (1) required
- [x] Documentation currency: `docs/sync-specs.md:58-65` now states the accepted format and both example forms. Placed under Option A but worded "Both sync targets", so Option B is covered. `CLAUDE.md` points at this doc and needs no edit
- [na] Cross-repo contracts: no wire key, enum, endpoint, or shared config name changed; `jamf/jss` is still consumed by ref only
- [x] Project rules compliance: no `.claude/rules/`; checked `CLAUDE.md` — the generated-code boundary and credential-input policy are untouched

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `Makefile` | 5/5 | Guards executed against injection payloads rather than read; write-site and read-site both close |
| `.github/workflows/sync-specs.yaml` | 4/5 | Validate step is correct and correctly ordered, but it is now the only thing stopping the make-expansion pass in (1) |

</details>

<details>

Incremental diff `d8cdf20..7b27577` (3 files of substance, +63/−21) plus the full PR diff against `main` (3 files, +85/−19) to confirm the merge brought nothing else in. The `go.mod`/`go.sum` movement in the incremental range is the dependabot `golang-x` bump that landed on `main` via #323 — it is not part of this PR's diff against `main` and was not reviewed here.

Every prior-finding status above was established by **executing** the construct — the real `sync-specs` and `sync-spec` targets, the real make variable database for `LDFLAGS`, a real spaced path for `JAMF_MONOLITH_SPEC` — not by reading the diff. Finding (1) was found the same way.

No specialists dispatched: this session's standing instruction is not to spawn subagents unless asked, and the input again suited direct execution. Never-run lanes: `security-reviewer` and `devil-advocate` (eligible — CI workflow + build-file changes — skipped again this round, named beside the rating); `test-quality-reviewer` (no test files changed, no harness exists for `make` targets); `performance-reviewer`, `usability-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `fidelity-reviewer` (inapplicable — no hot paths, no CLI surface, 3 files needs no decomposition, no error-handling code, no linked ticket).

</details>

<details>

✅ The reply verified the *amplification chain* before fixing it — confirming that `tr -d '[:space:]'` welds a two-line value into one payload inside `LDFLAGS` — and then fixed both ends: the write-site guard and the read-site sanitizer. Fixing only the guard would have left the file poisonable by hand or by a rebase, which the comment at `Makefile:7-13` now says out loud.

✅ Prior finding (2) was accepted on evidence rather than defended. The strict pattern shipped in the first round would have hard-stopped a real sync off `/api/v1/jamf-pro-version`, and the accepted format is now documented next to the step that produces it.

</details>


<sub>🤖 Generated by the `pr-review:review` skill v1.30.0 · reviewed head `7b27577`</sub>

PR #332, opened 2026-08-19
fix(pro): diff read one directory level, hiding every nested resource

### LINKED ISSUE #331: Diff command doesn't work with configuration profiles
### jamf-cli version

1.26.0

### Product namespace

pro (Jamf Pro / Platform API)

### Auth method

oauth2 (client credentials)

### OS / platform

macOS 26.6

### Command run

```shell
jamf-cli pro diff --source <source-dir> --target <target-dir>
```

### Expected behavior

Changes to configuration profiles are detected.

### Actual behavior

```shell
Only changes to other resources are identified.
```

### Anything else?

_No response_

neilmartin83:
Fixes #331.

## The report

`pro diff --source <dir> --target <dir>` found changes in other resources but never in configuration profiles.

## Root cause

`loadSnapshotFromDirectory` walked exactly one level of subdirectories and `readObjectsFromSubdir` skips directory entries. But `backup` lays out each resource at its `SubDir`, and **thirteen** curated resources nest two levels:

```
profiles/macos            profiles/ios
prestages/computers       prestages/mobile
extension-attributes/{computer,mobile,user}
smart-groups/{computers,mobile}
static-groups/{computers,mobile}
accounts/{users,groups}
```

For `profiles/` the loader found zero files. A bucket is only added to the snapshot when it holds at least one object, so the key never appeared on **either** side and `compareSnapshots` had nothing to compare. Exit 0, no warning, output indistinguishable from "these two instances agree" — the worst failure shape a diff can have. Profiles were just the resource the reporter happened to change; prestages, extension attributes, smart groups, static groups and accounts were equally invisible.

Repro on `main` (synthetic tree, one changed `profiles/macos` profile, one changed `smart-groups/computers` group, one changed flat policy):

```
$ bin/jamf-cli pro diff --source ./a --target ./b
[ { "resource": "policies", "change": "modified", ... } ]   ← only the flat resource
exit=0
```

Same tree after this change reports all three, with real object names.

## Two adjacent mismatches from the same repro

- **The loaders bucketed under different keys.** Live mode buckets by `FilterName` (`profiles`), the directory loader by top-level directory name. Those strings coincide for flat resources — which is why the bug looked profile-specific. For nested ones, `diff --source ./backup --target production` had the key populated on one side only, so every profile came back `added` or `removed`.
- **Classic objects were keyed by filename stem.** The loader read `obj["name"]`, but a Classic detail nests its name under `general` and a prestage calls it `displayName`. The stem is `SlugifyName` output, possibly carrying a `DeduplicateSlug` suffix, so it never equals the name live mode takes from the list item — every Classic object in a directory-vs-instance diff was reported removed **and** added.

## The fix

- `BackupSubDirs()` (`pro_resources.go`) maps each `SubDir` → `FilterName`, so both loaders derive keys from one table instead of two conventions.
- `loadSnapshotFromDirectory` **reads that table instead of walking the tree**: one `os.ReadDir` of the backup root picks up the directories no curated `SubDir` claims — keyed by their own name, which carries `blueprints` / `compliance-benchmarks`, hand-assembled trees, and object files left directly in a nested parent (`profiles/*.yaml` beside `profiles/macos/`, where the directory name is already the `FilterName`) — then every `BackupResources` entry in slice order, read at `filepath.Join(dir, r.SubDir)`. A missing directory is skipped; an unreadable **root** is an error, not a warning.
- That is what makes the two loaders agree on more than the key. Both merge shared buckets with `maps.Copy` in `BackupResources` order, so the entry listed *last* wins a name collision on both sides — a macOS and an iOS profile both called `Corporate Wi-Fi` resolve to the same object on disk as they do live. A walk visits lexically and inverts that for `profiles` and `accounts`. Reading by path also bounds the read to the paths the table names (which is what keeps `packages/files`, holding downloaded package binaries, out of the snapshot) and drops the nested-parent bookkeeping, the `fs.SkipDir` handling and the depth assumptions with it.
- `backupObjectName` replaces the inline `obj["name"]` read and takes the resource's `BackupEndpoint.NameField` — checked first for the same reason `extractName` checks it first: it is where the resources that do not call their name `name` keep it (prestages `displayName`, mobile-device smart/static groups `groupName`). Then `name`, then the two shapes no registry field names (a Classic detail nesting its name under `general`, and `displayName` for a directory read without a `NameField`), then the filename stem.
- Following a **symlinked** resource directory is a capability this loader *adds*, not one it preserves: `os.ReadDir` reports a symlink-to-directory with `IsDir() == false`, so the old loader skipped it. Reading the curated resources by path follows them for free, and `entryIsDir` keeps the root-level scan consistent with that. Worth knowing when handed an untrusted tree — `diff` will read and print field values from outside it.

## Tests

Twelve new tests. The ones written to hold the line rather than the symptom:

- `TestLoadSnapshotFromDirectory_EveryCuratedSubDirIsRead` writes a file into **every** `BackupResource.SubDir` and asserts each is readable under its `FilterName` — a test over the table, so a nested resource added later cannot reintroduce the class.
- `TestLoadSnapshotFromDirectory_EveryCuratedNameFieldIsHonoured` does the same for the key: an object carrying *only* its resource's declared `NameField` must be keyed by that value, so a resource added later whose name lives somewhere new fails rather than silently falling back to the stem.
- `TestBackupObjectName_MatchesLiveNameExtraction` asserts the directory key equals `extractName` of the equivalent live list item; nothing else in the codebase forced the two loaders to agree.
- `TestLoadSnapshotFromDirectory_SharedBucketCollisionMatchesLiveOrder` pins the collision winner for `profiles` and `accounts` by reading it out of `BackupResources`, so reordering the table moves the expectation with it.
- `TestLoadSnapshotFromDirectory_UnreadableRootIsAnError` + `TestRunDiff_UnreadableSourceDirectoryErrors` keep an unreadable source out of the "No differences found" path, at the loader and at the command boundary.

Every new test was verified to fail against the behaviour it pins (the one-level walk, a `backupObjectName` without the `NameField` branch, a lexically-ordered curated pass, a root scan that skips nested parents, and a root read that only warns) and pass after. `make test`, `make lint` (0 issues) and `make verify-generated` all clean; end-to-end re-checked with the built binary — three nested/flat resources reported, and an unreadable source now exits 1 instead of printing "No differences found".

Postmortem: `docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md`.

Note: no change to `backup`'s on-disk layout — existing backup directories diff correctly as-is.

### DIFF
--- a/docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md
+++ b/docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md
@@ -1,0 +1,112 @@
+---
+title: "pro diff read one directory level, so every nested resource — config profiles included — silently reported no change"
+date: 2026-08-19
+category: logic-errors
+module: internal/commands
+problem_type: bug
+severity: high
+applies_when:
+  - "Adding a BackupResource whose SubDir has more than one segment"
+  - "Changing how `diff` loads a backup directory, or how it keys objects"
+  - "A user reports `diff` finding changes for some resources but not others"
+  - "Comparing a backup directory against a live instance"
+tags:
+  - diff
+  - backup
+  - snapshot
+  - directory-layout
+  - silent-failure
+  - config-profiles
+issue: 331
+---
+
+# `diff` read one directory level, so nested resources reported no change
+
+## Context
+
+`pro diff` builds a `resourceSnapshot` — `resource → object name → fields` — from
+each side, then compares them. A side is either a live instance
+(`loadSnapshotFromProfile`) or a backup directory (`loadSnapshotFromDirectory`).
+`backup` lays a directory out from each curated resource's `SubDir`
+(`internal/commands/pro_resources.go`), and thirteen of those nest two levels:
+
+```
+profiles/macos            profiles/ios
+prestages/computers       prestages/mobile
+extension-attributes/{computer,mobile,user}
+smart-groups/{computers,mobile}
+static-groups/{computers,mobile}
+accounts/{users,groups}
+```
+
+## The bug
+
+The directory loader walked exactly one level (`os.ReadDir`, "each subdir is a
+resource type") and `readObjectsFromSubdir` skips directory entries. So for
+`profiles/` it found zero files, and because a bucket is only added to the
+snapshot when it holds at least one object, the key never appeared on *either*
+side. `compareSnapshots` had nothing to compare, so it reported nothing.
+
+That is the worst possible failure shape for a diff: not an error, not a warning,
+exit 0, and output that is indistinguishable from "these two instances agree".
+The reporter saw it as "diff doesn't work with configuration profiles"; profiles
+were just the resource they happened to change. Prestages, extension attributes,
+smart groups, static groups and accounts were equally invisible.
+
+Two adjacent mismatches came out of the same repro:
+
+- **Directory and live mode bucketed under different keys.** Live mode buckets by
+  `FilterName` (`profiles`), the directory loader by top-level directory name. For
+  flat resources those strings coincide, which is why the bug looked
+  profile-specific. For nested ones, a directory-vs-instance diff had the key
+  populated on one side only — so every profile reported as `added` or `removed`.
+- **Objects were keyed by filename stem for Classic resources.** The loader read
+  `obj["name"]`, but a Classic detail nests its name under `general`, and a
+  prestage calls it `displayName`. Falling through to the stem meant the key was
+  `SlugifyName` output — possibly with a `DeduplicateSlug` suffix — while live
+  mode used the real name from the list item. So `diff --source ./backup --target
+  production` reported every Classic object as removed *and* added.
+
+## Fix
+
+`BackupSubDirs()` (`pro_resources.go`) maps each `SubDir` to its `FilterName`, so
+the two loaders derive their keys from one table instead of two conventions.
+`loadSnapshotFromDirectory` now walks the tree with `filepath.WalkDir` and, per
+directory:
+
+- a path in `BackupSubDirs()` is read and then `fs.SkipDir` — a resource directory
+  owns its *files*, not its subdirectories, which is what keeps
+  `packages/files` (downloaded package binaries) out of the snapshot;
+- a parent of a nested resource path (`profiles`) is descended through, never read
+  as a resource itself;
+- an unknown *top-level* directory keeps the old behaviour, keyed by its own name
+  — that is what carries the SDK-backed resources written straight to the backup
+  root (`blueprints`, `compliance-benchmarks`) and any hand-assembled tree;
+- an unknown *nested* directory is skipped.
+
+Buckets merge with `maps.Copy` exactly as live mode does, so macOS + iOS profiles
+(and account users + groups) land in one bucket on both sides.
+
+`backupObjectName` replaces the inline `obj["name"]` read: `name`, then
+`displayName`, then `general.name`, then the stem as a last resort.
+
+## Guardrails
+
+`TestLoadSnapshotFromDirectory_EveryCuratedSubDirIsRead` writes one file into
+*every* `BackupResource.SubDir` and asserts each is readable under its
+`FilterName`. That is deliberately a test over the table rather than over the
+thirteen paths, so a nested resource added later cannot reintroduce the class.
+
+`TestBackupObjectName_MatchesLiveNameExtraction` asserts the directory key equals
+`extractName` of the corresponding list item — the two loaders are only
+comparable if they agree on this, and nothing else in the codebase forces it.
+
+## Watch for
+
+- A resource whose live list name differs from anything in its detail body would
+  break key parity again; `backupObjectName` can only see the file on disk.
+- `NameField` in the generated backup registry accepts dotted paths
+  (`computers-inventory` declares `general.name`) but `extractName` does not
+  traverse them. No curated backup resource relies on it today.
+- `diff` still says nothing when a resource directory exists but is empty on both
+  sides, which is correct but reads the same as "no such resource".
--- a/docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md
+++ b/docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md
@@ -71,24 +71,46 @@ Two adjacent mismatches came out of the same repro:
 
 `BackupSubDirs()` (`pro_resources.go`) maps each `SubDir` to its `FilterName`, so
 the two loaders derive their keys from one table instead of two conventions.
-`loadSnapshotFromDirectory` now walks the tree with `filepath.WalkDir` and, per
-directory:
+`loadSnapshotFromDirectory` then **reads that table instead of walking the tree**:
 
-- a path in `BackupSubDirs()` is read and then `fs.SkipDir` — a resource directory
-  owns its *files*, not its subdirectories, which is what keeps
-  `packages/files` (downloaded package binaries) out of the snapshot;
-- a parent of a nested resource path (`profiles`) is descended through, never read
-  as a resource itself;
-- an unknown *top-level* directory keeps the old behaviour, keyed by its own name
-  — that is what carries the SDK-backed resources written straight to the backup
-  root (`blueprints`, `compliance-benchmarks`) and any hand-assembled tree;
-- an unknown *nested* directory is skipped.
+- one `os.ReadDir` of the backup root picks up the directories no curated
+  `SubDir` claims, keyed by their own name — the SDK-backed resources written
+  straight to the root (`blueprints`, `compliance-benchmarks`), any
+  hand-assembled tree, and object files left directly in a parent of a nested
+  resource (`profiles/*.yaml` beside `profiles/macos/`), where the directory name
+  is already the `FilterName`. A failure to read the root is returned as an
+  error, not warned about: an empty snapshot there is not a partial result, it is
+  "the source was never read", and reporting that as no differences is the same
+  silent success this fix exists to remove;
+- then every entry of `BackupResources`, in slice order, read at
+  `filepath.Join(dir, r.SubDir)`. A missing directory is simply skipped.
 
-Buckets merge with `maps.Copy` exactly as live mode does, so macOS + iOS profiles
-(and account users + groups) land in one bucket on both sides.
+Reading the table rather than walking it is what makes the two loaders agree on
+more than the key. Both merge shared buckets with `maps.Copy` in
+`BackupResources` order, so the entry listed *last* wins a name collision on both
+sides — a macOS and an iOS profile both called `Corporate Wi-Fi` resolve to the
+same object on disk as they do live. A walk visits lexically (`profiles/ios`
+before `profiles/macos`, `accounts/groups` before `accounts/users`) and inverts
+that for exactly those two `FilterName`s, which turns a hidden object into a diff
+reporting every field of the survivor as modified. Reading by path also bounds
+the read to the directories the table names, which is what keeps `packages/files`
+(downloaded package binaries) out of the snapshot, and it drops the nested-parent
+bookkeeping, the `fs.SkipDir` handling and the depth assumptions with it.
 
-`backupObjectName` replaces the inline `obj["name"]` read: `name`, then
-`displayName`, then `general.name`, then the stem as a last resort.
+One behaviour is genuinely new rather than preserved: a resource directory may be
+a **symlink**. The old loader skipped symlinks (`os.ReadDir` reports a
+symlink-to-directory with `IsDir() == false`); reading the curated resources by
+path follows them for free, and `entryIsDir` keeps the root-level scan consistent
+with that. Worth knowing when handed an untrusted tree — `diff` will read and
+print field values from outside it.
+
+`backupObjectName` replaces the inline `obj["name"]` read and takes the
+resource's `BackupEndpoint.NameField`, checked first for the same reason
+`extractName` checks it first: it is where the resources that do not call their
+name `name` keep it — prestages use `displayName`, mobile-device smart and static
+groups use `groupName`. Then `name`, then the two shapes no registry field names
+(a Classic detail nesting its name under `general`, and `displayName` for a
+directory read without a `NameField`), then the filename stem as a last resort.
 
 ## Guardrails
 
@@ -97,16 +119,31 @@ Buckets merge with `maps.Copy` exactly as live mode does, so macOS + iOS profile
 `FilterName`. That is deliberately a test over the table rather than over the
 thirteen paths, so a nested resource added later cannot reintroduce the class.
 
+`TestLoadSnapshotFromDirectory_EveryCuratedNameFieldIsHonoured` is the same idea
+for the key: for every curated resource it writes an object carrying *only* its
+declared `NameField`, so a resource added later whose name lives somewhere new
+fails rather than silently falling back to the stem.
+
 `TestBackupObjectName_MatchesLiveNameExtraction` asserts the directory key equals
 `extractName` of the corresponding list item — the two loaders are only
 comparable if they agree on this, and nothing else in the codebase forces it.
+
+`TestLoadSnapshotFromDirectory_SharedBucketCollisionMatchesLiveOrder` pins the
+collision winner for `profiles` and `accounts` by reading it out of
+`BackupResources`, so reordering the table moves the expectation with it.
+
+`TestLoadSnapshotFromDirectory_UnreadableRootIsAnError` and
+`TestRunDiff_UnreadableSourceDirectoryErrors` keep an unreadable source out of
+the "No differences found" path, at the loader and at the command boundary.
 
 ## Watch for
 
 - A resource whose live list name differs from anything in its detail body would
   break key parity again; `backupObjectName` can only see the file on disk.
 - `NameField` in the generated backup registry accepts dotted paths
-  (`computers-inventory` declares `general.name`) but `extractName` does not
-  traverse them. No curated backup resource relies on it today.
+  (`computers-inventory` declares `general.name`) but neither `extractName` nor
+  `backupObjectName` traverses them — the latter happens to land on the right
+  value through its `general.name` branch, not because the dotted field was
+  honoured. No curated backup resource relies on it today.
 - `diff` still says nothing when a resource directory exists but is empty on both
   sides, which is correct but reads the same as "no such resource".
--- a/internal/commands/pro_diff.go
+++ b/internal/commands/pro_diff.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"maps"
 	"os"
 	"path/filepath"
@@ -134,31 +135,99 @@ func loadSnapshotFromDirectory(dir string, nameFilter []string) (resourceSnapsho
 
 	snapshot := make(resourceSnapshot)
 
-	// Walk one level of subdirectories; each subdir is a resource type.
-	entries, err := os.ReadDir(dir)
-	if err != nil {
-		return nil, fmt.Errorf("reading directory %q: %w", dir, err)
+	// knownSubDirs maps a resource's on-disk path to the FilterName it is
+	// bucketed under; nestedParents holds the directories that merely contain
+	// those paths (e.g. "profiles" contains profiles/macos and profiles/ios) and
+	// are walked through rather than read as a resource.
+	knownSubDirs := BackupSubDirs()
+	nestedParents := make(map[string]bool)
+	for sub := range knownSubDirs {
+		if i := strings.Index(sub, "/"); i > 0 {
+			nestedParents[sub[:i]] = true
+		}
 	}
 
-	for _, entry := range entries {
-		if !entry.IsDir() {
-			continue // skip files at the root (e.g., _failures.yaml)
-		}
-
-		resourceName := entry.Name()
-		if len(allowedResources) > 0 && !allowedResources[resourceName] {
-			continue
-		}
-
-		subDir := filepath.Join(dir, resourceName)
-		objects, err := readObjectsFromSubdir(subDir)
+	// readInto reads one resource directory and merges it into the snapshot
+	// under resourceName. Several curated resources share a FilterName (macOS +
+	// iOS profiles, account users + groups) and so merge into one bucket — the
+	// same shape live mode produces, which is what makes the two comparable.
+	readInto := func(resourceName, path string) {
+		objects, err := readObjectsFromSubdir(path)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "WARNING: reading %s: %v\n", subDir, err)
-			continue
+			fmt.Fprintf(os.Stderr, "WARNING: reading %s: %v\n", path, err)
+			return
 		}
-		if len(objects) > 0 {
+		if len(objects) == 0 {
+			return
+		}
+		if existing, ok := snapshot[resourceName]; ok {
+			maps.Copy(existing, objects)
+		} else {
 			snapshot[resourceName] = objects
 		}
+	}
+
+	walkErr := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "WARNING: walking %s: %v\n", path, err)
+			return nil
+		}
+
+		// A symlinked resource directory is read in place: WalkDir reports it as
+		// a non-directory entry and never descends into it.
+		symlinkedDir := false
+		if !d.IsDir() && d.Type()&fs.ModeSymlink != 0 {
+			if info, statErr := os.Stat(path); statErr == nil && info.IsDir() {
+				symlinkedDir = true
+			}
+		}
+		if !d.IsDir() && !symlinkedDir {
+			return nil // object files are read by readObjectsFromSubdir
+		}
+
+		rel, relErr := filepath.Rel(dir, path)
+		if relErr != nil {
+			return nil
+		}
+		rel = filepath.ToSlash(rel)
+		if rel == "." {
+			return nil
+		}
+
+		// skip stops the walk from descending, except for a symlinked entry
+		// where fs.SkipDir would skip the *containing* directory instead.
+		skip := fs.SkipDir
+		if symlinkedDir {
+			skip = nil
+		}
+
+		resourceName, known := knownSubDirs[rel]
+		switch {
+		case known:
+			// Read it, then stop: a resource directory owns its files, not its
+			// subdirectories — packages/files holds downloaded package
+			// binaries, which are not config objects.
+		case nestedParents[rel] && !symlinkedDir:
+			return nil // container — descend to the resource directories inside
+		case !strings.Contains(rel, "/"):
+			// A top-level directory outside the curated set, keyed by its own
+			// name. Covers the SDK-backed resources written straight to the
+			// backup root (blueprints, compliance-benchmarks) and any
+			// hand-assembled tree.
+			resourceName = rel
+		default:
+			return skip // unknown nested directory
+		}
+
+		if len(allowedResources) > 0 && !allowedResources[resourceName] {
+			return skip
+		}
+
+		readInto(resourceName, path)
+		return skip
+	})
+	if walkErr != nil {
+		return nil, fmt.Errorf("reading directory %q: %w", dir, walkErr)
 	}
 
 	return snapshot, nil
@@ -209,19 +278,34 @@ func readObjectsFromSubdir(subDir string) (map[string]map[string]any, error) {
 		// Strip _meta block added by backup — not part of the config.
 		delete(obj, "_meta")
 
-		// Use the "name" field as the object key; fall back to the filename stem.
-		objName := ""
-		if n, ok := obj["name"].(string); ok && n != "" {
-			objName = n
-		} else {
-			stem := strings.TrimSuffix(strings.TrimSuffix(name, ".yaml"), ".json")
-			objName = stem
-		}
-
-		objects[objName] = obj
+		stem := strings.TrimSuffix(strings.TrimSuffix(name, ".yaml"), ".json")
+		objects[backupObjectName(obj, stem)] = obj
 	}
 
 	return objects, nil
+}
+
+// backupObjectName picks the snapshot key for one object read off disk. It has
+// to agree with live mode, which keys on the list item's name — otherwise a
+// directory diffed against an instance reports every object as removed *and*
+// added. Modern resources carry "name", prestages carry "displayName", and
+// Classic details nest theirs under "general". Only when none of those is
+// present does the filename stem stand in, and a stem is a poor key: it is
+// SlugifyName output, possibly carrying a DeduplicateSlug suffix, so it never
+// equals the name the live side uses.
+func backupObjectName(obj map[string]any, fileStem string) string {
+	if n, ok := obj["name"].(string); ok && n != "" {
+		return n
+	}
+	if n, ok := obj["displayName"].(string); ok && n != "" {
+		return n
+	}
+	if general, ok := obj["general"].(map[string]any); ok {
+		if n, ok := general["name"].(string); ok && n != "" {
+			return n
+		}
+	}
+	return fileStem
 }
 
 // normaliseViaJSON round-trips a map through JSON to coerce yaml.v3 types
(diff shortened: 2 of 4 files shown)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Rewrites `diff`'s backup-directory loader to walk nested resource subdirectories and to key objects the way live mode does.
> Blocking: (1), (2), (3). See collapsed section for 3 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 with the `groupName` branch in (1), the root-unreadable error path in (2), and one merge order shared by both loaders in (3).
- Coverage: no specialist lane has ever run on this PR — `security-reviewer`, `usability-reviewer`, `fidelity-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate` and `test-quality-reviewer` were covered inline by the orchestrator because subagent dispatch was unavailable in this environment, which is not the same as the lanes having run.

### Findings

🟧 **(1)** (correctness, c=100) — `internal/commands/pro_diff.go:296`: `backupObjectName` misses `groupName`, so mobile-device groups still mis-key

The new key resolver covers `name`, `displayName` and `general.name`. Two curated resources declare neither: `mobile-device-groups-smart-groups` and `mobile-device-groups-static-groups` carry `NameField: "groupName", IDField: "groupId"` (`internal/commands/pro/generated/backup_registry.go:146-147`), and their detail schemas (`SmartGroupDetailV2` / `StaticGroup`, `specs/MobileDeviceGroup.yaml:559,581`) hold `groupName` with no `name`, no `displayName` and no `general`. `groupName` also survives `StripServerFields`, so it is on disk. The disk side therefore still falls through to the slug stem while live mode keys on `groupName` — the exact mismatch the PR set out to remove, left in place for `smart-groups/mobile` and `static-groups/mobile`.

**Failure scenario:** `diff --source ./backup --target production` where a mobile-device smart group is named `Static iPads` → disk key `static-ipads`, live key `Static iPads` → the group is reported `removed` **and** `added` on every run, reachable from the PR as shipped.

**Suggested fix:**
```diff
 	if n, ok := obj["displayName"].(string); ok && n != "" {
 		return n
 	}
+	// Mobile-device smart/static groups name their field groupName; the
+	// generated registry declares it as their NameField.
+	if n, ok := obj["groupName"].(string); ok && n != "" {
+		return n
+	}
 	if general, ok := obj["general"].(map[string]any); ok {
```

**Fixed when:** `backupObjectName` returns `groupName`, and `TestBackupObjectName_MatchesLiveNameExtraction` (`pro_diff_test.go:841`) gains a case `detail: {"groupName": "Static iPads"}`, `listItem: {"groupId": "6", "groupName": "Static iPads"}`, `nameFld: "groupName"`, `stem: "static-ipads"` that fails without the branch.

🟧 **(2)** (correctness, c=100) — `internal/commands/pro_diff.go:170-174`: an unreadable source directory now reports "No differences found" and exits 0

Pre-fix, `os.ReadDir(dir)` failing returned an error out of `loadSnapshotFromDirectory` and `runDiff` exited non-zero. The walk callback now warns and returns `nil` for **every** error including the root's own `ReadDir` failure — `filepath.WalkDir` calls the callback a second time with that error and then returns `nil` (verified against Go's implementation). Consequently `walkErr` can never be non-nil (the callback only ever returns `nil` or `fs.SkipDir`) and the wrapper at `pro_diff.go:229-231` is unreachable. The PR exists to remove a silent exit-0 "the two sides agree"; this reintroduces one for a different input.

**Failure scenario:** `diff --source ./backup --target production` where `./backup` is mode `0111` (stat succeeds, `ReadDir` fails with permission denied) → one stderr warning, empty snapshot, `No differences found.`, exit 0 → a CI gate reading exit status treats the two sides as identical.

**Suggested fix:**
```diff
 	walkErr := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
+			if path == dir {
+				// The root itself is unreadable — an empty snapshot here is not
+				// a partial result, it is a failure to read the source at all.
+				return err
+			}
 			fmt.Fprintf(os.Stderr, "WARNING: walking %s: %v\n", path, err)
 			return nil
 		}
```

**Fixed when:** an unreadable source directory makes `runDiff` return an error (non-zero exit) instead of printing "No differences found", the `walkErr` wrapper is reachable, and a test asserts it (`chmod 0o111`, skipped when running as root or on Windows).

🟧 **(3)** (correctness, c=75) — `internal/commands/pro_diff.go:154-168`: the two loaders merge shared buckets in opposite orders, so a name collision resolves differently on each side

`readInto` merges with `maps.Copy(existing, objects)` — last write wins — and `WalkDir` visits lexically: `profiles/ios` before `profiles/macos`, `accounts/groups` before `accounts/users`. Live mode merges in `BackupResources` slice order: macOS before iOS, users before groups (`pro_resources.go:48-49,97-98`). For those two FilterNames the surviving object is inverted between disk and live. (`extension-attributes`, `smart-groups`, `static-groups` and `prestages` happen to agree.) The same-name collision inside one bucket predates this PR on the live side; the cross-loader disagreement is new, and it turns a hidden object into a false positive.

**Failure scenario:** a macOS profile and an iOS profile both named `Corporate Wi-Fi` → the disk bucket keeps the macOS one, the live bucket keeps the iOS one → `diff --source ./backup --target production` reports `Corporate Wi-Fi` as `modified` in every field, and the other profile never appears in the comparison at all. Reachable from the PR as shipped.

**Suggested fix:** give both loaders one iteration order by reading the table instead of walking it — iterate `BackupResources` in slice order and read `filepath.Join(dir, filepath.FromSlash(r.SubDir))` per entry (a missing directory is simply skipped), then make one `os.ReadDir(dir)` pass for top-level directories that no `SubDir` claims. That is the same table `BackupSubDirs()` already exposes, and it deletes the four-branch switch, the `nestedParents` map, the `fs.SkipDir`/symlink special cases and findings (5) and (6) with it — `readObjectsFromSubdir` follows a symlinked resource directory on its own. If you would rather keep the walk, sort the shared-bucket reads into `BackupResources` order before merging.

**Fixed when:** a same-named `profiles/macos` + `profiles/ios` pair (and an `accounts/users` + `accounts/groups` pair) resolves to the same object on the disk side as on the live side, covered by a test that pins the winner.

<details>

🟩 **(4)** (correctness/docs, c=100) — `internal/commands/pro_diff.go:176-183`, `pro_diff_test.go:903-905`: the symlink branch adds a capability, it does not preserve one.

`os.ReadDir` returns the entry from the directory itself, so a symlink to a directory reports `IsDir() == false` (verified: `type=L---------`) and the old loader's `if !entry.IsDir() { continue }` skipped it. `TestLoadSnapshotFromDirectory_SymlinkedResourceDir` therefore fails against the pre-fix loader — it asserts new behavior while its comment, the PR body and the postmortem all describe it as "a capability `os.ReadDir` had for free". Worth being deliberate about, because following symlinks means `diff` now reads and prints field values from outside the backup tree when handed an untrusted tree.

**Fixed when:** the comment, PR body and postmortem say the loader now follows symlinked resource directories — or the branch is dropped and the test asserts they are skipped.

🟩 **(5)** (correctness, c=100) — `internal/commands/pro_diff.go:210`: object files sitting directly in a nested parent are silently dropped.

`case nestedParents[rel]` returns without reading, so `profiles/*.yaml`, `accounts/*.yaml` etc. now contribute nothing where the old loader read them under that key. The PR body credits the unknown-top-level fallback with carrying "any hand-assembled tree", which holds for every name except the six parents. A symlinked parent lands in the same hole: it falls to the top-level branch, is read in place (no files there), and `WalkDir` will not descend it. Cheapest fix: `readInto(rel, path)` before `return nil` — a no-op when the parent holds no object files.

🟩 **(6)** (correctness, c=100) — `internal/commands/pro_diff.go:143-148`: `nestedParents` only registers the first path segment.

`strings.Index(sub, "/")` records `a` for `a/b/c` but not `a/b`, so a three-segment `SubDir` added later would be skipped by the `default` branch; likewise a `SubDir` that is both a resource and the parent of another is read-and-skipped, hiding the child. `TestLoadSnapshotFromDirectory_EveryCuratedSubDirIsRead` catches both at test time, which is why this is not blocking. Register every ancestor prefix instead of the first one.

</details>

<details>

- [x] Design and architecture: the single `SubDir → FilterName` table is the right seam; the four-branch walk built on it is the weak part — see (3), (5), (6)
- [x] Correctness: traced every walk branch (known / nested parent / unknown top-level / unknown nested / symlink / non-dir), the filter interaction, `filepath.ToSlash` separator handling, depth bounding, and both loaders' bucket-merge order; findings (1), (2), (3), (5), (6). Empirically verified `os.ReadDir` symlink semantics and `WalkDir`'s root-error path in a scratch program
- [x] Security: no credentials, network or dependency surface; the one behavior change is symlink following on a local, user-owned path — traversal risk only for an untrusted backup tree, recorded as (4). `WalkDir` does not follow symlinks, so no loop or unbounded descent
- [x] Performance: reads stay bounded to depth 2 (`fs.SkipDir` on every resource directory), and `packages/files` is never enumerated — confirmed by `TestLoadSnapshotFromDirectory_IgnoresPackageFiles`. `BackupSubDirs()` rebuilds a 28-entry map twice per run
- [x] Test coverage: 7 new tests + a table-driven guard over `BackupResources`; 4 genuinely fail against the one-level loader. Gaps: no case for the `groupName` shape (1), the unreadable root (2), the collision order (3), or files directly in a parent (5); the symlink test asserts new behavior (4). `go test ./internal/commands/ -run 'TestLoadSnapshot|TestBackupObjectName|TestRunDiff'` passes at head
- [x] Reliability: `readInto` keeps per-directory read/parse failures as warnings, matching the old loader; the root-level error path is (2). `diff` still exits 0 when differences are found — pre-existing, unchanged by this PR, and no consumer contract in the diff depends on it
- [x] Code quality: comments are unusually load-bearing here and mostly accurate; the two that are not are (4) and (5). `BackupSubDirs()` is exported for a same-package caller, consistent with `BackupResources`/`BackupFilterNames` beside it
- [x] Simplification: the walk carries four branches plus two special cases where a read over `BackupResources` needs none — folded into (3) rather than raised separately
- [na] Frontend concerns: no UI in this repo
- [x] Documentation currency: `docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md` matches the repo's postmortem convention (frontmatter `module`/`tags`/`issue`), and no user-facing doc describes the directory walk — `grep` over `docs/**` and `README.md` for `pro diff` returns only the new postmortem. Two statements in it are corrected by (4) and (5)
- [na] Cross-repo contracts: no wire key, enum, endpoint or shared config changes; `BackupSubDirs()` is consumed only inside `internal/commands`
- [x] Project rules compliance: checked `CLAUDE.md` (no `.claude/rules/` exists) — no generated file edited, the curated `backup`/`diff` resource set is changed in `internal/commands/pro_resources.go` as the "Where to Make Changes" table requires, `docs/GLOSSARY.md` terminology respected, and `docs/solutions/` frontmatter matches the existing files

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `internal/commands/pro_diff.go` | 3/5 | Fixes the reported bug, but leaves one key shape unhandled and turns one hard error into a silent success — see (1), (2), (3) |
| `internal/commands/pro_resources.go` | 5/5 | Small pure derivation over the existing table; no behavior of its own, no caller outside the package |
| `internal/commands/pro_diff_test.go` | 4/5 | Verified 4 of the new tests fail against the pre-fix one-level loader, so the regression coverage is real rather than asserted; the table-driven `EveryCuratedSubDirIsRead` is the strongest test here |

</details>

<details>

Full diff (4 files, +519/−27) at head `5c9b750` against `origin/main` (`027396a`). Read beyond the diff: `internal/commands/pro_diff.go` in full (both loaders, `compareSnapshots`, `readObjectsFromSubdir`), `internal/commands/pro_resources.go` in full, `internal/commands/pro_backup.go` (write path, `extractName`, `unwrapClassicDetail`, slug/dedup), `internal/commands/pro_helpers.go` (`StripServerFields`), `internal/commands/pro/generated/backup_registry.go` (`NameField`/`IDField` for all 28 curated keys), `specs/MobileDeviceGroup.yaml` (group detail schemas), `CLAUDE.md`, `docs/solutions/`, plus `git log -S` on `SubDir` to confirm `backup` has never written a flat `profiles/` layout. Two claims were checked by running scratch Go programs rather than by reading docs: `os.ReadDir`'s symlink `IsDir()` result, and `filepath.WalkDir`'s callback/return behavior on an unreadable root.

Risk tier: **high** — 519 changed lines crosses the >500 threshold, and independently `loadSnapshotFromDirectory` is a guard deciding what the rest of the path sees, which the dispatch reference tiers high regardless of line count. In substance most of the diff is tests (278) and a postmortem (112) over ~160 lines of source.

Never-run lanes, all eight: `security-reviewer`, `usability-reviewer`, `fidelity-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate`, `test-quality-reviewer` — subagent dispatch is unavailable in this environment, so each lane's checks were performed inline by the orchestrator and the lane record stays empty so a later round retries them. Inline results worth recording: the fidelity pass confirms issue #331's stated expectation ("changes to configuration profiles are detected") is met for both directory-vs-directory and directory-vs-instance, with (1) leaving two sibling resources short of the PR's own stated goal; the scope pass recommends keeping this as one PR (single defect, its regression tests, and the repo-convention postmortem); no `test-quality` mutation spot-checks were run, so test effectiveness rests on reading plus the pre-fix failure claims in the PR body, which I re-derived by hand rather than by executing the old loader.

</details>

<details>

✅ The guard test is written over the table rather than over the thirteen paths that were broken (`TestLoadSnapshotFromDirectory_EveryCuratedSubDirIsRead`), so a nested resource added later cannot reintroduce the class — and it is what makes findings (6) non-blocking.

✅ Deriving both loaders' bucket key from one `SubDir → FilterName` table removes the two-conventions problem at its root, rather than special-casing the six nested parents at the call site.

</details>


<sub>🤖 Generated by the `pr-review:review` skill · reviewed head `5c9b750`</sub>

neilmartin83:
Addressed in 7ba8986. All six findings, blocking and nice-to-have.

**(3) is the one that shaped the fix.** Rather than sorting the walk's shared-bucket reads, I took the suggested route: the loader now reads `BackupResources` by path instead of discovering the tree. One `os.ReadDir` of the root for the directories no curated `SubDir` claims (keyed by name — `blueprints`, hand-assembled trees, and files left directly in a nested parent), then every table entry in slice order. Both loaders now merge with `maps.Copy` in the same order, so the entry listed *last* wins a collision on both sides. `TestLoadSnapshotFromDirectory_SharedBucketCollisionMatchesLiveOrder` pins the winner for `profiles` and `accounts` by reading it out of the table, so reordering `BackupResources` moves the expectation with it — verified failing against a lexically-ordered pass (`kept the object from macos, want "ios"`).

**(1)** Fixed, but not with a `groupName` branch — that would have been the third hard-coded field. `backupObjectName` now takes the resource's `BackupEndpoint.NameField`, the same field `extractName` checks first, so `groupName` and `displayName` both come from the registry and a resource added later with a new name field needs no edit here. `TestLoadSnapshotFromDirectory_EveryCuratedNameFieldIsHonoured` is the table guard: for every curated resource it writes an object carrying *only* its declared `NameField`. Your suggested `MatchesLiveNameExtraction` case is in too, and both fail without the branch.

**(2)** Fixed. There is no walk callback left to swallow the root error — `os.ReadDir(dir)` failing returns an error out of the loader. Covered at both levels: `TestLoadSnapshotFromDirectory_UnreadableRootIsAnError` and `TestRunDiff_UnreadableSourceDirectoryErrors` (mode `0111`, skipped as root and on Windows). Confirmed with the binary: `exit=1` with `permission denied` where it previously printed `No differences found.`

**(4)** You're right, and the wording is corrected in the comment, the test, the PR body and the postmortem: following a symlinked resource directory is a capability this loader *adds*. Kept deliberately — the table paths follow symlinks for free, and `entryIsDir` makes the root-level scan consistent with that rather than silently inconsistent. The traversal consequence for an untrusted tree is recorded in the postmortem. The symlink test now covers both routes (a curated `scripts` symlink and an unclaimed `blueprints` one).

**(5)** Fixed by the (3) rewrite: a nested parent is not an exact `SubDir`, so the root scan reads it under its own name — which for all six parents is already the `FilterName`. `TestLoadSnapshotFromDirectory_ObjectFilesInNestedParent` asserts the parent's files and the subdirectory's merge into one bucket.

**(6)** Gone with `nestedParents` — there is no prefix registration left to get wrong, and a three-segment `SubDir` is read like any other table entry.

Also from the review notes: `BackupSubDirs()` is now built once per load rather than twice, and its doc says what `diff` uses it for.

ktn-jamf (approved):
> [!NOTE]
> ✅ **Merge-ready** — all three blocking findings from round 1 are fixed, and each fix is pinned by a test I confirmed fails when the fix is reverted.
> The three nice-to-haves are addressed too. Two new 🟩 suggestions below; neither blocks.

## Rating: 5/5
- Clean: no open 🟥 or 🟧. Round 1's (3) was taken as the suggested structural rewrite rather than a patch — the loader now reads `BackupResources` by path instead of discovering the tree, which deletes `nestedParents`, the four-branch `switch`, the `fs.SkipDir` handling and findings (5) and (6) along with them. Highest-value follow-up is 🟩 **(1)**: `compliance-benchmarks` still mis-keys on a directory-vs-instance diff.
- Coverage: no specialist lane has ever run on this PR — `security-reviewer`, `usability-reviewer`, `fidelity-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate` and `test-quality-reviewer` were covered inline by the orchestrator because subagent dispatch is unavailable in this environment, which is not the same as the lanes having run. The rating reflects the dimensions that were searched.

<details>
<summary>Prior findings status (6 items)</summary>

| (1) c=100 | `internal/commands/pro_diff.go:296` | ✅ Fixed | Fixed generically rather than with a `groupName` branch: `backupObjectName` now takes the resource's `BackupEndpoint.NameField` and checks it first. Mutation-verified — deleting the `nameField` branch fails `TestBackupObjectName_MatchesLiveNameExtraction/mobile-device_group_uses_groupName` **and** the new table guard `TestLoadSnapshotFromDirectory_EveryCuratedNameFieldIsHonoured` |
| (2) c=100 | `internal/commands/pro_diff.go:170-174` | ✅ Fixed | No walk callback left to swallow it — `os.ReadDir(dir)` failing returns an error out of the loader (`:182-184`). Mutation-verified at both levels: turning the return into a warning fails `TestLoadSnapshotFromDirectory_UnreadableRootIsAnError` **and** `TestRunDiff_UnreadableSourceDirectoryErrors` |
| (3) c=75 | `internal/commands/pro_diff.go:154-168` | ✅ Fixed | Curated pass iterates `ResolveBackupResources(nameFilter)` in `BackupResources` slice order (`:206-211`), the same order and the same `maps.Copy` last-write-wins the live loader uses (`:396-404`). Mutation-verified: sorting `defs` by `SubDir` (what `WalkDir` did) fails `TestLoadSnapshotFromDirectory_SharedBucketCollisionMatchesLiveOrder` with `kept the object from macos, want "ios"` |
| (4) c=100 | `internal/commands/pro_diff.go:176-183`, `pro_diff_test.go:903-905` | ✅ Fixed | Wording corrected in all four places the finding named — the `entryIsDir` doc comment (`:217-223`), the test comment (`pro_diff_test.go:911-915`), the PR body, and the postmortem's new "One behaviour is genuinely new rather than preserved" paragraph. The traversal consequence for an untrusted tree is recorded in each. Behaviour kept deliberately, and the symlink test now covers both routes (curated `scripts`, unclaimed `blueprints`) — mutation-verified, an `entryIsDir` that stops following symlinks fails it |
| (5) c=100 | `internal/commands/pro_diff.go:210` | ✅ Fixed | A nested parent is no longer a special case: it is not an exact `SubDir` key, so the root scan reads it under its own name. `TestLoadSnapshotFromDirectory_ObjectFilesInNestedParent` asserts the parent's files and the subdirectory's merge into one bucket — mutation-verified (removing the root pass fails it, plus `UnknownTopLevelDirStillRead` and the symlink test) |
| (6) c=100 | `internal/commands/pro_diff.go:143-148` | ✅ Fixed | `nestedParents` and its `strings.Index` prefix registration are gone; a three-segment `SubDir` is read like any other table entry. Successor concern carried forward as 🟩 (2) below |

</details>

### Findings

Both are 🟩 — see the collapsed block.

<details>

🟩 **(1)** (correctness, c=100) — `internal/commands/pro_diff.go:288-306`: `compliance-benchmarks` still mis-keys, so every benchmark reports removed **and** added against a live instance.

`backup` writes a benchmark as `{"title": …, "description": …, "baselineId": …, …}` (`pro_backup.go:828-835`) — no `name`, no `displayName`, no `general`. It is an SDK-backed resource written straight to the backup root, so it comes through the root-level scan with `nameField == ""`, and `backupObjectName` falls all the way through to the filename stem. Live mode keys it on `bm.Title` (`pro_diff.go:464`). Reproduced by probe: a benchmark titled `CIS Level 1` gets disk key `"cis-level-1"` against live key `"CIS Level 1"`.

`blueprints` is fine — `blueprintExport` carries `name` (`pro_blueprints.go:309-310`), which is why the root-scan branch works for one of the two resources its comment names.

**Kept at 🟩, not 🟧, on the §4 tie-break:** the defect pre-exists this PR unchanged (the old loader read `obj["name"]` and missed it identically), it sits outside the `NameField` mechanism this PR added, and the new comment claims only that the root scan *buckets* these resources correctly, which it does. The postmortem's "Watch for" already names the class — "a resource whose live list name differs from anything in its detail body would break key parity again".

**Promote to 🟧 when:** key parity for the SDK-backed resources is treated as in scope for this PR rather than as a follow-up — it is the same user-visible failure the PR's title is about, reachable today on any platform-auth profile with `pro backup` + `pro diff --source ./backup --target <platform-profile>`.

**Suggested fix** — a `title` branch is the two-line version; passing the key explicitly is the version that cannot rot:
```diff
 	if n, ok := obj["displayName"].(string); ok && n != "" {
 		return n
 	}
+	// compliance-benchmarks: backup writes the SDK's Title, and the live
+	// loader keys on bm.Title. No registry NameField covers it — the
+	// resource has no BackupEndpoint.
+	if n, ok := obj["title"].(string); ok && n != "" {
+		return n
+	}
 	if general, ok := obj["general"].(map[string]any); ok {
```
Worth a case in `TestBackupObjectName_MatchesLiveNameExtraction` and, better, a parity assertion over the two `nonStandardBackupFilters` that `diff` actually loads — nothing in the suite covers the SDK-backed pair today.

🟩 **(2)** (correctness/evolution, c=75) — `internal/commands/pro_diff.go:169-195`: the root scan assumes a nested `SubDir`'s first segment *is* the `FilterName`, and nothing pins it.

The comment states the assumption plainly — "for those the directory name is already the `FilterName`" — and I verified it holds for all six nested parents today (`profiles`, `prestages`, `extension-attributes`, `smart-groups`, `static-groups`, `accounts`). But add a `{SubDir: "foo/bar", FilterName: "baz"}` entry later and loose files in `foo/` land in a `"foo"` bucket the live loader never produces, so they all report as removed. `EveryCuratedSubDirIsRead` and `EveryCuratedNameFieldIsHonoured` both pass regardless, and `ObjectFilesInNestedParent` only covers `profiles`, so no test catches it — unlike round-1 (6), which the table guard did catch.

This is the same shape as round-1 (6): reachable only after a future table change, hence 🟩. Cheapest guard is a loop over `BackupResources` asserting `strings.SplitN(r.SubDir, "/", 2)[0] == r.FilterName` for every nested entry — five lines, and it fails at the moment the assumption is broken rather than at the moment a user notices.

</details>

<details>

- [x] Design and architecture: the rewrite is the right shape. Reading the `SubDir → FilterName` table instead of discovering the tree removes the iteration-order question rather than answering it, and deletes `nestedParents`, the four-branch `switch`, the `fs.SkipDir` bookkeeping and the symlink special case with it. The loader is shorter than the version it replaces and has strictly fewer branches, for strictly more correct behaviour
- [x] Correctness: re-traced both passes; confirmed the six nested parents' directory names all equal their `FilterName`; confirmed the root pass runs *before* the curated pass so the layout `backup` writes wins a collision; confirmed `curatedDirs` keys are full slash-paths so a top-level `profiles` is never mistaken for a claimed directory while a top-level `scripts` is; confirmed live-side merge (`:396-404`) and disk-side merge use the same `defs` order and the same `maps.Copy`. Findings (1) and (2) above
- [x] Test coverage — **mutation-verified, five mutations in a scratch copy, each reverting one fix**: (a) drop the `nameField` branch → 2 tests fail; (b) root `ReadDir` error becomes a warning → 2 tests fail, at the loader and at `runDiff`; (c) sort the curated pass lexically by `SubDir` → collision test fails naming the wrong winner; (d) remove the root-level pass → 3 tests fail; (e) `entryIsDir` stops following symlinks → symlink test fails. No mutation passed silently. `go test ./internal/commands/ -count=1` green (9.7s), `golangci-lint run ./internal/commands/...` **0 issues**, `go vet` clean, `go build ./...` clean
- [x] Error paths, probed rather than read: unreadable **root** (`0111`) → error, exit non-zero ✅; **missing** curated directory → silently skipped, correct and correctly scoped to `fs.ErrNotExist`; curated path that is a **file** → `ENOTDIR` is *not* suppressed, warns ✅; unreadable curated **subdirectory** → warns and continues with a partial snapshot, exit 0 — same as the pre-PR loader, and it does warn loudly, so not raised as a finding; **malformed JSON** → warns, skips the file, reads the rest; **empty** directories both sides → empty snapshot, no error, "No differences found", which is correct and already flagged in the postmortem's "Watch for"
- [x] Rewrite-regression hunt: resources on disk but absent from the table are still read (`UnknownTopLevelDirStillRead`, mutation-pinned); unknown *nested* directories are dropped, exactly as the old walk's `default` branch did; `packages/files` still excluded (`IgnoresPackageFiles`); symlinked resource directories work on both routes; files directly inside a nested parent are now read where round-1 (5) dropped them
- [x] `internal/commands/pro_resources.go`: the +8/−3 is **comment-only** — the `BackupSubDirs` doc block. `BackupResources`, `ResolveBackupResources` and `BackupSubDirs()` are behaviourally untouched, so `backup` is unaffected. `BackupSubDirs()` is now built once per load rather than twice
- [x] Security (inline): no credential, network or dependency surface. The one behaviour change is symlink following, now deliberate and documented in four places — `diff` will read and print field values from outside a backup tree it is handed, which is a consideration only for an untrusted tree on a local user-owned path. No walk left, so no symlink loop or unbounded descent
- [x] Performance (inline): one `os.ReadDir` of the root plus one per curated `SubDir` (28), against a full-tree `WalkDir` before. `BackupSubDirs()` built once. Strictly cheaper than both the round-1 head and the base
- [x] Reliability / silent-failure (inline): the finding-(2) fix restores base behaviour for the root and the tests pin it at the command boundary, which is where the exit code is decided. Two pre-existing exit-0 paths remain, both unchanged by this PR and both out of its scope: a bogus `--resources` token on a directory-vs-directory diff yields an empty snapshot and "No differences found" (the live loader errors on the same input, `:354-356`), and `--resources blueprints` against a live profile errors in `loadSnapshotFromProfile` before the platform block that would have served it. Neither line is in this diff
- [x] Documentation currency: the postmortem is rewritten to describe the table read rather than the walk, corrects both statements round-1 (4) and (5) flagged, and gains the new guardrail tests. Frontmatter convention preserved. PR body matches the code
- [x] Code quality: comments remain load-bearing and I found no remaining inaccuracy in them. `entryIsDir` is a small, well-named, single-purpose helper
- [x] Project rules compliance: no generated file edited; the curated set is still owned by `internal/commands/pro_resources.go` per the "Where to Make Changes" table; `docs/solutions/` frontmatter matches its siblings
- [na] Frontend concerns: no UI in this repo
- [na] Cross-repo contracts: no wire key, enum, endpoint or shared config change; every caller of the three changed signatures is inside `internal/commands` (verified by grep)

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `internal/commands/pro_diff.go` | 5/5 | The rewrite is smaller and stricter than what it replaces. Two 🟩 remain, neither reachable from the curated table as it stands today |
| `internal/commands/pro_diff_test.go` | 5/5 | The strongest part of the round. Five independent mutations, one per fix, each caught — including the two the author says he verified himself. Two table-driven guards (`EveryCuratedSubDirIsRead`, `EveryCuratedNameFieldIsHonoured`) hold the class rather than the symptom, and `SharedBucketCollisionMatchesLiveOrder` reads its expectation out of `BackupResources` so reordering the table moves the assertion with it |
| `internal/commands/pro_resources.go` | 5/5 | Comment-only |
| `docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md` | 5/5 | Rewritten to match the new loader; both statements corrected by round-1 (4) and (5) are fixed, and the dotted-`NameField` caveat is sharpened |

</details>

<details>

Incremental diff `5c9b750..7ba8986` (4 files, +354/−111) in the worktree at `/tmp/pr-review-Jamf-Concepts-jamf-cli-332`. Because the head commit is a structural rewrite of the loader rather than a patch to three sites, the 181 changed lines in `pro_diff.go` were reviewed on their own merits, not only against the prior findings.

Read beyond the incremental diff: `internal/commands/pro_diff.go` in full (both loaders, `readObjectsFromSubdir`, `backupObjectName`, `compareSnapshots`), `internal/commands/pro_resources.go` in full, `internal/commands/pro_backup.go` (`backupBlueprints`, `backupBenchmarks`, slug/dedup), `internal/commands/pro_blueprints.go` (`blueprintExport`), `internal/commands/pro/generated/backup_registry.go` (`NameField`/`IDField` for every curated key), the full whole-PR diff `027396a..7ba8986` for `pro_diff.go`, plus `git show 027396a:` on the loader to confirm the base's root-`ReadDir` behaviour first-hand. PR body and the author's round-1 reply comment read in full; no other reviewer comments and no review-thread comments exist.

**Verified by executing, not by reading** — the gap round 1 flagged in itself: the five fix-reverting mutations listed under Review coverage, run in a scratch copy of the worktree (`go test ./internal/commands/`), plus four behavioural probes (compliance-benchmark key, empty directories, bogus `--resources`, malformed JSON) and three error-path probes (unreadable curated subdirectory, curated path that is a file, `--source` pointed at a leaf directory). The scratch copy is separate from the worktree; no git operation was run against either.

Risk tier: **medium** for the increment — 181 source lines, but they replace a guard that decides what the rest of the path sees, which the dispatch reference tiers high regardless of line count.

Never-run lanes, all eight: `security-reviewer`, `usability-reviewer`, `fidelity-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate`, `test-quality-reviewer` — subagent dispatch is unavailable in this environment, so each lane's checks were performed inline by the orchestrator and the lane record stays empty so a later round retries them. Inline results worth recording: the **test-quality** pass is the one that changed this round's outcome, since the five mutations are what turned "the commit message says it is fixed" into evidence; the **fidelity** pass confirms issue #331's stated expectation is met for both directory-vs-directory and directory-vs-instance, with 🟩 (1) the one resource still short of it; the **scope** pass keeps this as one PR (single defect, its regression tests, the convention postmortem); the **devil's-advocate** probes are the three error-path cases above plus the root-pass-before-curated-pass ordering question, which is deliberate and documented.

</details>

<details>

✅ Round 1 asked for one merge order shared by both loaders and offered "read the table instead of walking it" as the way to get it. The author took the rewrite rather than the cheaper sort, and it paid for itself: three separate findings — (3), (5) and (6) — closed with one change, and the `nestedParents` map, the four-branch `switch`, the `fs.SkipDir` handling and the symlink special case all disappeared rather than being individually patched.

✅ Finding (1) was fixed *better* than it was reported. The suggested fix was a `groupName` branch; the author declined it as "the third hard-coded field" and plumbed `BackupEndpoint.NameField` through instead, so `groupName` and `displayName` both come from the registry and a resource added later with a new name field needs no edit here. `EveryCuratedNameFieldIsHonoured` then guards that over the whole table.

✅ `SharedBucketCollisionMatchesLiveOrder` reads its expected winner out of `BackupResources` with a `lastLeaf` helper instead of hard-coding `ios`/`groups`. Reordering the table moves the assertion with it, which is the difference between a test that pins the invariant and a test that pins today's slice.

✅ Every fix is pinned by a test that fails without it — I checked all five by reverting them one at a time, and the author's claim that "every new test was verified to fail against the behaviour it pins" held on every one.

</details>


<sub>🤖 Generated by the `pr-review:review` skill · reviewed head `7ba8986`</sub>
</content>
</invoke>

PR #335, opened 2026-08-20
fix(generator): share one scaffold walker, and show what an array holds

neilmartin83:
## Problem

`--scaffold` had **three** implementations that had drifted into three different answers for the same schema:

| | read-only | spec examples | nested objects | arrays |
|---|---|---|---|---|
| `generator/parser/generator.go` `scaffoldJSON` (Pro) | skipped | honoured | **not recursed** — emitted `{}` | `[]` |
| `generator/platform/emitter.go` `schemaExample` | **kept** | **ignored** | recursed | `[]` |
| `generator/security/emitter.go` `schemaExample` | **kept** | **ignored** | recursed | `[]` |

The platform and security versions were byte-identical copies of each other. So a scaffold meant something different depending on which API served the command, and nothing failed while they disagreed.

That last column was the expensive one. `pro platform-device-groups create --scaffold` printed `"criteria": []` for a **seven-field** element, and the only feedback on a wrong guess was a 400 from the server.

Worse for a **bare-array request body**: the Pro generator gated `--scaffold` on `len(Schema.Properties) > 0`, and an array body has none — so `pro app-requests update`, the one Pro endpoint whose body is a top-level array, shipped with **no `--scaffold` flag at all**.

### Before / after

```
# pro platform-device-groups create --scaffold
-  "criteria": [],
+  "criteria": [
+    {
+      "attributeName": "Device Name",
+      "attributeValue": "Some Device Name",
+      "hasClosingParenthesis": false,
+      "hasOpeningParenthesis": false,
+      "joinType": "",
+      "operator": "IS",
+      "order": 0
+    }
+  ],
```

```
# pro app-requests update --scaffold
- unknown flag: --scaffold
+ [
+   {
+     "description": "How many of these would you like?",
+     "priority": 1,
+     "title": "Quantity"
+   }
+ ]
```

## Change

One exported walker — `parser.ScaffoldJSON` — plus `parser.HasScaffoldShape` for the emit gate. All three generators delegate; the duplicated trios are gone. The parser now captures array element schemas it previously read only for query parameters: `Property.Items` for an array-typed property, `Schema.Items` for a schema that *is* an array.

Consolidated rules, and why each:

- **Read-only omitted.** It is a request template. This was Pro's rule; unifying it changed no output for the other two, because no platform or security request body declares a read-only property at any depth (walked all of them; the only `readOnly` in `specs/platform/` sits in response schemas).
- **Write-only kept.** The asymmetry is the point: passwords and secrets never come back from a GET, so a scaffold is the only place a caller sees them. See `write-only-fields-dropped-by-update-set-2026-07-22.md`.
- **A spec example beats a placeholder.** `"searchType": "is"` teaches the format; `""` does not.
- **An array shows one element when the element is an object**, and stays empty otherwise. The object shape is what a caller cannot guess; a scalar element's type is evident from the field name, and `[""]` would imply an empty string is a meaningful entry.

## The recursion cap is load-bearing, not defensive

`maxSchemaDepth` (8) bounds `parseSchemaDepth`. Object nesting never needed it — a property whose own `Properties` map is empty ends the walk. **Array elements break that:** a schema may name itself as its own element type (a node with a `children[]` of nodes), and kin-openapi resolves `$ref` inline, so following element schemas without a cap recurses until the stack dies. Populating `Items` therefore *introduced* a hang risk that did not previously exist.

- `TestParseSchema_RecursiveArrayTerminates` covers the cycle — note it would **hang** rather than fail without the cap.
- `TestParseSchema_DepthCapUnreachedByLiveSpecs` asserts nothing is truncated today. Deepest schema across every committed spec is **3** (`ComputersInventory.yaml` → `storage.disks[].partitions[]`), so a future ingest that pushes past the cap fails loudly instead of silently shortening scaffolds.

## Trade-off taken deliberately

Filling a synthesised array element with spec examples can stop a scaffold posting verbatim. `advanced-mobile-device-searches create --scaffold` previously emitted `"criteria": []`, which the server accepted; it now emits an element whose example criterion name (`"Account"`) the test tenant rejects with `400 INVALID_FIELD "The criterion Account is not valid"`.

Kept anyway, because verbatim-postability was never a property worth protecting — most creates already carried `"name": ""` and failed validation — and zero-valuing element fields would throw away the useful part: `"andOr": "and"` and `"searchType": "is"` document enum-ish values that `""` cannot. `--scaffold` prints *an example request body* to edit, not a valid document.

## Verification

Wire-verified against a live Jamf Pro tenant:

- The generated `criteria` element for `advanced-mobile-device-searches` matches the field set a real record returns exactly — `andOr`, `closingParen`, `name`, `openingParen`, `priority`, `searchType`, `value`.
- The scaffold posted successfully once the example criterion was replaced with one the tenant has, was read back faithfully, and deleted cleanly.

Regenerating touched 56 generated files. The largest scaffold, `computers-inventory create`, went from flat to 193 lines of real nested shape with spec examples throughout.

`make test`, `make lint`, `make verify-generated`, `make verify-site` all green. All six pre-existing `scaffoldJSON` tests pass **unchanged** — the shared walker preserves Pro's documented contract (`nil`/empty → `{}`, deterministic alphabetical order, read-only skipped, examples honoured).

## Notes for review

- No behaviour change for platform or security read-only handling, at any depth — neither has a read-only property in any request body. Unifying that rule is a guard against a future spec, not a visible change.
- `docs/solutions/conventions/one-scaffold-walker-2026-08-20.md` records the convention, including "never add a fourth scaffold builder".

---

## Review round 1 — `ktn-jamf`

All eight findings addressed in `f3dbb44`. **No generated output changes** — all three trees regenerate byte-identical, so every item below is a guard rather than a behaviour change.

**(1) 🟧 the two `Items` hunks had no test.** Correct, and the sharpest catch here: every new test built its `parser.Schema` by hand, so none could notice the parser stopping deriving a shape from a spec. Three tests now parse a spec end to end — `TestParseSchema_ArrayPropertyItemsFromSpec` pins `Property.Items` alone so a failure names the cause; `TestParseSchema_ArrayRequestBodyEndToEnd` pins `Schema.Items` through parse → `HasScaffoldShape` → render, and also checks read-only omission reaches *inside* a synthesised element; `TestParseSchema_RecursiveArrayTerminates` now asserts the element's shape rather than just the `children` key. Mutation-verified: deleting the top-level branch fails the array-body test, deleting the property branch fails the other two.

**(2) 🟩 `hasApplyScaffold`.** Fixed as suggested.

**(3) 🟩 untyped property renders `null`.** Fixed. `scaffoldArray` and `scaffoldValue` both already treat a declared type as unreliable; `propertyValue` now agrees, and a property with no resolved shape still renders `null` — matching what the default arm gave `rules[].odv`.

**(4) 🟩 canary glob.** Added `specs/security/*.json`. Deepest is still 3.

**(5) 🟩 swallowed marshal error.** `ScaffoldJSON` returns `(string, error)`; every caller propagates and generation aborts. The six pre-existing `scaffoldJSON` tests keep their assertions via a `mustScaffold` helper.

**(6) 🟩 raw Go string literal — done differently.** Took the risk seriously but kept the raw literal, because `%q` would collapse each scaffold to one escaped line, and a `make generate` diff is the review artefact for a spec ingest that rewrites dozens of them at once. Instead `scaffoldRawLiteral` fails generation with the offending scaffold in the message, so a backtick can never reach the generated package. The asymmetry with Platform/Security is now written down with its reason.

**(7) 🟩 prose.** All three fixed: the ZTNA read-only claim is gone from the body above (verified — zero read-only properties at any depth in either spec tree), the security emitter's comment named its own file as its sibling, and "would hang" is now "stack-overflows in about a second", which is the better guard.

**(8) 🟩 per-generator gate + CI.** Both halves. Platform and Security now gate `--scaffold` on a `HasScaffold` field from `parser.HasScaffoldShape`; `--file`/`--set` stay on `HasBody`, since a body can be worth sending without being worth scaffolding. And CI ran `verify-generated` only, which diffs the Pro tree alone even though `make generate` writes all three — `verify-platform-specs` and `verify-security-specs` now run there too, which is what makes a shared-walker change safe to land.

`make test`, `make lint`, `make verify-generated`, `make verify-platform-specs`, `make verify-security-specs`, `make verify-site` all green.


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Collapses three drifted `--scaffold` builders into one shared walker (`parser.ScaffoldJSON`) and teaches it to show an array's element shape.
> Blocking: (1). See collapsed section for 7 nice-to-have suggestions.

## Rating: 3/5
- Would be a 5 with a test that actually pins the parser-side `Items` population in (1) — I deleted both hunks and the whole suite stayed green.
- Coverage: no specialist agent ran on this PR; the `security-reviewer`, `usability-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate` and `test-quality-reviewer` lanes were all covered inline by the orchestrator (subagent dispatch unavailable in this run), including bounded mutation spot-checks. See Scope of review.

### Findings

🟧 **(1)** (test coverage, c=100) — `generator/parser/parser.go:1775` and `generator/parser/parser.go:1831`: the two hunks that are this PR's actual mechanism have no test

Both `Items`-population branches can be deleted and `go test ./...` stays green. I removed the top-level `s.Items` block, ran the suite: all packages `ok`. I removed the `p.Items` block: all packages `ok`. Every new test builds its `parser.Schema` by hand, so nothing asserts that `parseSchema` derives `Items` from a spec. `TestParseSchema_RecursiveArrayTerminates` only checks the `children` **key** exists, which it does even when the value is `[]`; `TestParseSchema_DepthCapUnreachedByLiveSpecs` still passes at a shallower measured depth. The only remaining net is `make verify-generated`, which reports "generated code is out of date" — a message whose documented remedy is `make generate && commit`, i.e. commit the degraded output.

**Failure scenario:** a later refactor narrows or drops either `Items` branch → every array in every scaffold silently reverts to `[]`, the parser package tests pass, and CI's only complaint is a golden-file drift whose prescribed fix is to accept the new output. Present as shipped: the guard does not exist.

**Suggested fix:**
```diff
--- a/generator/parser/scaffold_test.go
+++ b/generator/parser/scaffold_test.go
@@ TestParseSchema_RecursiveArrayTerminates
 	if _, ok := parsed["children"]; !ok {
 		t.Errorf("expected children rendered, got %s", out)
 	}
+	// The element shape is the point — a bare [] would satisfy the key check.
+	kids, ok := parsed["children"].([]any)
+	if !ok || len(kids) != 1 {
+		t.Fatalf("expected one example child element, got %#v", parsed["children"])
+	}
+	if _, ok := kids[0].(map[string]any)["name"]; !ok {
+		t.Errorf("expected the parsed element schema, got %#v", kids[0])
+	}
 }
```
Plus one sibling test that parses a spec whose request body **is** an array (mirroring `AppRequest.yaml`) and asserts `HasScaffoldShape(op.RequestBody.Schema) == true` and that `ScaffoldJSON` yields a one-element array — that covers the `Schema.Items` branch and the emit gate that the headline fix depends on.

**Fixed when:** deleting either `Items`-population branch in `parser.go` fails at least one test in `./generator/parser/`, and a spec-parsed bare-array request body is covered end to end (parse → `HasScaffoldShape` → `ScaffoldJSON`).

<details>
<summary>Nice-to-have suggestions (7 items)</summary>

🟩 **(2)** (design, c=100) — `generator/parser/generator.go:400`: `hasApplyScaffold` still gates on `len(Properties) > 0`

Three of the four scaffold gates moved to `HasScaffoldShape`; the `apply` gate did not. The solutions doc added in this same PR states the rule as "Gate `--scaffold` on `parser.HasScaffoldShape`, not on `len(Properties) > 0`. The latter silently denies the flag to array bodies."

**Suggested fix:**
```diff
-if op.Name == "create" && op.Method == "POST" && op.RequestBody != nil && op.RequestBody.Schema != nil && len(op.RequestBody.Schema.Properties) > 0 {
+if op.Name == "create" && op.Method == "POST" && op.RequestBody != nil && HasScaffoldShape(op.RequestBody.Schema) {
```

`Promote to 🟧 when:` a spec ingest adds a resource whose `create` POST body is a bare array with object items — today the only array body in `specs/` is `app-requests update` (PUT), so the divergence is invisible.

🟩 **(3)** (correctness, c=75) — `generator/parser/scaffold.go:88`: `propertyValue` renders an untyped property as `null`, discarding a resolved `Nested`

`scaffoldArray` (line 108) says it outright — "Type is not a reliable test on its own: plenty of spec schemas carry properties without declaring `object`" — and `scaffoldValue` (line 52) treats `""` as an object. `propertyValue` matches only the literal `"object"`, so a property with `properties`/`allOf` and no declared `type` falls off the switch and returns `nil`, even though `parser.go:1825` already populated `p.Nested` for it. I probed every request body in `specs/`, `specs/platform/` and `specs/security/`: exactly one untyped property exists today (`compliance_benchmark_engine.json` → `rules[].odv`, a bare `oneOf[OdvRequest, null]` with no `Nested`), so nothing regresses now.

**Suggested fix:**
```diff
-	case "object":
+	case "object", "":
 		if p.Nested != nil {
 			return scaffoldValue(p.Nested)
 		}
-		return map[string]any{}
+		if p.Type == "object" {
+			return map[string]any{}
+		}
```

`Promote to 🟧 when:` any spec declares a request-body property with `properties` or `allOf` and no `type` — that property's whole shape then renders as `null`.

🟩 **(4)** (test coverage, c=100) — `generator/parser/scaffold_test.go:269`: the depth-cap canary skips `specs/security/*.json`

The glob is `{"../../specs/*.yaml", "../../specs/platform/*.json"}`, but the shared walker also renders the three Security Cloud specs (I measured their deepest request-body schema at 1). Both the test's own comment and the solutions doc claim the assertion holds "across every committed spec", so the stated guarantee is wider than the glob. One-line fix: add `"../../specs/security/*.json"`.

🟩 **(5)** (silent failure, c=100) — `generator/parser/scaffold.go:43`: a marshal error becomes an empty scaffold, at generation time

`ScaffoldJSON` swallows `json.MarshalIndent`'s error into `"{}"`. In a build-time generator that is the wrong direction: the operation ships a useless scaffold and `make generate` exits 0. This is the exact failure mode `TestParseSchema_DepthCapUnreachedByLiveSpecs` exists to prevent one tier up ("fails loudly instead of silently shortening scaffolds"). No live instance today — I checked every request body where `HasScaffoldShape` is true and none renders as `{}`. Have `ScaffoldJSON` (or a new `ScaffoldJSONErr`) return the error and let the generator abort.

🟩 **(6)** (robustness, c=75) — `generator/parser/generator.go:1909` and `:2903`: Pro embeds the scaffold in a raw Go string literal

Platform and Security both emit theirs through `{{printf "%q" .Scaffold}}`; the Pro template interpolates into a backtick literal. Spec examples now reach the scaffold from nested objects and array elements as well as the top level, so the surface for an example containing a backtick is much wider than before, and one would emit uncompilable generated code. Zero occurrences today (verified across all specs). Switching Pro to `%q` closes it and matches the other two.

🟩 **(7)** (documentation, c=100) — three inaccuracies in the prose

- The PR description's read-only bullet and its "Notes for review" say unifying read-only handling "drops nested read-only fields the other two showed, e.g. ZTNA's `dedicatedIps.ips`". No platform or security request body declares a read-only property at any depth (I walked all of them), and neither `dedicatedIps` nor `ztna` appears anywhere in the repo. The commit message and `docs/solutions/conventions/one-scaffold-walker-2026-08-20.md` say the opposite and are correct: it changed no output.
- `generator/security/emitter.go:220-223` — "shared with the Jamf Pro and **Security Cloud** generators" and "byte-identical between this file and **generator/security/emitter.go**". Both should read Platform; the comment currently points at its own file.
- `scaffold_test.go:213` and the solutions doc say `TestParseSchema_RecursiveArrayTerminates` "would hang rather than fail without the cap". With `maxSchemaDepth` raised, it terminates in ~a second with `fatal error: stack overflow` and the test fails. The guard is better than advertised.

🟩 **(8)** (design, c=100) — the *gate* is still per-generator, and the two dirs this PR changed have no CI golden check

The rendering is now shared, but Platform (`generator/platform/template.go:111`) and Security both gate `--scaffold` on `.HasBody`, so an empty-shape body there gets a `--scaffold` that prints `{}` where Pro suppresses the flag entirely. The new convention doc's second rule ("Gate `--scaffold` on `parser.HasScaffoldShape`") therefore only landed in one of three generators. Separately, `.github/workflows/ci.yaml` runs `make verify-generated` (Pro only) — `verify-platform-specs` and `verify-security-specs` are not in CI, so the platform and security generated output this PR rewrites is unguarded there. I regenerated all three locally and the committed output is byte-identical, so nothing is stale today.

</details>

<details>

- [x] Design and architecture: one walker for three generators is the right consolidation; the residual asymmetries are (2), (3) and (8). Confirmed Classic has no scaffold builder, so "all three generators" is exhaustive — `generator/classic/generator.go` has no scaffold code at all.
- [x] Correctness: traced `parseSchemaDepth` → `Property.Items`/`Schema.Items` → `scaffoldValue`/`propertyValue`/`scaffoldArray`. Confirmed the built schema tree is acyclic (each `parseSchemaDepth` call allocates a fresh `*Schema`), so `scaffoldValue`'s "no separate depth guard needed" comment is sound. Termination verified empirically: with the cap removed, `TestParseSchema_RecursiveArrayTerminates` stack-overflows; with it, ~0.2s. `oneOf`/`anyOf` elements degrade to `[]`/`null` rather than misrendering — see (3).
- [x] Security: no credential-shaped example reaches any scaffold (grepped the generated diff for password/secret/token/passphrase/key value pairs — no hits). Write-only retention is deliberate and matches `write-only-fields-dropped-by-update-set-2026-07-22.md`. No auth, client, or config code touched.
- [x] Performance: generation-time only. Depth cap 8 bounds depth but not breadth, and `$ref` expansion is exponential in depth; measured deepest live schema is 3 (`ComputersInventory.yaml` → `storage.disks[].partitions[]`), `make generate` completes in normal time, so not a present problem.
- [x] Test coverage: mutation spot-checks — `scaffoldArray`→always-empty fails 2 tests; read-only skip removed fails 2 tests; both `Items` branches deleted fails **nothing**, which is finding (1). Both depth-cap tests confirmed meaningful. All six pre-existing `scaffoldJSON` tests are untouched and pass.
- [x] Reliability: `make verify-generated`, `verify-platform-specs` and `verify-security-specs` would all pass — I ran `make generate` in a scratch copy of the head and diffed all three generated dirs against the committed output: identical. Full `go test ./...` green; `golangci-lint run ./generator/...` 0 issues; `gofmt -l generator/` clean.
- [x] Code quality: naming and doc-comment density match the surrounding generator code. `hasScaffold` (`generator.go:1696`) is registered in the funcmap but referenced by no template — pre-existing dead code the PR touches; not worth a finding on its own.
- [x] Simplification: removes ~110 lines of duplicated walker. `scaffoldJSON` is now a one-line wrapper kept for the six existing tests, which is reasonable.
- [na] Frontend concerns: no `docs/site/` or UI changes; `commands.json` is generated at deploy time, not committed.
- [x] Documentation currency: `CLAUDE.md`, `generator/README.md`, and a new `docs/solutions/conventions/` doc all updated; the doc matches sibling frontmatter shape. Accuracy issues in (7).
- [na] Cross-repo contracts: nothing here crosses a repo boundary. The scaffold is advisory output a human edits before sending — no wire key, enum, endpoint, or shared storage key changes. Request-body *shapes* come from the committed specs, which this PR does not modify (`git diff --stat` shows no `specs/` change).
- [x] Project rules compliance: checked `CLAUDE.md` (no `.claude/rules/` directory exists). Generated-code boundary respected — every `internal/commands/*/generated/` change is reproducible from `make generate`, verified byte-for-byte. Credential-input policy untouched. `docs/solutions/` grepped for `module: generator/parser` before reviewing.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/parser/parser.go` | 3/5 | The two `Items` branches are the change's mechanism and are untested — see (1) |
| `generator/parser/scaffold.go` | 4/5 | Clean walker, well-documented rules; `propertyValue`'s untyped arm (3) and the swallowed marshal error (5) |
| `generator/parser/scaffold_test.go` | 3/5 | Six good behavioural tests, but the parse boundary is uncovered (1) and the canary's glob is narrower than its claim (4) |
| `generator/parser/generator.go` | 4/5 | One of four gates un-migrated (2); raw-literal embedding (6) |
| `generator/platform/emitter.go` | 5/5 | Straight delegation, ~50 duplicated lines deleted, no behaviour change (verified: zero read-only properties in any platform request body) |
| `generator/security/emitter.go` | 4/5 | Same clean delegation; the new doc comment names the wrong sibling generator (7) |
| `internal/commands/pro/generated/` (56 files) | 5/5 | Not reviewed line by line, per the repo's generated-code boundary. Verified instead by regenerating from the head and diffing: byte-identical. The only structural change across all 62 generated files is the new `--scaffold` flag on `app-requests update`; everything else is scaffold string content |

</details>

<details>

Full diff (62 files, +2323/−336) at head `ef4d74e` against `origin/main` `027396a`. Read in full: `generator/parser/scaffold.go`, `scaffold_test.go`, the `parser.go` and `generator.go` hunks, both emitters, the new solutions doc, `CLAUDE.md` and `generator/README.md`. Generated output reviewed by regeneration equivalence rather than line by line, per `CLAUDE.md`'s generated-code boundary. Also read for context: `generator/parser/generator_test.go` (the six pre-existing scaffold tests), `generator/platform/template.go`, `generator/security/template.go`, `internal/commands/pro/generated/registry.go` (`printScaffoldOutput`), `Makefile` verify targets, `.github/workflows/ci.yaml`, `docs/solutions/README.md`.

Work done in a scratch copy of the head (the review worktree was left untouched): `go build ./...`, `go test ./...`, `golangci-lint run ./generator/...`, `gofmt -l`, `make generate` + three-way diff of the generated dirs, five bounded mutations (each reverted), and six throwaway probe tests over every committed spec (untyped properties, read-only properties in platform/security bodies, array-typed request bodies, per-scope schema depth, collapsed scaffolds, backticks in rendered scaffolds).

No specialist agents were dispatched — subagent dispatch was unavailable for this run — so the orchestrator covered every eligible lane inline: `security-reviewer`, `usability-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate`, `test-quality-reviewer` (including the mutation spot-checks above). Findings from those passes are tagged by dimension, not by lane. `fidelity-reviewer` is inapplicable: no Jira key in the title, body or branch name, and no linked issue — the PR description is the only spec, so a fidelity pass would have graded the change against itself. On decomposition: this reads as one concern correctly delivered as one PR — the generated churn is mechanical output of the single template change and splitting it would leave a commit where the three generators disagree.

Not verified: the PR's wire-verification claims against the `pro-nmartin` tenant (no tenant access), and whether `POST /v1/computers-inventory` accepts the ~40 nested `general`/`hardware` fields the 193-line scaffold now offers — the spec does not mark them read-only, so the walker is faithful to the spec whatever the server does.

</details>

<details>

No prior reviews, review comments, or issue comments on this PR.

</details>

<details>

✅ `TestParseSchema_DepthCapUnreachedByLiveSpecs` is the right kind of test — it asserts the cap is *unreached*, so a future spec ingest that would silently shorten scaffolds fails loudly instead. Rare and worth copying; the only gap is its glob (4).

✅ The reasoning about *why* the cap became load-bearing (populating `Items` made schema cycles reachable where object nesting had always self-terminated) is correct and non-obvious, and the recursion test confirms it — I removed the cap and got a stack overflow in under a second.

✅ Writing the convention down with a "never add a fourth scaffold builder" rule, rather than trusting three now-identical call sites to stay identical, is the fix that actually prevents the recurrence.

</details>



ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready** — Round 1's blocking finding is genuinely closed, and I verified it by mutation rather than by reading: deleting either `Items`-population branch in `generator/parser/parser.go` now fails tests in `./generator/parser/`. Six of the seven 🟩 items are fixed too. Two 🟩 remain, neither blocking.

## Rating: 5/5
- Coverage: no specialist agent ran on this PR in either round — subagent dispatch is unavailable in this run, so the `security-reviewer`, `usability-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate` and `test-quality-reviewer` lanes were covered inline by the orchestrator again, including the mutation spot-checks below. `fidelity-reviewer` stays dimension-inapplicable (no Jira key, no linked issue). See Scope of review.

<details>
<summary>Prior findings status (8 items)</summary>

| (1) 🟧 c=100 | `generator/parser/parser.go:1775`, `:1831` | ✅ Fixed | Mutation-verified. Deleting the `Schema.Items` block fails `TestParseSchema_ArrayRequestBodyEndToEnd`; deleting the `Property.Items` block fails `TestParseSchema_ArrayPropertyItemsFromSpec` **and** `TestParseSchema_RecursiveArrayTerminates`. Three new tests parse a spec end to end instead of hand-building a `Schema`, and the recursion test now asserts on the populated element (`children[0].name`), not the bare key |
| (2) 🟩 c=100 | `generator/parser/generator.go:404` | ✅ Fixed | `hasApplyScaffold` now gates on `HasScaffoldShape`. All four gates are migrated; no generated output changed |
| (3) 🟩 c=75 | `generator/parser/scaffold.go:97` | ✅ Fixed | `case "object", ""` with the `p.Type == "object"` discriminator preserved for the empty-object case. `TestScaffoldJSON_UntypedPropertyWithNestedShape` fails when the arm is reverted |
| (4) 🟩 c=100 | `generator/parser/scaffold_test.go:293` | ✅ Fixed | `specs/security/*.json` added, and it is not a decorative glob — I probed `ParseSpec` over the three security specs: all parse cleanly and contribute 5 request bodies with schemas. Deepest committed depth still 3 |
| (5) 🟩 c=100 | `generator/parser/scaffold.go:44` | ⚠️ Partial | `ScaffoldJSON` returns the error and every direct caller propagates it — but on the Pro path it lands in `generator/main.go:182-183`, which logs and `continue`s, so `make generate` still exits 0. See active finding (1) |
| (6) 🟩 c=75 | `generator/parser/generator.go:1675` | ✅ Fixed, differently | Kept the raw literal (for line-by-line `make generate` diffs) and added `scaffoldRawLiteral`'s backtick check with the reason written down. The risk the finding named — uncompilable generated code — is closed: I injected a backtick into `specs/Building.yaml`'s example and the generator refused to emit `buildings.go` |
| (7) 🟩 c=100 | prose (3 items) | ⚠️ Partial | ZTNA read-only claim removed from the PR body ✅; `generator/security/emitter.go:227-230` now names Platform as its sibling ✅; but the PR body's recursion-cap section still reads "note it would **hang** rather than fail without the cap" — the solutions doc and the test comment were both corrected, the description was not |
| (8) 🟩 c=100 | `generator/platform/template.go`, `.github/workflows/ci.yaml` | ✅ Fixed | Platform and Security gate `--scaffold` on a new `HasScaffold` field while `--file`/`--set` stay on `HasBody`; the templates are internally consistent (the `Args` branch moved too, so no generated code references an undeclared `scaffoldFlag`). CI now runs `verify-platform-specs` and `verify-security-specs`; I read both targets — each `rm`s the generated `.go` files, regenerates and `git diff --quiet`s, so they do fail on drift — and ran all three locally: byte-identical, zero residual diff. CI is green on `f3dbb44` |

</details>

### Findings

<details>

🟩 **(1)** (silent failure, c=100) — `generator/parser/scaffold.go:44` and `generator/parser/generator.go:1675`, swallowed at `generator/main.go:182-183`: the new render errors do not abort `make generate` on the Pro path

The two errors this round introduces are propagated correctly right up to `main.go`, where the Pro loop logs and continues:

```go
outPath, err := gen.Generate(resource)
if err != nil {
    fmt.Fprintf(os.Stderr, "Error generating %s: %v\n", resource.Name, err)
    continue
}
```

Platform (`main.go:274-278`) and Security (`:333-337`) both `os.Exit(1)`; Pro does not. Measured, not inferred — I set `example: "Apple \`Park\`"` in `specs/Building.yaml` and ran the generator:

```
Error generating buildings: executing template: ... error calling opScaffoldJSON:
scaffold for schema "" contains a backtick, ...
$ echo $?           → 0
$ ls .../generated/buildings.go → No such file or directory
$ go build ./...    → registry.go:49:18: undefined: NewBuildingsCmd
```

So the failure mode is not the old empty scaffold — it is a whole resource command silently dropped while `make generate` reports success. That makes two prose claims overstated: the new doc's "**A render failure must abort generation.** `parser.ScaffoldJSON` returns an error; every caller propagates it", and `scaffoldRawLiteral`'s "Fail at generation rather than at the build of the generated package, where the cause is not obvious." On Pro it fails at *both* — a stderr line at generation, then `undefined: NewBuildingsCmd` at build.

Two secondary notes on the same paths: `%q` on `s.Name` renders `schema ""` for an inline request-body schema, so the message identifies the resource only via the template-execution wrapper; and `TestParseSchema_DepthCapUnreachedByLiveSpecs` has the same shape of hole at `scaffold_test.go:329` (`if err != nil { continue }`), where a spec that stops parsing drops out of the assertion silently — pre-existing, but the same class the new error plumbing is trying to close.

**Mitigation that keeps this 🟩:** `make verify-generated` sees the deleted file and fails, and `make build` fails, so nothing reaches `main` broken. The cost is a local `make generate` that lies.

**Suggested fix:** `os.Exit(1)` after the per-resource error (or accumulate and fail at the end), matching what the Platform and Security paths already do. Reword the two claims to match whichever behaviour lands.

`Promote to 🟧 when:` a spec ingest introduces a backtick or an unmarshallable example — the exact case this guard was written for — and the ingest is committed off a green-looking `make generate` without a build.

🟩 **(2)** (test coverage, c=100) — `generator/parser/generator.go:404` and `generator/platform/emitter.go:347`: the newly migrated gates are themselves unpinned

Same shape as round 1's blocking finding, one tier down. I reverted `hasApplyScaffold` to `len(op.RequestBody.Schema.Properties) > 0` and `go test ./generator/...` stayed green; I reverted platform's `buildScaffold` to the old `Schema == nil` check and it stayed green too. The four gates now agree, and nothing fails if one drifts back.

Lower value than round 1's version because both produce byte-identical output on every committed spec today (verified — the incremental diff touches no generated file, and all three verify targets pass), so a test has to be spec-synthetic. `TestParseSchema_ArrayRequestBodyEndToEnd` already builds exactly the fixture needed: asserting `hasApplyScaffold`-equivalent behaviour on a `create` POST with a bare-array body would cover Pro's `apply` gate for a few extra lines.

`Promote to 🟧 when:` a spec ingest adds a resource whose `create` POST body is a bare array with object items, or a platform/security op whose body has a schema but no shape — then the divergence becomes user-visible as a missing or useless `--scaffold`.

</details>

<details>

- [x] Prior blocking finding, empirically: both mutations now fail. `--- FAIL: TestParseSchema_ArrayRequestBodyEndToEnd` (`bare-array request body should earn --scaffold; Items = (*parser.Schema)(nil)`) on deleting the `Schema.Items` block; `--- FAIL: TestParseSchema_RecursiveArrayTerminates` (`expected one example child element, got []interface {}{}`) plus `--- FAIL: TestParseSchema_ArrayPropertyItemsFromSpec` on deleting the `Property.Items` block. Each mutation applied to a scratch copy and reverted.
- [x] New tests judged on their own: five new tests, all meaningful. Mutation-checked four of the five in turn — reverting the untyped-property arm fails `TestScaffoldJSON_UntypedPropertyWithNestedShape`; re-swallowing the marshal error fails `TestScaffoldJSON_ReportsRenderFailure`; dropping the backtick guard fails `TestScaffoldRawLiteral_RejectsBacktick`. The `mustScaffold`/`mustScaffoldJSON` helpers keep the six pre-existing tests' assertions intact rather than weakening them to absorb the signature change.
- [x] Correctness of the new code: read `scaffold.go` in full. `propertyValue`'s new `case "object", ""` keeps the `p.Type == "object"` discriminator, so a declared empty object still renders `{}` and only a genuinely shapeless untyped property renders `null` — the switch's old default answer. `Example` is still checked before the type switch, so the new arm cannot shadow an example.
- [x] Templates: `scaffoldFlag` is declared, read in `Args`, read in `RunE` and registered as a flag under `{{- if .HasScaffold }}` in both Platform and Security, with a `cobra.ExactArgs` else-branch — no path emits a reference to an undeclared variable. `HasScaffold: scaffold != ""` is a sound derivation because `ScaffoldJSON` never returns an empty string on success.
- [x] Error plumbing: `tmpl.Execute` failure removes the partial output file (`generator.go:651-655`), so an aborted render leaves no half-written `.go`. The gap is one level up in `main.go` — finding (1).
- [x] CI change: read the two verify targets. Both `rm` the generated `.go` files, run `make generate`, and `git diff --quiet` the tree, so drift fails the step. The `sync-*-specs` half is a no-op in CI (`specs/.platform-source/` and `specs/.security-source/` are not committed, and the `; true` swallows the error) but the generated-code half always runs, which is the half this PR needed. All three pass locally with zero residual diff, and CI is green on `f3dbb44`.
- [x] Reliability / reproducibility: `make verify-generated`, `make verify-platform-specs`, `make verify-security-specs` all pass in an isolated scratch clone of `f3dbb44` — all three generated trees regenerate byte-identically and `git status --porcelain` is empty afterwards.
- [x] Test / lint / format: `go test ./...` green. `golangci-lint run` → 0 issues (checked three times, and against `ef4d74e` for comparison — an early run reporting `SA5011` was an artefact of two copies of the module sitting side by side in `/private/tmp`, not a real regression). `gofmt -l` and `gofumpt -l` clean over `generator/`.
- [x] Documentation currency: the solutions doc gained accurate sections on the two deliberate `type` loosenings and on Pro's raw literal, plus a new "do not test a scaffold rule only against a hand-built `parser.Schema`" rule that generalises round 1's finding (1) correctly. The two overstated claims are in finding (1); the PR body's "would hang" is in prior finding (7).
- [x] Security: no credential-shaped example reaches a scaffold; no auth, client or config code touched. Unchanged from round 1.
- [x] Performance: generation-time only. No generated output changed, so runtime behaviour is identical to `ef4d74e`.
- [x] Scope: the incremental diff is 10 files, all of them the fix or its test/doc. No unrelated drive-by changes, and no generated file touched — which is itself the load-bearing evidence that eight fixes changed zero output.
- [na] Frontend concerns: no `docs/site/` or UI changes.
- [na] Cross-repo contracts: unchanged from round 1 — nothing crosses a repo boundary and `specs/` is untouched.
- [x] Project rules compliance: generated-code boundary respected (verified by regeneration, not by reading); credential-input policy untouched.

</details>

<details>

Incremental diff `ef4d74e..f3dbb44` (10 files, +448/−65) read in full. Also re-read at the new head: `generator/parser/scaffold.go` complete, the `generator.go` funcmap and `scaffoldRawLiteral`, `parseSchemaDepth`, both emitters, both templates, `generator/main.go`'s generation loops, the Makefile verify targets, `.github/workflows/ci.yaml`, and the solutions doc. The PR description was re-read for the round-1 prose items.

Work done in scratch copies (the review worktree was left untouched and no git write was issued against it — `git archive` and read-only `git diff`/`gh` only):
- **Mutations, seven in total, each applied and reverted:** the two `Items` branches in `parser.go`; `propertyValue`'s untyped arm; `ScaffoldJSON`'s error return; `scaffoldRawLiteral`'s backtick guard; `hasApplyScaffold`'s gate; platform `buildScaffold`'s gate.
- **End-to-end probe:** injected a backtick into a committed spec example and ran the generator, to test finding (6)'s fix and to establish finding (1)'s exit-code behaviour.
- **Probe test:** `ParseSpec` over `specs/security/*.json`, to check that finding (4)'s new glob measures real schemas.
- `go test ./...`, `golangci-lint run` (at both heads, with cache cleans), `gofmt`/`gofumpt`, `go vet ./generator/...`, and all three `verify-*` targets in an isolated scratch clone.

Refetched from GitHub: reviews, review comments and issue comments. The only review is mine from round 1; there are no review comments and no issue comments, so no other reviewer's feedback is outstanding. The author's round-1 response is in the PR description.

Specialist lanes: `ever` was empty coming into this round, so under the re-review protocol every eligible lane should have been dispatched once here. Subagent dispatch is unavailable in this run, so I covered them inline instead — `security-reviewer`, `usability-reviewer`, `performance-reviewer`, `scope-reviewer`, `silent-failure-hunter`, `devil-advocate`, `test-quality-reviewer`. Findings are tagged by dimension, not by lane. `fidelity-reviewer` remains inapplicable: no Jira key in title, body or branch, and no linked issue.

Not verified: the PR's wire-verification claims against the `pro-nmartin` tenant (no tenant access). Unchanged from round 1.

</details>

<details>

✅ The fix for the blocking finding does the thing that was actually missing rather than the thing that would have made the finding go away: three tests now cross the parse boundary, and `TestParseSchema_ArrayPropertyItemsFromSpec` deliberately isolates one branch so a failure names its cause. The author also mutation-verified it themselves and said which mutation breaks which test — that is checkable, and it checked out.

✅ Finding (6) was solved differently rather than as prescribed, with the trade-off stated (`%q` would collapse each scaffold to one escaped line, and a `make generate` diff is the review artefact for a spec ingest). The alternative closes the risk, so the disagreement is the useful kind.

✅ The generalisation landed in the convention doc — "do not test a scaffold rule only against a hand-built `parser.Schema`" — rather than only the specific tests. That is the part that survives the next contributor.

✅ Eight fixes across three generators with zero generated-output change, proven by three verify targets rather than asserted.

</details>

PR #330, opened 2026-08-18
feat(protect): backup, restore, and the tenant overlay on Jamf-managed analytics

neilmartin83:
## Summary

Adds `protect backup` and `protect restore`, plus `protect analytics overrides` — and fixes seven export/apply bugs found by actually cloning one tenant into another.

> [!IMPORTANT]
> **Breaking output change.** `groups`, `users`, `api-clients` and `exception-sets` now export cross-resource references as **names** rather than server IDs. Anything that *parses* those four commands' output needs updating — see [Output shape changes](#output-shape-changes). Input stays backwards compatible in both directions; only the emitted shape moved.

Jamf Pro has had `backup` and `diff` for a while. Protect had neither, and no product in this repo had a restore, so cloning or rebuilding a Protect tenant meant a hand-rolled shell loop over fifteen `export` commands in the right order.

## What's new

**`protect analytics overrides`** (list/get/set/clear/export/apply). Jamf publishes analytics centrally — a stock tenant returns ~156, all `jamf: true`, and the server refuses `updateAnalytic` on any of them:

```
This mutation may only be used for custom analytics.
```

What a tenant *can* change is an overlay: the severity it reports at and the actions it triggers, written by a separate `updateInternalAnalytic` mutation. The SDK has exposed it since v0.8.0; nothing in the CLI called it. Since the definitions are identical in every tenant, that overlay is the **only** part of a Jamf-managed analytic worth capturing — and it was the part being dropped.

**`protect backup` / `protect restore`.** Backup writes each object to its own file under a per-resource directory. Restore walks it in dependency order — members before the sets naming them, sets before the plans binding them, roles before groups before users — resolving every reference by name against the target as it goes. Existing objects are updated, absent ones created; nothing is ever deleted. Both take `--resources` (allowlist) and `--exclude` (denylist), which compose; restore adds `--dry-run` and `--include-defaults`.

The `Order` field in `protectResources()` is the correctness argument, so a test asserts the dependency chain rather than the numbers, and every resource must either be restorable or carry a reason it is not.

Deliberately not replayed, each reported at runtime: Jamf-managed content, tenant defaults (built-in roles, `Default` group, `Default Analytic Set`), API clients (the server issues a new secret on create), data forwarding (its response is not its update shape and embeds a tenant-specific IAM ExternalId), and identity provider connections (no create API).

## Bugs fixed

| Bug | Cause |
|---|---|
[...]

PR #336, opened 2026-08-20
fix(pro): diff mis-keyed every compliance benchmark

ktn-jamf:
## Problem

`backup` writes a compliance benchmark's name in the `title` field, and it names the file `SlugifyName(title)`. Live mode keys the benchmark on `bm.Title`.

The root scan of the backup directory passed no name field, so `backupObjectName` found neither `name`, `displayName` nor `general.name`. It fell back to the filename stem. A disk key of `cis-level-1` then faced a live key of `CIS Level 1`.

Result: `jamf-cli pro diff --source ./backup --target production` reported every compliance benchmark as both **removed** and **added**, on every run.

This is the same class of defect as the `groupName` mis-keying in #332. That fix made the curated resources read their declared `BackupEndpoint.NameField`. Compliance benchmarks are an SDK-backed Platform resource that `backup` writes straight to the backup root, so they are not in the `BackupResources` table and the root scan had no name field to consult.

## Fix

`platformNameFields` declares the name field for the resources that `backup` writes to the backup root:

```go
var platformNameFields = map[string]string{
    "compliance-benchmarks": "title",
}
```

The root scan passes `platformNameFields[name]` where it passed `""`. A directory with no entry gets `""`, which is the current behaviour.

Blueprints need no entry. `blueprintToExport` emits a `name` field, which `backupObjectName` already finds. Blueprints were never affected.

## Scope

A diff between two directories is not affected. Both sides read through the same loader, so both used the stem and the keys agreed.

Only a directory-against-instance diff was wrong.

## Tests

Two tests, both confirmed to fail before the fix:

- `TestLoadSnapshotFromDirectory_ComplianceBenchmarkKeyedOnTitle` — asserts the benchmark is keyed `CIS Level 1`, and asserts the stem `cis-level-1` is **not** a key.
- `TestLoadSnapshotFromDirectory_EveryPlatformNameFieldIsHonoured` — the `platformNameFields` analogue of `TestLoadSnapshotFromDirectory_EveryCuratedNameFieldIsHonoured`. A Platform resource added later with a new name field fails here, not in a diff against a live tenant. The test fails if the table is emptied, so it cannot pass by asserting nothing.

Verified by mutation:

| Mutation | Result |
|---|---|
| Revert the root-scan wiring to `""` | Both tests fail |
| Delete the `compliance-benchmarks` entry | Both tests fail |

`go test ./...` passes. `go vet`, `gofmt`, `gofumpt` clean. `golangci-lint run ./internal/commands/` reports 0 issues.

## Found by

Round-2 review of #332 ([review](https://github.com/Jamf-Concepts/jamf-cli/pull/332#pullrequestreview-4984316829)), where it was held as non-blocking because it pre-dates that PR and sits outside the `NameField` mechanism.


neilmartin83 (commented):
> [!WARNING]
> ⚠️ **Needs changes** — Declares `compliance-benchmarks`' name field (`title`) for the backup-root scan, so a directory-vs-instance diff keys benchmarks the way live mode does.
> The keying half is right and well-tested. Blocking: (1) — the two sides still disagree on the *field set*, so the same diff now reports every benchmark `modified` instead of removed+added. Two nice-to-haves in the collapsed section.

## Rating: 3/5
- Would be a 5 with the benchmark projection shared between `backupBenchmarks` and the diff live loader, per (1).
- Coverage: `test-quality-reviewer`, `devil-advocate`, `silent-failure-hunter` and `usability-reviewer` have never run on this PR — the rating reflects the dimensions that were searched, not the whole diff.

### Findings

🟧 **(1)** (correctness, c=75) — `internal/commands/pro_diff.go:466-476`: benchmark objects still don't match field-for-field, so the diff remains permanently dirty

`backupBenchmarks` writes seven fields (`pro_backup.go:829-841`): the five the live loader builds, plus `rules` unconditionally, plus `sources` / `selectedOsVersions` when non-empty. The live loader builds only five. `diffObjects` (`pro_diff.go:589`) unions both key sets and flags any key present on one side, so once this PR makes the keys agree, every benchmark carrying rules lands as `modified` on field `rules` — the same "never reports clean" outcome the PR set out to remove, one change-kind over. Blueprints avoid this because both sides call one `blueprintToExport`; benchmarks have two independent projections. The asymmetry pre-dates this PR, but it is on the path the PR claims to fix, and the fix is what surfaces it.

**Failure scenario:** back up a tenant holding `CIS Level 1` with rules selected, then `pro diff --source ./backup --target production` → row `compliance-benchmarks | CIS Level 1 | modified | rules`, `old=[…] new=<nil>`, on every run with no drift on either side. Reachable from the PR as shipped.

**Suggested fix:** extract the projection `backupBenchmarks` already builds into one function beside `blueprintToExport` and call it from both sides, so a field can't be added to one projection only:

```diff
+// benchmarkToExport is the one projection of a benchmark both backup and diff
+// read, for the same reason blueprintToExport is shared: two projections drift
+// and a directory-vs-instance diff reports every benchmark modified.
+func benchmarkToExport(bm *compliancebenchmarks.BenchmarkResponseV2) map[string]any {
+	obj := map[string]any{
+		"title": bm.Title, "description": bm.Description,
+		"baselineId": bm.BaselineID, "enforcementMode": bm.EnforcementMode,
+		"target": bm.Target, "rules": bm.Rules,
+	}
+	if len(bm.Sources) > 0 { obj["sources"] = bm.Sources }
+	if len(bm.SelectedOsVersions) > 0 { obj["selectedOsVersions"] = bm.SelectedOsVersions }
+	return obj
+}
```

then in the diff live loader:

```diff
-				obj := map[string]any{
-					"title":           bm.Title,
-					"description":     bm.Description,
-					"baselineId":      bm.BaselineID,
-					"enforcementMode": bm.EnforcementMode,
-					"target":          bm.Target,
-				}
-				objects[bm.Title] = normaliseViaJSON(obj)
+				objects[bm.Title] = normaliseViaJSON(benchmarkToExport(bm))
```

**Fixed when:** both loaders derive a benchmark's object from one function, and a test asserts the disk-side and live-side objects for one benchmark have identical key sets — so `diffObjects` returns no field diffs for an unchanged benchmark. If keeping `rules` out of the diff is deliberate (it is bulky), drop it on *both* sides rather than one.

<details>

🟩 **(2)** (test-coverage, c=100) — `internal/commands/pro_diff_test.go:1157`: the guard test can't fail for the reason its comment claims

`TestLoadSnapshotFromDirectory_EveryPlatformNameFieldIsHonoured` derives the directory name from the same map key it validates, so a key that matches no directory `backup` writes — a typo, or a future entry naming the resource differently from `backupX`'s subdir literal — still passes. Its comment says "A platform resource added later with a new name field fails here", but a platform resource added later and *not* added to `platformNameFields` passes. Contrast the curated analogue at line 956, which iterates `ResolveBackupResources(nil)` — the authoritative table — on both axes. Anchor to the authoritative list instead:

```diff
 	for resource, field := range platformNameFields {
+		if !slices.Contains(nonStandardBackupFilters, resource) {
+			t.Errorf("platformNameFields key %q names no backup filter; the root scan will never see this directory", resource)
+		}
 		objName := "named-" + resource
```

Structurally cleaner (and what would make the guard a real analogue): give `nonStandardBackupFilters` (`pro_resources.go:163`) a name field per entry and have the loader read that, so there is one table rather than two lists that must agree.

**Promote to 🟧 when:** a second platform resource is added to `platformNameFields`, at which point the guard is the only thing standing between a mis-keyed entry and a silently dirty diff.

🟩 **(3)** (documentation, c=75) — `docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md`: the postmortem this PR extends still describes only the curated mechanism

That doc's `applies_when` names "Changing how `diff` loads a backup directory, or how it keys objects", it explains the root scan and `NameField` in detail (lines 76-125), and #332 shipped its fix together with an update to it. This PR adds a second name-field mechanism and a second guard test, leaving the doc saying the root scan keys directories "by their own name". Add a short section on `platformNameFields` — it is the file CLAUDE.md sends the next agent to grep.

</details>

<details>

- [x] Design and architecture: `platformNameFields` mirrors `BackupEndpoint.NameField` for resources with no table entry; the second-table concern is (2)
- [x] Correctness: traced disk keying (`backupObjectName`, nameField→`name`→`general.name`→`displayName`→stem) against both live projections; keying now agrees, field sets don't — (1). Verified the PR's blueprints claim: `blueprintExport.Name` carries `json:"name"` (`pro_blueprints.go:297,310`), and both blueprint sides share `blueprintToExport`, so blueprints are genuinely unaffected. Verified `compliance-benchmarks` is not a `BackupSubDirs()` key, so it does reach the root scan
- [na] Security: no auth, credential, transport or privilege surface in the diff
- [na] Performance: one map lookup per root directory entry
- [x] Test coverage: both new tests pass on the branch (`go test ./internal/commands/ -run TestLoadSnapshotFromDirectory`); the pinning test would fail pre-fix (empty nameField → stem key). The guard's self-reference is (2). No test covers the disk↔live field-set agreement in (1)
- [x] Reliability: root-scan read failures still fatal, per-file parse failures still warn — unchanged by this diff
- [x] Code quality: comments explain the mechanism and why blueprints need no entry; consistent with the surrounding style
- [na] Simplification: 3 changed production lines
- [na] Frontend concerns: no UI
- [x] Documentation currency: repo convention is a `docs/solutions/` postmortem for this defect class (#332 did exactly that) — (3)
- [na] Cross-repo contracts: `title` is read from `jamfplatform-go-sdk`'s `BenchmarkResponseV2.Title` (v0.16.1 `compliancebenchmarks/types.go:81`, alongside `Rules []RuleInfo` at :74), consumed only inside this repo; no wire key, enum or path is introduced
- [x] Project rules compliance: CLAUDE.md — no generated-code boundary crossed (`internal/commands/pro_diff.go` is hand-written), no credential flags, `docs/solutions/` convention is (3). No `.claude/rules/` directory in the repo
</details>

<details>

Full diff (2 files, +76/−2) at head `0f2b08e`, plus `loadSnapshotFromDirectory` / `readObjectsFromSubdir` / `backupObjectName` / `diffObjects` and the live platform loader in `pro_diff.go`, `backupBlueprints` / `backupBenchmarks` in `pro_backup.go`, `nonStandardBackupFilters` / `BackupSubDirs` in `pro_resources.go`, `blueprintExport` in `pro_blueprints.go`, the curated guard test at `pro_diff_test.go:956`, the #332 postmortem, and the SDK's `BenchmarkResponseV2` (v0.16.1). CI green (`ci` pass).

No specialists dispatched. Two reasons, both properties of the input rather than of what was found: the diff's whole production surface is one map lookup inside one function, and correctness turns entirely on three projections (`backupBenchmarks`, the diff live loader, `backupObjectName`) that were read in full here — a scoped-diff lane would re-read the same three; and this session carries a standing instruction not to spawn subagents unless asked, which also means the §12 confidence pass was self-scored under the orchestrator carve-out (every cited line is quoted above).

Never-run lanes: `test-quality-reviewer`, `devil-advocate`, `silent-failure-hunter`, `usability-reviewer` — eligible on this diff (production logic plus tests changed; a new lookup table; `if err` in the touched hunks; CLI-visible `diff` output) but skipped this round. `security-reviewer`, `performance-reviewer`, `scope-reviewer`, `fidelity-reviewer` — inapplicable: no security or hot-path file, 78 lines in one concern, and no linked spec or ticket (the PR cites a review thread on #332).

Not re-verified: the mutation table in the PR body (both tests fail with the wiring reverted / the entry deleted) was checked by reading rather than by mutating, since the review worktree is read-only.
</details>

<details>

✅ The scope analysis in the PR body is right and load-bearing: directory-vs-directory diffs read both sides through the same loader, so both used the stem and agreed — which is why this bug survived the tests that exist. Stating the unaffected case is what makes the fix reviewable.

✅ Not adding a `blueprints` entry, and saying why in the comment, is the correct call — `blueprintExport` carries `json:"name"`, so an entry would be dead configuration that reads as necessary.
</details>


<sub>🤖 Generated by the `pr-review:review` skill v1.30.0 · reviewed head `0f2b08e`</sub>

ktn-jamf:
Thanks — finding (1) was correct, and it was the more important half. Fixed in `2ace905`.

**(1) field-set disagreement — fixed.** `benchmarkToExport` is now the one projection, called from `backupBenchmarks` and from the `diff` live loader, as you suggested. `rules` stays on both sides rather than being dropped from both: it is part of what a benchmark *is*, so a backup without it could not restore one.

**(2) guard test — fixed structurally rather than with a `slices.Contains` check.** `platformNameFields` is gone. `nonStandardBackupFilters` now carries a `NameField` per entry and the root scan reads it through `nonStandardBackupNameField`, so there is one table instead of two lists that must agree. `TestLoadSnapshotFromDirectory_EveryNonStandardNameFieldIsHonoured` iterates that table on both axes.

**(3) postmortem — updated.** New section on the non-standard name field and on why keying and field set are two halves of one requirement, plus the new guards and a new `applies_when` entry.

**One thing I could not pin, stated plainly.** I mutation-tested the three fixes. Two are caught:

| Mutation | Result |
|---|---|
| Drop `NameField: "title"` from the table | 2 tests fail |
| Rename a table entry to a directory no backup writes | 2 tests fail |
| **Revert the live loader to its old 5-field projection** | **suite stays green** |

The parity test calls `benchmarkToExport` for both sides, so it proves the projection is single and correct but cannot reach the live loader's call site — that path needs a Platform SDK client. Re-inlining a literal projection there would not fail a test. I recorded this limit in the postmortem rather than leaving it implied. If you want it genuinely pinned, the seam would be extracting the list-and-project loop behind an interface, which is a larger change than this PR should carry.

`go test ./...`, `go vet`, `gofmt`, `gofumpt` clean; `golangci-lint run ./internal/commands/` 0 issues.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

neilmartin83 (approved):
> [!TIP]
> ✅ **Merge-ready** — Shares one benchmark projection between `backup` and `diff`, and folds the name field into `nonStandardBackupFilters` so there is one table rather than two lists that must agree.
> The blocking finding (1) is genuinely fixed: both sides now run the same `List` → `Get` → `benchmarkToExport` path, and a parity test asserts an unchanged benchmark produces no field diffs. Two nice-to-haves in the collapsed section, neither holding the merge.

## Rating: 5/5
- Clean — the 2 nice-to-have suggestions below are optional. Both are about what the *guard* can prove, not about whether the fix is right.
- Coverage: `test-quality-reviewer`, `devil-advocate`, `silent-failure-hunter` and `usability-reviewer` have never run on this PR — the rating reflects the dimensions that were searched, not the whole diff.

## Prior findings status

| (1) `c=75` | `pro_diff.go:466-476` | ✅ Fixed | `benchmarkToExport` (`pro_compliance_benchmarks.go:301`) is now the single projection, called from `backupBenchmarks` (`pro_backup.go:828`) and the diff live loader (`pro_diff.go:459`). Both reach it via the same `ListBenchmarks` → `GetBenchmark` path, so `rules` cannot be populated on one side only. `TestBenchmarkToExport_DiskAndLiveAgreeFieldForField` pins it through `diffObjects` |
| (2) `c=100` | `pro_diff_test.go:1157` | ⚠️ Partial | The cross-list drift is structurally gone — better than the `slices.Contains` check suggested — but the guard still derives the name field from the entry it validates, and its comment still claims a coverage it doesn't have. Carried forward as (1) below |
| (3) `c=75` | `docs/solutions/logic-errors/diff-missed-nested-backup-subdirs-2026-08-19.md` | ✅ Fixed | New "non-standard resources have their own name field" section, the keying-and-field-set framing, a new `applies_when` entry, and the new guards — including an honest note of what the parity test cannot reach |

### Findings

<details>

🟩 **(1)** (test-coverage, c=100) — `internal/commands/pro_diff_test.go:1171`: the guard still can't fail for either reason it now claims

Moving the name field into `nonStandardBackupFilters` removed the two-lists-must-agree failure mode, which was the right call. But the test still derives its fixture from the entry under test — `writeBackupFileForTest(filepath.Join(dir, entry.FilterName, ...))` with `field := entry.NameField` defaulting to `"name"` — so:

- **An entry naming a directory no backup writes still passes.** The test creates that directory itself, and the root scan reads any directory it finds. The comment (`:1155`) and the postmortem (`:154`) both claim this case "fails here"; it doesn't. What actually catches a renamed entry is `TestLoadSnapshotFromDirectory_ComplianceBenchmarkKeyedOnTitle`, which hardcodes `compliance-benchmarks` — so the author's mutation table is right about the count and wrong about which test earns it.
- **A future entry added without the `NameField` it needs still passes.** The fixture writes whatever field the entry declares, so `NameField: ""` gets a `name:` file and keys correctly. This is the original (2) unchanged, just relocated.

Closing it needs one assertion the table cannot make about itself — that the declared field is where the *writer* puts the name:

```diff
+	// The declared field is only right if it is where the writer puts the name.
+	// Nothing else in the suite couples the table to the projection.
+	nameField := nonStandardBackupNameField("compliance-benchmarks")
+	if got := benchmarkToExport(&compliancebenchmarks.BenchmarkResponseV2{Title: "x"}); got[nameField] != "x" {
+		t.Errorf("compliance-benchmarks declares NameField %q, but benchmarkToExport keys the name elsewhere: %v", nameField, got)
+	}
```

**Fixed when:** either the assertion above exists, or the two comments are corrected to claim only what the test proves. Correcting the comments is the cheaper half and worth doing regardless — the postmortem is the file CLAUDE.md sends the next agent to grep, so an overclaim there is load-bearing.

🟩 **(2)** (code-quality, c=100) — `internal/commands/pro_compliance_benchmarks.go:288`: `benchmarkToPortable`'s doc comment now documents `benchmarkToExport`

The new function landed between `benchmarkToPortable`'s comment block and its `func` line, with no blank line separating them. Lines 288-300 are now one comment run attached to `benchmarkToExport`, so `go doc` / IDE hover opens its documentation with *"benchmarkToPortable converts a benchmark API response to the portable export format"* — directly contradicting the next paragraph of the same block, which exists to say this is the snapshot shape and **not** the portable one. `benchmarkToPortable` (`:319`) is left with no doc comment at all.

```diff
-// benchmarkToPortable converts a benchmark API response to the portable export format.
-// Device group IDs are replaced with name+type for cross-instance portability.
-// Rules are converted from the detailed info format to the request format.
 // benchmarkToExport is the one projection of a benchmark that both `pro backup`
@@
 func benchmarkToExport(bm *compliancebenchmarks.BenchmarkResponseV2) map[string]any {
@@
 }
 
+// benchmarkToPortable converts a benchmark API response to the portable export format.
+// Device group IDs are replaced with name+type for cross-instance portability.
+// Rules are converted from the detailed info format to the request format.
 func benchmarkToPortable(bm *compliancebenchmarks.BenchmarkResponseV2, groupByID map[string]devicegroups.DeviceGroupListReadRepresentationV1) *benchmarkPortableInput {
```

**Fixed when:** each function's doc comment sits directly above it. `gofmt` and `golangci-lint` both pass on the current arrangement, so nothing in CI will catch this.

</details>

<details>

- [x] Design and architecture: one table (`nonStandardBackupFilter` struct) replaces the parallel-lists design (2) objected to; one projection replaces two, matching the `blueprintToExport` precedent. All four call sites of `nonStandardBackupFilters` converted to the struct field (`isKnownBackupFilter:151`, `BackupFilterNames:208`, `nonStandardBackupNameField:189`, the test) — none left iterating strings
- [x] Correctness: traced both benchmark paths to confirm the field sets can no longer diverge — `backupBenchmarks` and the diff live loader both do `ListBenchmarks` → `GetBenchmark(b.ID)` → `benchmarkToExport(bm)`, so `rules`/`sources`/`selectedOsVersions` are populated identically. Confirmed `normalizeViaJSON` (`pro_backup.go:640`) and `normaliseViaJSON` (`pro_diff.go:325`) are functionally identical marshal/unmarshal round-trips, so the disk write and the live read do not diverge on types. Confirmed `_meta` is stripped on read (`pro_diff.go:281`), so the field `backup` adds is not diffed. `nil` `Rules` round-trips to `nil` through both YAML and JSON, so an empty benchmark is also clean. Root-scan lookup returns `""` for every curated directory, so the new argument is a no-op outside the three non-standard resources
- [na] Security: no auth, credential, transport or privilege surface in the diff
- [na] Performance: one linear scan of a 3-entry slice per root directory entry
- [x] Test coverage: `go test ./internal/commands/ -run 'TestLoadSnapshotFromDirectory|TestBenchmarkToExport|TestBackupFilterNames|TestApplyProGroups'` passes on the branch. The new parity test asserts both the field-set agreement *and* that `tenantId`/`benchmarkId`/`lastUpdatedAt` stay out of the snapshot, which is more than (1) asked for. Residual guard weakness is finding (1); the author's disclosed gap (a re-inlined literal in the live loader is not caught) is accepted — see Scope
- [x] Reliability: unchanged — root read failures still fatal, per-file parse failures still warn
- [x] Code quality: the comments carry the *why* on both the struct field and the table; finding (2) is a mechanical placement slip, not a style disagreement. `gofmt` clean, `go vet` clean, CI `ci` pass
- [na] Simplification: the change is a net reduction in duplicated projections
- [na] Frontend concerns: no UI
- [x] Documentation currency: postmortem extended with the new mechanism, the keying-vs-field-set framing, and the new guards. One overclaimed sentence is folded into finding (1). CLAUDE.md needs no edit — `pro_resources.go` is already its listed entry point for backup resources
- [na] Cross-repo contracts: `Title`/`Rules`/`Sources`/`SelectedOsVersions` are read from `jamfplatform-go-sdk`'s `BenchmarkResponseV2` and consumed only inside this repo; no wire key, path or enum introduced or renamed
- [x] Project rules compliance: CLAUDE.md — no generated-code boundary crossed (all four files hand-written), no credential flags, `docs/solutions/` convention honoured. No `.claude/rules/` directory in the repo

</details>

<details>

Incremental diff since baseline `0f2b08e` is one fix commit, `2ace905`, plus a merge of `main` bringing in #330. The merge contributes nothing to this PR's own diff — `origin/main...HEAD` is still the same 6 files, +243/−37 — so review scope is `2ace905` and the whole-PR production surface it lands in: `benchmarkToExport`, `backupBenchmarks`, the diff live loader, `loadSnapshotFromDirectory` / `readObjectsFromSubdir` / `backupObjectName`, `nonStandardBackupFilters` and its four call sites, the three new tests, and the postmortem. CI green.

**Accepted gap, disclosed by the author.** `TestBenchmarkToExport_DiskAndLiveAgreeFieldForField` calls `benchmarkToExport` for both sides, so re-inlining a literal projection at `pro_diff.go:459` would not fail a test. Verified: nothing else in the suite reaches that call site, because it needs a Platform SDK client. Accepted rather than raised — pinning it means extracting the list-and-project loop behind an interface, which is a larger change than this PR should carry, and the limit is recorded in the postmortem rather than left implied.

**Not re-verified by mutation.** The author's mutation table was checked by reading the test bodies against the table, not by running mutations — the review worktree is read-only. That reading is what produced finding (1): the row "rename a table entry to a directory no backup writes → 2 tests fail" is correct on the count, but the two tests that fail are the benchmark-specific ones, not the table guard.

No specialists dispatched. Two properties of the input: the whole production surface is one shared function plus one struct-field lookup, and correctness turns on three projections read in full here, so a scoped-diff lane would re-read the same three. This session also carries a standing instruction not to spawn subagents unless asked — which likewise means the §12 confidence pass was self-scored under the orchestrator carve-out, with every cited line quoted above.

Never-run lanes: `test-quality-reviewer`, `devil-advocate`, `silent-failure-hunter`, `usability-reviewer` — eligible on this diff (production logic plus three new tests; a new lookup table; warn-and-continue paths in the touched function; CLI-visible `diff` output) but skipped for the reasons above. `security-reviewer`, `performance-reviewer`, `scope-reviewer`, `fidelity-reviewer` — inapplicable: no security or hot-path file, 243 lines in one concern, and no linked spec or ticket (the PR cites a review thread on #332).

</details>

<details>

✅ Fixing (2) structurally instead of taking the suggested `slices.Contains` check is the better call, and worth saying so explicitly: a `slices.Contains` guard would have made two lists agree, where giving `nonStandardBackupFilters` a `NameField` per entry removes the second list entirely. The suggestion was the cheap fix; this is the right one.

✅ Recording the un-pinnable mutation in the postmortem, with the reason it cannot be reached and what the seam would cost, rather than omitting it from the table or over-claiming the guard. A known gap written down is worth more than a green suite that implies there isn't one — and it is what let this round confirm the gap in one read instead of re-deriving it.

✅ Keeping `rules` on both sides rather than dropping it from both, with the reason given ("a backup without it could not restore one"). That is the correct half of the either/or (1) offered — the diff being bulky is a display concern, an unrestorable backup is a data concern.

</details>


<sub>🤖 Generated by the `pr-review:review` skill v1.30.0 · reviewed head `9e06e2d`</sub>

PR #345, opened 2026-09-03
fix(catalog): gatewaySuccessor never reached `commands -o json`, and correct the v1.28.0 changelog

neilmartin83:
Found while wire-verifying the v1.28.0 wiki update against the built binary and live gateway credentials at all three scope levels.

## The bug

`gateway.Successor` takes a command path whose **first field is the binary name** and strips it — its other two callers pass `cmd.CommandPath()`. `collectCommands`' `prefix` is deliberately binary-less, because it becomes the catalog's own `command` field. So `gatewaySuccessorOf(child, fullPath)` looked up `static-computer-groups` against a table keyed `pro static-computer-groups`. Every lookup missed:

```console
$ jamf-cli commands -o json | jq '[.[] | select(.gatewaySuccessor != null)] | length'
0        # of 1755 rows
```

CLAUDE.md describes the curated `successors` table as read by the runtime refusal, the `--help` caveat and the catalog "so the three cannot disagree". The first two were fine. The third — the one a script or an agent reads *instead of* running `--help` — was silently dead, so `pro static-computer-groups` was reported as `unserved` and nothing more, while the refusal named a working replacement.

After the fix:

```console
$ jamf-cli commands -o json | jq -r '.[] | select(.gatewaySuccessor) | .command + "  ->  " + .gatewaySuccessor' | head -1
pro static-computer-groups apply  ->  jamf-cli pro computer-groups-static-groups
```

Six rows now carry it.

## Why the test did not catch it

`TestCatalogCarriesTheSuccessorForARefusedCommand` called `collectCommands(root, "jamf-cli", "", "")`. That extra prefix field absorbed `gateway.Successor`'s binary-name strip, so the test exercised a call shape production never makes and passed against the exact bug it was written to catch — including its own `withSuccessor == 0` guard.

It now passes the empty prefix `commands -o json` uses. Verified it fails when the defect is reintroduced:

```
--- FAIL: TestCatalogCarriesTheSuccessorForARefusedCommand
    gateway_coverage_test.go:769: no refused command carries a successor; gateway.Successor is not reaching the catalog
```

## Changelog corrections

The `v1.28.0` section still said `## Unreleased`, and was wrong in three ways:

- **75 refused commands, not 48.** The undercount hid every MDM device action — 16 on `pro mobile-devices`, 8 on `pro computers-inventory` — refused because the published API declares GET on those paths but not POST. Losing `lock`, `restart`, `shutdown` and lost mode is the sharpest consequence of moving to a gateway profile, and the entry named none of it while listing `api-roles` and `policy-properties`. Now called out separately, with the note that `pro comp erase` / `remove-mdm` are hand-written and unaffected.
- **`protect backup`, `protect restore` and `protect analytics overrides` were missing entirely.** Added, including the two `0600` files that capture an HTTP report client's bearer token verbatim, that git records no non-exec permissions so a clone hands them back `0644`, and that `analytics export` omits the severity overlay.
- **The `custom-path` v2 `scope` narrowing was absent** (`[APP, FONT, PLUGIN]` → `APP`) — a silent breaking change for anyone managing font or plugin inventory paths, with nothing in `--help` to reveal it.

## Verification

`make test`, `make lint`, `make verify-generated` all clean. No generated code touched.

## Not in this PR

Five further defects found in the same pass, left alone since `main` is already tagged — happy to split them out:

1. `platform audit list` unwraps `items` and discards `nextCursor`, so `--cursor` is a flag whose input the command gives you no way to obtain. Cursor paging via `list` is impossible; `lineage` is unaffected.
2. `pro computer-groups` is registered twice (`pro/generated/registry.go:71-72`), so `jamf-cli pro --help` lists it twice and the catalog reports 1755 entries for 1753 paths. Two specs collapse to one resource name and the generator has only an *operation*-name collision check.
3. `checkScopeConflict`'s three detailed messages are unreachable — `resolveAuth` refuses the pair first with a terser one, which names `--environment-id`/`--tenant-id` even when the conflict came from the env vars.
4. `auditScopeNote` is gated on `REQUEST_CONTEXT_NOT_PROVIDED`, but an organization credential on audit actually answers `INVALID_REQUEST_CONTEXT_TYPE … 'tenant' is invalid. Expected any of 'environment'` — wire-confirmed, so the note is likely unreachable.
5. MCP's `blockedChildFlags` blocks `--tenant-id` but not `--environment-id`, so a model can redirect the scope on an environment-scoped launch profile.

Also: CLAUDE.md states Platform/Security Cloud `--file` accepts piped stdin (it does not — `os.Stdin` appears only in `confirm.go`) and that DNS has two whole-list replaces (only `PUT /v1/dns/custom-hostname-mappings`).


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Fixes `gatewaySuccessorOf` so the JSON catalog carries the curated
> successor, and corrects the v1.28.0 changelog's refused-command count.
> Blocking: (1). See the collapsed section for coverage details.

## Rating: 3/5
- Would be a 5 once `docs/guides/platform-api-ga.md`'s refused-command table matches the count this PR just corrected in `CHANGELOG.md`.

The code fix is correct and well-verified. I rebuilt the binary at both `origin/main` and
the PR head: `commands -o json` carries 0 successors before the fix and 6 after, exactly
matching the PR body's own before/after claim, and the updated test
(`TestCatalogCarriesTheSuccessorForARefusedCommand`) now exercises the same empty-prefix
call shape `commands -o json` actually uses. `go build ./...` and
`go test ./internal/commands/...` both pass clean.

⚠️ **IMPORTANT** (1) (documentation, docs/guides/platform-api-ga.md:227-247) — the migration guide's refused-command table still says 48 and omits every MDM device action

This PR corrects `CHANGELOG.md`'s refused-command count from 48 to 75, and adds the missing
24-command MDM device-action category. `docs/guides/platform-api-ga.md` — the user-facing
migration guide CLAUDE.md names as authoritative for this exact table — still says
"48 commands in total" and has no `pro mobile-devices` or `pro computers-inventory` row at
all. I rebuilt the PR-head binary and confirmed the live count: 75 unserved commands, 16
under `pro mobile-devices` and 8 under `pro computers-inventory`. The guide's own listed
rows also undercount two entries it does carry: `pro api-roles` is 6, not 5, and
`pro static-computer-groups` is 6, not 5. The guide now contradicts the changelog entry
this PR ships in the same commit, on the exact defect class the PR exists to fix.

**Failure scenario:** an operator reads the corrected `v1.28.0` changelog entry (75
commands, MDM lockout called out), then opens the linked migration guide for the
authoritative table and sees 48 with no MDM row — under-preparing for the lock/restart/
shutdown loss CLAUDE.md calls "the sharpest consequence of moving to a gateway profile."

**Suggested fix:**
```diff
 | `pro api-integrations` | 6 | outside the published API — withdrawn to close a privilege-escalation path |
-| `pro api-roles` | 5 | as above |
+| `pro api-roles` | 6 | as above |
 | `pro api-roles-privileges` | 2 | as above |
 | `pro authentications` | 6 | outside the published API |
 | `pro oauth-token-sessions` | 1 | outside the published API |
 | `pro environment-type` | 1 | outside the published API |
 | `pro database-connections` | 1 | outside the published API |
 | `pro systems` | 2 | `initialize` / `platform-initialize`, withdrawn upstream |
 | `pro mac-os-managed-software-updates` | 1 | `list` (the deprecated available-updates endpoint) |
 | `pro mdm-commands commands` | 1 | the gateway declares GET on that path, not POST |
 | `pro classic-computer-configs` | 7 | outside the published Classic API 11.28.0 |
-| `pro static-computer-groups` | 5 | the deprecated v2 endpoint — use `pro computer-groups-static-groups` |
+| `pro static-computer-groups` | 6 | the deprecated v2 endpoint — use `pro computer-groups-static-groups` |
 | `pro classic-patch-reports` | 2 | withdrawn from the published Classic API 11.28.0 |
 | `pro classic-patch-titles` | 5 | `list`, `get`, `update`, `delete`, `apply` — withdrawn; `create` still works |
 | `pro classic-patch-policies` | 1 | `list` — withdrawn; `get`, `create`, `update`, `delete` still work |
 | `pro policy-properties` | 2 | `GET`/`PUT /settings/obj/policyProperties`, withdrawn from the published API |
+| `pro mobile-devices` | 16 | the gateway declares GET on those paths, not POST — loses `lock`/`restart`/`shutdown` and lost-mode |
+| `pro computers-inventory` | 8 | as above — loses `lock`/`restart`/`shutdown` and remote-desktop control |

-48 commands in total (a wholly-refused resource contributes its group node too).
+75 commands in total (a wholly-refused resource contributes its group node too).
```

**Fixed when:** the table's row counts and total match
`jamf-cli commands -o json | jq '[.[] | select(.gateway=="unserved")] | length'` (75), and
`pro mobile-devices` / `pro computers-inventory` each have a row.

<details>

- [x] Design and architecture: single-argument signature fix in `gatewaySuccessorOf`, no design concern
- [x] Correctness: reproduced 0→6 successors before/after against the actual production call site (`root.go:1038`)
- [x] Test coverage: `TestCatalogCarriesTheSuccessorForARefusedCommand` now matches production's empty-prefix call, confirmed it fails when the bug is reintroduced
- [x] Code quality/simplification: root-cause comment added on `gatewaySuccessorOf`, no cleanup needed
- [x] Documentation currency: `CHANGELOG.md` count verified accurate (75); `docs/guides/platform-api-ga.md` found stale, see (1)
- [x] Project rules compliance: CLAUDE.md's changelog/doc-currency convention checked against both files
- [na] Security, performance, reliability, cross-repo contracts, frontend.

Diff: 4 files, +55/−11, head `2c933d5`. Related file read: `docs/guides/platform-api-ga.md`
(outside the diff, for the documentation-currency check). Lanes run: none — risk was
medium by line-count only, with no auth/security/API-surface files touched, and the fix
was independently verified by rebuilding both revisions. Never run: `devil-advocate`,
`fidelity-reviewer` (no Jira key found), `performance-reviewer`, `scope-reviewer`,
`security-reviewer`, `silent-failure-hunter`, `test-quality-reviewer`,
`usability-reviewer` — inapplicable to this diff's size and shape.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `internal/commands/gateway_coverage.go` | 5/5 | `gatewaySuccessorOf` now takes `cmd.CommandPath()`, matching `gateway.Successor`'s binary-name strip; verified by build reproduction |
| `internal/commands/gateway_coverage_test.go` | 5/5 | Test call shape now matches production's empty-prefix `collectCommands` invocation; confirmed it catches the regression |
| `internal/commands/root.go` | 5/5 | One-line call-site update for the new signature |

</details>

neilmartin83:
@ktn-jamf addressed — finding (1) is fixed in a6102acc, and the guide's table now matches the binary.

Your row counts were right on `pro api-roles` (6) and `pro static-computer-groups` (6), and the two missing rows are 16 and 8 as you measured. One more the suggested diff kept at its stale value: **`pro api-integrations` is 7, not 6** — `client-credentials` is refused alongside the five CRUD verbs plus `apply`. With that, the rows sum to exactly 75:

```console
$ jamf-cli commands -o json --no-hints | jq '[.[] | select(.gateway=="unserved")] | length'
75
$ grep -E '^\| `pro ' docs/guides/platform-api-ga.md | head -18 | awk -F'|' '{gsub(/ /,"",$3); s+=$3} END {print s}'
75
```

The two MDM rows went at the **top** of the table rather than the bottom, so the "last three rows" sentence below it still points at the partial-group examples it was written for — that sentence now names both sets.

Beyond the table, the MDM refusal gets its own paragraph matching the changelog's, because a row of "16" does not tell an operator what they lose. It names all 24 subcommands, says the refusal is **per method** (the published API declares GET on those paths but not POST, so the inventory reads on both resources are unaffected), and records that `pro comp erase` / `pro comp remove-mdm` are hand-written against a different path and are **not** refused — verified in the catalog, both carry no gateway verdict.

Also bumped the snapshot note's SDK ref from `v0.20.1` to `v0.21.0`, the ingest this tree is actually on (`specs/gateway/coverage.json` records `sdkCommit fc34c1d`). That paragraph is what governs the table's snapshot claim, so leaving it a version behind while correcting the table would have been the same defect one line up. The surfaces it quotes are unchanged — Jamf Pro API 11.31.0, Classic API 11.28.0 — so no other number in the guide moved.

`main` is merged in (0c4a5ecc) and re-verified on top of it: 75 unserved and 6 rows carrying `gatewaySuccessor`, with `make test`, `make lint` and `make verify-generated` all clean.

The five defects listed under "Not in this PR" are still unfixed and still out of scope here.

ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Fixes `gatewaySuccessorOf` so the JSON catalog carries the curated
> successor, and corrects the v1.28.0 changelog and migration guide.
> Blocking: (1), (2). See the collapsed section for 1 nice-to-have suggestion.

## Rating: 3/5
- Would be a 5 once `CHANGELOG.md`'s own "the rest are" enumeration adds up to 75 (1), and a test asserts `gatewaySuccessor` survives into the actual JSON map key (2).

The round-1 finding is fixed and verified against the built binary. Two new findings
surfaced this round from two specialist lanes that had never run on this PR before —
dispatched now per the re-review catch-up rule, not because the fix commits themselves
looked risky.

## Prior findings status

|---|---|---|---|
| (1) | `docs/guides/platform-api-ga.md:227-247` | ✅ Fixed | `a6102acc8`: added the missing `pro mobile-devices` (16) and `pro computers-inventory` (8) rows, corrected two more undercounts I had missed (`api-integrations` 6→7, `api-roles` 5→6) alongside the one I flagged (`static-computer-groups` 5→6). All 18 rows now sum to 75, verified against `jamf-cli commands -o json` on the rebuilt binary. |

⚠️ **IMPORTANT** (1) (documentation, via usability-reviewer) — `CHANGELOG.md:36-53`: the changelog's own refused-command enumeration still does not add up to 75

Round 1 fixed the migration guide; `CHANGELOG.md`'s v1.28.0 entry — the file this PR opened specifically to correct — has the identical defect and was never touched by the fix commit. It claims 75 total, attributes 24 to MDM device actions, then lists "the rest" by name. Summing those named groups' counts (from the now-correct guide table) gives 50, not the 51 needed (75−24). `pro environment-type` (count 1) is absent from the sentence.

**Failure scenario:** an operator reads `CHANGELOG.md` alone — the file, not the guide — to get the full refused-command list, tries to reconcile the named groups against "75 commands in total," gets 74, and has no way to learn from this file that `pro environment-type` is also refused.

**Suggested fix:**
```diff
 The rest are `pro api-roles`, `pro api-integrations`, `pro api-roles-privileges`,
-`pro authentications`, `pro oauth-token-sessions`, `pro classic-computer-configs`,
+`pro authentications`, `pro oauth-token-sessions`, `pro environment-type`,
+`pro classic-computer-configs`,
 `pro static-computer-groups` (use `pro computer-groups-static-groups`),
 `pro classic-patch-reports`, five `pro classic-patch-titles` subcommands,
 `pro classic-patch-policies list`, `pro policy-properties`, `pro systems`,
 `pro mdm-commands commands`, `pro mac-os-managed-software-updates` and
 `pro database-connections`.
```

**Fixed when:** the groups named in `CHANGELOG.md`'s sentence, summed against the guide's counts, add up to 75.

---

⚠️ **IMPORTANT** (2) (test-coverage, via test-quality-reviewer) — `internal/commands/gateway_coverage_test.go:730-771`: the new test verifies the struct field, not the JSON key the PR title is about

`TestCatalogCarriesTheSuccessorForARefusedCommand` reads `commandEntry.GatewaySuccessor` straight off `collectCommands`'s return value. It never calls `commandEntriesToMaps` (`root.go:1147-1216`), which is the function that copies `GatewaySuccessor` into the `m["gatewaySuccessor"]` key actually marshaled by `commands -o json` — and `TestCommandsSubcommand_JSON` runs that real command but asserts nothing about its stdout. Verified by mutation: deleting `commandEntriesToMaps`'s `if e.GatewaySuccessor != "" { m["gatewaySuccessor"] = e.GatewaySuccessor }` block and running `go test ./internal/commands/...` passes clean, 0 failures — reproducing, one function later, the exact class of defect ("computed correctly, never reaches the JSON catalog") this PR exists to fix.

**Failure scenario:** a future refactor of `commandEntriesToMaps`'s positive-only field block (three other fields sit in the same `if` guard) drops or renames the `gatewaySuccessor` copy → `commands -o json` silently loses the field again → the entire suite stays green.

**Suggested fix:**
```diff
+func TestCommandEntriesToMaps_GatewaySuccessor(t *testing.T) {
+	entries := []commandEntry{{Command: "x", GatewaySuccessor: "jamf-cli y"}, {Command: "z"}}
+	maps := commandEntriesToMaps(entries)
+	if maps[0]["gatewaySuccessor"] != "jamf-cli y" {
+		t.Errorf("gatewaySuccessor missing from JSON map for an entry that has one")
+	}
+	if _, ok := maps[1]["gatewaySuccessor"]; ok {
+		t.Errorf("gatewaySuccessor present for an entry with none")
+	}
+}
```

**Fixed when:** a test fails when the `gatewaySuccessor` copy is removed from `commandEntriesToMaps`, without inspecting the internal `commandEntry` struct directly.

---

<details>

💡 **NICE-TO-HAVE** (3) (documentation, via usability-reviewer) — `CHANGELOG.md`: no entry says `gatewaySuccessor` is now populated

Before this PR every one of the 1,755 catalog rows always emitted an empty `gatewaySuccessor` — a script could not have relied on it. This PR makes it populate for the first time, a genuine script-visible change to a documented field, with no changelog bullet saying so.

**Fixed when:** a short bullet near the existing catalog-fields note says `gatewaySuccessor` now reports a value where it never did before.

</details>

<details>

- [x] Correctness: `gatewaySuccessorOf`/`cmd.CommandPath()` fix, single call site, reproduced 0→6 successors — unchanged since round 1
- [x] Documentation currency: guide table fixed (1); `CHANGELOG.md` gap found this round, see (1) and (3)
- [x] Test coverage: mutation-verified gap between the struct field and the JSON map key, see (2)
- [x] Design and architecture, security, reliability, code quality: no new concern from `usability-reviewer`, `test-quality-reviewer`, or `silent-failure-hunter`
- [na] Performance, cross-repo contracts, frontend.

Diff: 5 files, +76/−18 vs `main`, head `a183a0a8`. Lanes run this round (never run before on
this PR, dispatched per the re-review catch-up rule): `usability-reviewer`,
`test-quality-reviewer`, `silent-failure-hunter` (no findings). Never run: `security-reviewer`
(no auth/secrets/config/deps files), `performance-reviewer` (no DB/loop/cache files),
`scope-reviewer` (single-concern, well under the size threshold), `devil-advocate` (trivial
one-liner, single call site), `fidelity-reviewer` (no linked issue or Jira ticket).

</details>



neilmartin83:
@ktn-jamf both blocking findings fixed in 16c3704c.

**(1) — `CHANGELOG.md` enumeration.** You were right that it summed to 50, not 51, and that `pro environment-type` was the missing one. Rather than just inserting the name, the sentence now carries each group's count, so the arithmetic is checkable in the file itself instead of only against the guide's table:

```console
$ bin/jamf-cli commands -o json --no-hints | jq -r '.[] | select(.gateway=="unserved") | .command' \
    | grep -vE '^pro (mobile-devices|computers-inventory) ' | awk '{print $1" "$2}' | sort | uniq -c \
    | awk '{s+=$1} END {print s}'
51
```

51 + 24 = 75. The 24 MDM subcommand names already in that bullet were checked name-for-name against the catalog too — all 24 match.

**(2) — the test asserted the struct field, not the JSON key.** Reproduced your mutation before writing anything: with `commandEntriesToMaps`' `m["gatewaySuccessor"]` copy deleted, `go test ./internal/commands/...` is green, 0 failures. `TestCatalogJSONCarriesTheSuccessorKey` runs the real `collectCommands` output through `commandEntriesToMaps(entries, true)` and asserts on the **map keys only**, so it cannot pass by re-reading the value the copy was supposed to move. Confirmed it fails with the copy still removed:

```
--- FAIL: TestCatalogJSONCarriesTheSuccessorKey
    gateway_coverage_test.go:817: no gatewaySuccessor key in the JSON catalog; the field is computed but not marshaled
```

Two notes on the suggested diff. `commandEntriesToMaps` takes `(entries, full bool)` and the `gatewaySuccessor` copy is nested inside the `if e.Gateway != ""` guard, so the snippet as written would not compile and its first entry would not have carried the key even if it did. And the test pins the positive-only contract in **both** directions — absent for a served command, never present-but-empty — because that guard is the reason the field is nested there at all.

**(3), the nice-to-have** — taken. There is now a bullet beside the existing `privileges`/`gatewayPrivileges`/`gatewayPermissions` note saying `gatewaySuccessor` reports a value where it was emitted on no row at all.

`make test`, `make lint` and `make verify-generated` clean; 75 unserved and 6 rows carrying `gatewaySuccessor` on the rebuilt binary, and the guide's 18 rows still sum to 75.

One thing worth flagging for whoever merges: the 75 is about to become **67**. The SDK v0.22.0 ingest on `chore/sdk-ingest-2026-09-05` restores the Classic patch family (`public-apis-oas#438`), which un-refuses the eight `classic-patch-*` commands. Both counts here are correct for `main` as it stands, and both files will need the same pass again when that lands — which is the argument for the count-per-group form the sentence now uses, since the eight are three named rows rather than a number to re-derive.

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready**. Fixes `gatewaySuccessorOf` so the JSON catalog carries the curated
> successor, and corrects the v1.28.0 changelog and migration guide. No open findings.

## Rating: 5/5
- Both round-2 findings are fixed and independently re-verified, including reproducing the mutation myself. Clean.

## Prior findings status

|---|---|---|---|
| (1) | `CHANGELOG.md:36-53` | ✅ Fixed | `16c3704c4`: the sentence now carries each group's count (51 non-MDM, including `pro environment-type`) rather than only names, so the arithmetic is checkable in the file itself. Re-summed by hand against the rebuilt binary's per-group counts: 51 + 24 = 75, exact match on every group. |
| (2) | `internal/commands/gateway_coverage_test.go:730-771` | ✅ Fixed | `16c3704c4`: new `TestCatalogJSONCarriesTheSuccessorKey` runs `commandEntriesToMaps(entries, true)` — the real `commands -o json` path — and asserts on the map keys, not the struct field. I independently reproduced the mutation: removing the `gatewaySuccessor` copy inside `commandEntriesToMaps`'s positive-only block fails this test with the exact message reported, and the rest of the suite stays green either way — confirming the new test is what actually closes the gap. |
| (3) | *(nice-to-have)* | ✅ Fixed | A `commands -o json`'s `gatewaySuccessor` reports a value bullet was added beside the existing `privileges`/`gatewayPrivileges`/`gatewayPermissions` note. |

No active findings. `go build ./...`, `go test ./internal/commands/...`, `gofmt`,
`golangci-lint run` (`0 issues.`), and `make verify-generated` are all clean on the rebuilt
binary. The guide's table and the changelog's enumeration both independently sum to 75, and
the built binary's `commands -o json` reports 75 unserved / 6 with `gatewaySuccessor`,
matching all three.

<details>

- [x] Correctness: unchanged since round 1, still verified (0→6 successors, single call site)
- [x] Documentation currency: both files now sum correctly against the live coverage manifest, verified against the rebuilt binary
- [x] Test coverage: the struct-vs-JSON-key gap from round 2 is closed and mutation-verified by me directly, not just by the reported claim
- [x] Design and architecture, security, reliability, code quality: no new production code in this round — CHANGELOG.md prose and one new test function only
- [na] Performance, cross-repo contracts, frontend.

Diff since round 2: 2 files, +60/−7, head `16c3704c`. No specialist re-dispatch this round —
the fix touches no production code, so none of the three already-run lanes
(`usability-reviewer`, `test-quality-reviewer`, `silent-failure-hunter`) meet the re-run bar,
and the never-run lanes' eligibility is unchanged from round 2 (`security-reviewer`,
`performance-reviewer`: no matching files; `scope-reviewer`, `devil-advocate`: still a
single-concern, trivial-shape diff; `fidelity-reviewer`: no linked issue or Jira ticket).

</details>

PR #348, opened 2026-09-03
feat(pro report): add --bundle-id and --path to software-installs

### LINKED ISSUE #325: Add flag to include app bundle id to jamf-cli pro report software-installs
### Product namespace

pro (Jamf Pro / Platform API)

### What problem does this solve?

Be useful to show the app bundle id in addition to app name in `jamf-cli pro report software-installs`

### Proposed solution

Add `--bundle_id` flag to `jamf-cli pro report software-installs`

### Alternatives considered

_No response_

ktn-jamf:
Closes #325.

`pro report software-installs` reported three columns: `title`, `version`, `device_count`. This adds `--bundle-id` and `--path`, which report an application's bundle identifier and its install path.

Both fields were already on the wire. The `ComputerApplication` schema declares `bundleId` and `path`, and the function already read `path` to feed `isSystemApp`. No spec change and no regeneration are needed.

## Each flag extends the grouping key

This is the correctness point of the change. One title and version can map to several bundle identifiers or several install paths — a repackaged application, or the same application installed in two locations. A display-only column would print one arbitrary member of the group and hide the rest, with nothing in the output to reveal it. So each flag joins the aggregation key:

```
$ jamf-cli pro report software-installs --title Zoom --bundle-id -o table
 BUNDLE_ID                DEVICE_COUNT  TITLE  VERSION
 com.example.zoom-repack  1             Zoom   5.0
 us.zoom.xos              2             Zoom   5.0
```

Without the flag those three installs are one row reading `3` for `Zoom 5.0`.

Column order is alphabetical. The table formatter reads its column set from `sortedKeys(rows[0])`, which floats only `id` and `name`, so every table in this CLI orders the rest that way. Use `--select title,version,bundle_id,device_count` for a different order.

`softwareKey` gains `bundleID` and `path`, left at the zero value when the flag is off, so the counting loop keys itself correctly in all four flag combinations with no branching. Sorting gained tie-breakers on bundle identifier then path, because a title and version can now span rows.

## The default output is unchanged

The row builder gates each optional key on the flag, never on whether the value is empty. A table's columns are the keys of its first row, so an emptiness gate would make the column set depend on sort order, and an application with no `bundleId` on the wire would drop the column for every row. Both flags are off by default, so anything parsing the current three columns is unaffected. A test asserts that the default rows carry exactly those three keys and no more.

## Flag names

Kebab-case, matching `--include-system` and every other flag in this CLI. The issue text asks for `--bundle_id`; snake_case would be the only such flag in the binary. `--path` covers the `--file_path` request in the issue comment.

## Tests

Four cases beside the existing four:

- `TestRunReportSoftwareInstalls_BundleIDExtendsTheGroupingKey`
- `TestRunReportSoftwareInstalls_PathExtendsTheGroupingKey`
- `TestRunReportSoftwareInstalls_DefaultRowsCarryOnlyTheThreeOriginalKeys`
- `TestRunReportSoftwareInstalls_BundleIDAndPathTogether`

The first two are the ones that fail if the grouping is ever reduced to a display column. That was confirmed by mutation: discarding the two key assignments makes exactly the grouping tests fail, and no others.

`make test` passes in full with zero failures. `go vet ./internal/commands/` is clean. `make verify-site` and `make verify-site-output` both pass, since the change touches `docs/site`.

## Second commit, from review

A deep review found one confirmed defect and two real test gaps.

Both flags are boolean and both names invite a value, so `--path /Applications/Foo.app` set the flag and left the path as a positional argument, which the command discarded in silence and then reported the whole fleet. The operator got fleet-wide output for what they had typed as a filter. `Args: cobra.NoArgs` refuses it: cobra reports it as `unknown command`, which `ClassifyError` already maps to exit 2, and `ValidateArgs` runs ahead of `PersistentPreRunE`, so the refusal lands before any credential resolves or any request is sent.

The hole itself is repo-wide and pre-existing — `pro report inventory-summary junkarg` and `pro categories list junkarg` swallow a positional the same way — so the fix is scoped to the one command whose new flag names make a latent hole reachable. The general case deserves its own change.

The two test gaps: nothing in this file asserted row order at all, so the new tie-breakers were unobservable, and neither empty-value case was covered. `TestRunReportSoftwareInstalls_RowOrderFollowsEveryComparatorLevel` puts a tie at each of the four comparator levels and pins the full sequence; flipping any level in turn fails it. Two further tests cover an application the wire reports with no `bundleId` and one with no `path`, both of which still get a row carrying the column, because the row builder gates on the flag rather than on the value. Mutating either gate to test the value fails them.

The two flags also joined `docs/site/examples.json`. No test validates that file, so it was checked by parsing it, comparing its key set against every other entry, and running both documented commands.

The command needs a live Jamf Pro tenant, so the fixture-client tests are the runtime evidence for the aggregation. The rendered `--help` was checked on the built binary.

`make lint` was not run. The `golangci-lint` on the development machine is built with go1.26.2 while `go.mod` targets 1.27.0, so it aborts at config load before it reads any file. CI installs its own version and is unaffected.

### DIFF
--- a/docs/site/examples.json
+++ b/docs/site/examples.json
@@ -271,6 +271,14 @@
     {
       "description": "Include system apps",
       "command": "jamf-cli pro report software-installs --include-system"
+    },
+    {
+      "description": "Report a row per bundle ID",
+      "command": "jamf-cli pro report software-installs --bundle-id"
+    },
+    {
+      "description": "Report a row per install path",
+      "command": "jamf-cli pro report software-installs --path"
     }
   ],
   "pro report update-status": [
--- a/internal/commands/pro_report_software.go
+++ b/internal/commands/pro_report_software.go
@@ -39,6 +39,10 @@ merged row. Neither flag is on by default, so the default output is unchanged.
 
 Output columns: title, version, device_count (plus bundle_id with --bundle-id,
 path with --path)`,
+		// --bundle-id and --path are boolean, so their names invite a value that
+		// cobra would leave as a positional. Without this the whole report runs
+		// as if the value had never been typed.
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			rows, err := runReportSoftwareInstalls(cmd.Context(), cliCtx.Client, titleFilter, includeSystem, showBundleID, showPath)
 			if err != nil {
--- a/internal/commands/pro_report_software.go
+++ b/internal/commands/pro_report_software.go
@@ -18,6 +18,8 @@ func newReportSoftwareInstallsCmd(cliCtx *registry.CLIContext) *cobra.Command {
 	var (
 		titleFilter   string
 		includeSystem bool
+		showBundleID  bool
+		showPath      bool
 	)
 
 	cmd := &cobra.Command{
@@ -30,9 +32,15 @@ Use --include-system to show all applications.
 
 Use --title to filter to a specific application name (case-insensitive substring match).
 
-Output columns: title, version, device_count`,
+Use --bundle-id to add a bundle_id column, and --path to add a path column.
+Each also becomes part of the grouping, so one title and version that spans
+several bundle IDs or install paths reports a row for each rather than a single
+merged row. Neither flag is on by default, so the default output is unchanged.
+
+Output columns: title, version, device_count (plus bundle_id with --bundle-id,
+path with --path)`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			rows, err := runReportSoftwareInstalls(cmd.Context(), cliCtx.Client, titleFilter, includeSystem)
+			rows, err := runReportSoftwareInstalls(cmd.Context(), cliCtx.Client, titleFilter, includeSystem, showBundleID, showPath)
 			if err != nil {
 				return err
 			}
@@ -43,6 +51,8 @@ Output columns: title, version, device_count`,
 
 	cmd.Flags().StringVar(&titleFilter, "title", "", "filter to application names containing this substring (case-insensitive)")
 	cmd.Flags().BoolVar(&includeSystem, "include-system", false, "include system apps from /System/ and /Library/")
+	cmd.Flags().BoolVar(&showBundleID, "bundle-id", false, "add a bundle_id column and group by bundle ID")
+	cmd.Flags().BoolVar(&showPath, "path", false, "add a path column and group by install path")
 
 	return cmd
 }
@@ -57,14 +67,20 @@ func isSystemApp(path string) bool {
 }
 
 // softwareKey groups installs by application name + version.
+//
+// --bundle-id and --path extend this key rather than adding a display column:
+// one title + version can span several bundle IDs or install paths, and a
+// column would print one arbitrary member and silently hide the rest.
 type softwareKey struct {
-	title   string
-	version string
+	title    string
+	version  string
+	bundleID string
+	path     string
 }
 
 // runReportSoftwareInstalls fetches computer inventory with the APPLICATIONS
-// section and aggregates device counts per (title, version).
-func runReportSoftwareInstalls(ctx context.Context, client registry.HTTPClient, titleFilter string, includeSystem bool) ([]map[string]any, error) {
+// section and aggregates device counts per softwareKey.
+func runReportSoftwareInstalls(ctx context.Context, client registry.HTTPClient, titleFilter string, includeSystem, showBundleID, showPath bool) ([]map[string]any, error) {
 	computers, err := FetchAllPaginated(ctx, client, "/v4/computers-inventory?section=APPLICATIONS", 100)
 	if err != nil {
 		return nil, fmt.Errorf("fetching computer inventory: %w", err)
@@ -83,6 +99,7 @@ func runReportSoftwareInstalls(ctx context.Context, client registry.HTTPClient, 
 			name, _ := app["name"].(string)
 			version, _ := app["version"].(string)
 			path, _ := app["path"].(string)
+			bundleID, _ := app["bundleId"].(string)
 
 			if name == "" {
 				continue
@@ -94,7 +111,14 @@ func runReportSoftwareInstalls(ctx context.Context, client registry.HTTPClient, 
 				continue
 			}
 
-			counts[softwareKey{name, version}]++
+			key := softwareKey{title: name, version: version}
+			if showBundleID {
+				key.bundleID = bundleID
+			}
+			if showPath {
+				key.path = path
+			}
+			counts[key]++
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Adds opt-in `--bundle-id` and `--path` to `pro report software-installs`, each extending the aggregation key rather than adding a display column.
> Blocking: (1). See the collapsed sections for 1 nice-to-have and 4 tentative findings.

## Rating: 3/5
- Would be a 5 once a value typed after either new flag stops being silently discarded (1).

The core design decision is right, and the PR body's reasoning for it holds up under checking. Extending the grouping key rather than adding a display column is correct, and gating the row keys on the flag rather than on emptiness is the only version that keeps a stable column set. See "What's done well" for the parts I verified positively.

---

⚠️ **IMPORTANT** (1) (usability, via usability-reviewer+devil-advocate) — `internal/commands/pro_report_software.go:54-55`: a value typed after `--bundle-id` or `--path` is silently discarded

Both flags are `BoolVar`, but each names a concept that has a value. The same command declares `--title` as a value-taking filter on line 52. A user primed by `--title "Chrome"` types `--path /Applications/Slack.app` expecting a path filter. The command sets no `Args` validator, so cobra binds the boolean and leaves the value as an unconsumed positional, which is then dropped. Accepting a stray positional is pre-existing. What is new is two flag names that invite one. Flag names are a released contract, so this is the cheapest moment to reconsider them.

**Failure scenario:** an operator runs `pro report software-installs --path /Applications/Slack.app` expecting a filtered report → gets the full unfiltered fleet grouped by install path, exit 0, with nothing saying the value was ignored. Verified on the built binary: the invocation parses cleanly and reaches the auth check.

**Suggested fix:**
```diff
 	cmd := &cobra.Command{
 		Use:   "software-installs",
 		Short: "Installed software version distribution",
+		Args:  cobra.NoArgs,
```

`cobra.NoArgs` turns the silent discard into an error, and it covers every future flag on this command rather than these two. Renaming `--path` to `--install-path` is a reasonable additional step, but it does not close the hole on its own.

**Fixed when:** `pro report software-installs --path /some/value` exits non-zero naming the unexpected argument, and a test asserts that.

<details>

💡 **NICE-TO-HAVE** (2) (test-quality, via test-quality-reviewer) — `internal/commands/pro_report_software.go:137-140`: the two new sort tie-breakers have no test that can observe them

I flipped both comparisons (`<` to `>`) in an isolated sandbox and all eight `TestRunReportSoftwareInstalls_*` tests still passed. The new tests key a map by `bundle_id` or `path` before asserting, which is order-independent by construction. Flipping the pre-existing `title` comparator on line 132 also survives, so row order has never been asserted in this file. The PR adds two untested branches to an already-untested comparator.

**Fixed when:** one two-flags-on test asserts `rows[0]` and `rows[1]` positionally, so flipping either tie-breaker fails.

Promote to IMPORTANT when: the command's help text or docs promise a specific row order.

</details>

<details>
<summary>Tentative findings (4 items)</summary>

These sit below the main confidence cutoff. Each is real as described, and each is either pre-existing or resting on something I could not close.

💡 **NICE-TO-HAVE** (3) (correctness) `c=50` — `internal/commands/pro_report_software.go:121`: `device_count` counts application entries, not distinct devices

`counts[key]++` runs once per application entry per computer. I probed a fixture of one device reporting two variants of the same title and version at different paths. The summed `device_count` is 2 in all four flag combinations, including the default. So the overcount is pre-existing and this PR is exactly neutral to it. What changes is discoverability: with `--path` the one inflated row splits into two rows of 1, which reveals the cause. A line in the `Long` text saying `device_count` is an install count would help.

`unresolved:` whether the column name is deliberate, since it predates this PR and no comment explains it.

💡 **NICE-TO-HAVE** (4) (performance, via performance-reviewer) `c=50` — `internal/commands/pro_report_software.go:118-120`: `--path` removes the bound on row count, and the help text does not warn

The key `{title, version}` bounds rows by the software catalogue, which does not grow with fleet size. Folding the raw install path into the key makes rows grow per device for any app under a user-scoped location, because `isSystemApp` excludes `/System/`, `/Library/`, `/usr/`, `/bin/` and `/Applications/Utilities/`, but not `/Users/`. There is no `--limit` on this command. A note on the flag pointing at `--title` would cover it.

`unresolved:` the real ratio of user-scoped installs on a large fleet, which needs a live tenant. Two adjacent claims I checked and rejected: table is not this command's default format, the root default is `-o json`, so the 50-row narrowing hint does fire; and `FetchAllPaginated` already holds the entire fleet inventory in one slice before aggregation, so the row set is strictly smaller than data already resident and there is no new peak-memory class.

💡 **NICE-TO-HAVE** (5) (documentation) `c=50` — `docs/site/examples.json:262-275`: the new flags are absent from this command's example list, which already covers its other two

That block carries three examples: the bare command, `--title "Google Chrome"`, and `--include-system`. Both pre-existing optional flags are represented, and neither new flag is. Other report commands in the same file demonstrate their optional flags too.

`unresolved:` whether updating that file is enforced anywhere. CLAUDE.md describes the site as introspecting the built binary for `commands.json` and does not name `examples.json`, so this rests on the file's evident convention alone.

💡 **NICE-TO-HAVE** (6) (silent-failure, via silent-failure-hunter) `c=50` — `internal/commands/pro_report_software.go:102,151`: an empty `bundle_id` cell cannot be told apart from "the server reported no value"

`bundleID, _ := app["bundleId"].(string)` discards the comma-ok. That is this repo's established convention across every `pro_report_*.go` file, so the style is not a deviation. The consequence changed: a missing key, a JSON null, or a non-string type all yield `""`, which is now part of the grouping identity and renders as an empty cell. No test covers an absent or empty `bundleId` with `--bundle-id` on.

`unresolved:` whether the Jamf Pro API ever omits `bundleId` in practice, which needs a live tenant.

</details>

<details>

**Recommendation: deliver as a single PR.**

`--bundle-id` and `--path` are two instances of one mechanism applied to the same function. Each extends `softwareKey` with a field, gates it on its flag when counting and when building the row, and adds a tie-breaker to the comparator. They land in the same struct, the same function signature, and the same test file. Splitting them would widen that signature twice and touch the same comparator twice, for a change a reviewer verifies identically both times. That is coordination cost with no independent review value.

The 217 test lines against roughly 45 changed production lines are proportionate. The risk being covered is combinatorial, namely the four flag combinations of a two-flag boolean space.

The comparator and row-builder restructuring is not a separable refactor. It is the minimum shape the feature requires, and it would not exist without the feature.

Nothing in the diff falls outside the stated purpose. Every hunk is either the flag mechanism or a call-site update to the four pre-existing tests.

Whether `--path` belonged in this PR at all, given issue #325 asked only for the bundle ID, is a scope question rather than a delivery one. It came from a non-member comment and is disclosed in the PR title and body rather than smuggled in.

</details>

<details>

- [x] Design and architecture: grouping-key extension vs display column, `softwareKey:74-79`
- [x] Correctness: all four flag combinations probed; sort is a total order
- [x] Security: read-only diff, no new request, no credential surface
- [x] Performance: row cardinality under `--path`, see (4)
- [x] Test coverage: 6 mutations run, 5 killed, 1 survived, see (2)
- [x] Reliability: empty and non-string wire values, see (6)
- [x] Documentation currency: `examples.json` gap, see (5)
- [x] Cross-repo contracts: none; `bundleId` and `path` already on the wire, no spec change
- [x] Project rules compliance: CLAUDE.md conventions and credential policy
- [na] Frontend, simplification.

Diff: 2 files, +260/−14, head `d0b3fd5`, plus `internal/output/output.go`, `internal/commands/root.go`, `pro_helpers.go` and `docs/site/examples.json` read for context. Lanes run: all eight. Never run: none.

</details>

<details>

✅ **The grouping-key decision is correct, and the reasoning for it checks out.** One title and version really can span several bundle IDs or paths, and a display-only column would print one arbitrary group member. Extending the key is the honest answer.

✅ **Gating the row keys on the flag rather than on emptiness is the right call, for the reason the PR gives.** `printTable` and `printCSV` both derive their column set from `sortedKeys(rows[0])` in `internal/output/output.go`, so an emptiness gate would make the column set depend on sort order. CLAUDE.md documents this first-row-keys behaviour as a past bug in `config list`. The PR avoided it.

✅ **The unstable sort is safe here, which is worth stating because `sort.Slice` gives no guarantee.** Two distinct map keys must differ in at least one keyed field, and the comparator now covers all four, so it is a total order and the output is deterministic in every flag combination.

✅ **The PR body's mutation claim is true.** I reproduced it independently: neutralising both key assignments fails exactly the three grouping tests and nothing else. One refinement, dropping the assignments outright does not compile, because `bundleID` becomes unused, so the compiler is a partial guard on that specific edit.

</details>


grahampugh (approved):

PR #361, opened 2026-09-04
perf: cut the CI Test step by 57% by removing repeated spec parsing

ktn-jamf:
## What this does

The CI `Test` step is the critical path of the job. It takes 160s of a 209s job total, as a median of 11 green runs. Two changes cut that time by more than half. Both changes remove repeated work. Neither change removes a check.

| | before | after | change |
|---|---|---|---|
| suite wall time | 71.9s | **31.2s** | **−56.6%** |
| suite CPU time | 165.8s | **50.8s** | **−69.4%** |
| peak RSS | 375.8 MB | 384.9 MB | flat, inside noise |
| tests run | 3567 | 3568 | +1, a new assert |

CPU time is the figure that transfers to a 4-core runner. The measurement pins `-p 4 -parallel 4` to match `ubuntu-latest`. A measurement at the local core count overstates the gain.

CI confirms the gain. Two samples on this branch measured the `Test` step at 77s and 74s, against a baseline median of 160s and a baseline range of 134s to 179s. Both samples sit below the baseline minimum, and the job total fell to 138s and 127s.

The three `verify-generated` steps also fell from about 5s each to 2s each, and this change does not touch them. `make generate` runs the same parser.

## The two changes

**1. One shared OpenAPI loader** (`generator/parser`). 126 of the 165 committed specs use an external `$ref` to `specs/_MonolithLibrary.yaml`. That file holds 100 KB of YAML. `ParseSpec` created a new `openapi3.Loader` for each call. The loader therefore unmarshalled the library file one time for each referring spec. One shared loader unmarshals that file one time for each process.

`make generate` receives the same gain. This change is therefore more than a test change.

**2. Build the gateway op fixture one time for each test binary** (`generator/gateway`). Three tests need the same fixture. Each test rebuilt the fixture from a fresh parse of all 165 specs. `sync.OnceValues` builds the fixture one time. The accessor returns `slices.Clone`, so one test cannot contaminate the view of another test. The `gatewayOp` struct holds only strings, therefore the copy is complete.

## Why the gain is this large

`-race` is a multiplier, not the cost. One affected test runs in 1.99s without `-race`. The same test runs in 23.8s with `-race`. Its CPU profile shows `runtime.madvise` and `runtime.kevent`. Those functions are allocation bookkeeping, not YAML parsing. Each allocation that this change removes therefore pays about four times over.

`-race` is unchanged. It still runs on every package.

## Safety of the shared loader

A shared loader gives the same `*openapi3.T` to two specs that both `$ref` the library. One in-place edit can therefore contaminate the parse of every other spec.

The proof is `make verify-generated`. That target regenerates all three command trees and diffs them. **The output is identical, byte for byte.** No file under `internal/commands/*/generated/` appears in this diff.

`ParsePlatformSpec` keeps its own local loader. A normalized temporary spec therefore never enters the shared cache.

## The assert

The gain depends on one behavior of `openapi3.Loader`: the loader caches a document by the path that it was read from. kin-openapi does not document that behavior. If a later version removes the cache, the loader reads every spec again. The gain then disappears, and no test fails.

`TestSharedSpecLoaderCachesByPath` makes that silent loss loud. The message of the test names the remedy. A revert of the change makes the test fail, as intended.

## Verification

Every gate that CI enforces passed on a local run, in the order that CI uses:

- `make verify-generated`
- `make verify-platform-specs`
- `make verify-security-specs`
- `golangci-lint run`
- `make build`
- `go build ./generator/site/...`
- `./scripts/verify-site-products.sh`
- `node --check` on the site JS files
- `make verify-site-output`

The suite is green. This change also holds four counts equal or better: tests run, test functions, assertions, and skips. It removes no assertion. It skips no test. It weakens no test body.

## Rejected

- **Removal of `-race`, or a narrower scope for it.** A per-package allowlist goes stale in silence. If someone adds a goroutine to a package on the list, the check no longer covers that package.
- **A second concurrent job for the race pass.** This split lowers wall time on the critical path. It also costs runner minutes. The choice belongs to the workflow, not to this change.
- **Memoization of `parseSchema` by pointer.** This approach needs a three-part key. It also shares `*Schema` across resources. The complexity is not worth the gain.
- **A load of normalized platform specs from memory.** Wall time stayed flat. CPU time rose 3.7%.
- **A higher parallelism setting.** Go defaults `-p` and `-parallel` to `GOMAXPROCS`. The runner has 4 cores. CI therefore pins the parallelism already.

## Next lever

The critical path is now `internal/client`, at 19.3s. Two real `time.Sleep` calls hold 14s of that time. The calls are in `TestDo_RateLimited429` and `TestDoWithRetry_ExhaustsRetries`. A correction needs an injectable backoff. That work belongs in its own change.


neilmartin83 (approved):
> [!TIP]
> ✅ **Merge-ready**. Shares one `openapi3.Loader` across `ParseSpec` calls and memoizes the gateway test fixture, cutting the CI `Test` step by more than half. No critical or important issues found.
> See the collapsed section for 5 nice-to-have suggestions.

## Rating: 5/5
- Clean, and the headline claim reproduces independently. The 5 nice-to-have suggestions below are optional and do not hold the merge.
- Coverage: `performance-reviewer` and `test-quality-reviewer` have never run on this PR. Both dimensions were checked directly instead, by measuring both branches and reading the new test.

<details>

💡 **NICE-TO-HAVE** (1) (design) — `generator/parser/parser.go:568`: the mutex does not make `ParseSpec` concurrency-safe

The comment says the mutex exists so a future concurrent caller is not the one to discover that `openapi3.Loader` is unsafe for concurrent use. The lock releases before `ParseLoadedSpec` runs, and the documents handed back are shared: two specs that both `$ref` `_MonolithLibrary.yaml` get the same `*openapi3.T`. One goroutine can read library-resident schemas in `ParseLoadedSpec` while another's `LoadFromFile` re-walks that subtree and writes `ref.Value` and `Discriminator.Mapping[k]` (kin-openapi `loader.go:1157`). So the mutex reads as reassurance it does not give.

**Suggested fix:**
```diff
-	specLoaderMu.Lock()
-	doc, err := specLoader.LoadFromFile(specPath)
-	specLoaderMu.Unlock()
+	specLoaderMu.Lock()
+	defer specLoaderMu.Unlock()
+	doc, err := specLoader.LoadFromFile(specPath)
```

Free today, since generation is sequential. Dropping the mutex and saying plainly that `ParseSpec` is not concurrency-safe is equally honest.

**Promote to IMPORTANT when:** any caller invokes `ParseSpec` from more than one goroutine.

💡 **NICE-TO-HAVE** (2) (correctness, also flagged independently by the devil-advocate lane) — `generator/parser/parser.go:556`: "nothing here writes to the parsed documents" is incomplete, and one write is parse-order dependent

kin-openapi writes into the documents it returns: `resolveSchemaRef` assigns `value.Discriminator.Mapping[k] = v` (`loader.go:1157`), gated on `inExternalDoc` (`loader.go:1141`), computed from `loader.rootLocation` — which `loadFromDataWithPathInternal` sets exactly once, on the first document a loader instance ever reads. A per-call loader made that the spec being parsed. A shared loader makes it whichever spec the process loaded first, so every later spec's own `#/components/schemas/…` discriminator mappings now resolve where they were skipped. Two committed specs carry such mappings (`specs/MobileDeviceInventoryDetail.yaml:565`, `specs/MdmCommands.yaml:440`). Inert today — `parseSchemaDepth` reads only `Discriminator.PropertyName` (`parser.go:1998`) — but the comment is what a future reader trusts.

**Fixed when:** the comment names kin-openapi's own in-place writes and the `rootLocation` order dependence, instead of asserting the documents are never written.

💡 **NICE-TO-HAVE** (3) (test-coverage) — `generator/parser/spec_loader_test.go:22`: the guard asserts a weaker property than the win rests on, and writes to shared package state

The saving comes from the external-`$ref` path reusing the cached library document across 126 referring specs. The test loads that library directly through `specLoader` twice and compares pointers, which asserts the top-level-load cache. Both share `visitedDocuments` in v0.147.0, so it passes — but an upstream change that kept top-level caching and altered the ref-resolution key leaves it green with the win gone, the exact silent loss it was written to catch. It also caches the library under a new key on the package-level loader, and pins `rootLocation` when it runs first in the binary.

**Suggested fix:** load two specs that both `$ref` the library on a locally-constructed loader with the same config, then assert pointer identity of a resolved library schema across the two. That asserts the mechanism and touches no shared state.

💡 **NICE-TO-HAVE** (4) (code-quality) — `generator/parser/parser.go:596`: `ParseLoadedSpec` need not be exported

Its only callers are `ParseSpec` and `ParsePlatformSpec` (`platform.go:77`), both in `package parser`. Exporting it adds a public entry point and puts `*openapi3.T` on the package's exported surface.

**Suggested fix:**
```diff
-func ParseLoadedSpec(doc *openapi3.T, specPath string) ([]*Resource, error) {
+func parseLoadedSpec(doc *openapi3.T, specPath string) ([]*Resource, error) {
```

💡 **NICE-TO-HAVE** (5) (documentation) — `CLAUDE.md`: the shared loader is unrecorded in the living docs

CLAUDE.md's "Code Generation Pipeline" section describes `ParseSpec` and says nothing about a process-wide loader whose gain rests on undocumented kin-openapi behavior. Nothing under `docs/solutions/` records it either. This repo documents smaller decisions than this one at length.

**Fixed when:** one paragraph in the pipeline section, or a `docs/solutions/conventions/` note, records the shared loader, why it is sound, and that the win depends on path-keyed caching upstream.

</details>

<details>

- [x] Design and architecture: shared mutable loader traced, see (1), (2)
- [x] Correctness: `parseSchemaDepth`, `composedRequired`, kin-openapi write paths
- [x] Performance: measured both branches, `-p 4 -parallel 4`, claim reproduces
- [x] Test coverage: new guard is (3); memoized fixture clones value data
- [x] Reliability: `main.go:187` warn-and-continue against the cache; loud, not silent
- [x] Documentation currency: CLAUDE.md pipeline section is (5)
- [x] Project rules compliance: `CLAUDE.md`, generated-code boundary, toolchain pin
- [na] Security, cross-repo contracts, frontend, simplification.

Diff: 4 files, +111/−10, head `cc2b0dd`. Read `generator/parser/parser.go`, `generator/parser/platform.go`, `generator/main.go:185-195`, kin-openapi v0.147.0 `openapi3/loader.go`, and the discriminator blocks of three Pro specs. Ran `make verify-generated`, `golangci-lint run`, `go test -race` on both branches. Lanes run: devil-advocate. Never run: performance-reviewer, test-quality-reviewer (skipped, verified directly); security-reviewer, scope-reviewer, fidelity-reviewer, usability-reviewer, silent-failure-hunter (inapplicable).

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/parser/parser.go` | 4/5 | Sound sharing decision; the safety comment overstates it (see (1), (2)) |
| `generator/gateway/overrides_test.go` | 5/5 | Correct memoization: `gatewayOp` is value data, so `slices.Clone` is a full copy |
| `generator/parser/spec_loader_test.go` | 4/5 | Real guard against a real silent loss, aimed one layer off (see (3)) |
| `generator/parser/platform.go` | 5/5 | Keeps a temp-spec path out of the shared cache |

</details>

<details>

✅ The gain reproduces, and is larger than claimed. Measured here on the two affected packages, `go test -race -count=1 -p 4 -parallel 4`: `generator/parser` 64.0s → 11.6s, `generator/gateway` 90.0s → 5.3s, wall 92.0s → 12.5s, CPU 165.4s → 16.2s. `make verify-generated` is byte-identical, `golangci-lint` clean, `-race` clean.

✅ Routing `ParsePlatformSpec`'s fallback through `ParseLoadedSpec` rather than `ParseSpec` is the non-obvious half. It keeps a temp path the caller then deletes out of a now process-lived cache, and it drops a redundant read and unmarshal of a file already in hand.

✅ Moving `t.Fatalf` out of the memoized builder is the right call, not incidental. A `t.FailNow` inside a `sync.Once` body fails whichever test happened to trigger the build, which on a `-run` filter or a reordered run is not the test that needs the message.

✅ The change is a generator win, not only a test win. `make generate` reads the same loader, and the three `verify-generated` steps fell with it.

</details>

PR #360, opened 2026-09-04
fix(commands): refuse a stray positional on every leaf that documents none

ktn-jamf:
Closes #350

Cobra validates `Args` for each command and supplies no default. A runnable leaf
that declared no validator therefore accepted any positional argument and threw
it away. `guardStrayPositionals` now walks the assembled tree beside
`guardUnknownSubcommands` and gives a validator to every leaf that documents no
placeholder in its `Use` string. 736 leaves are covered.

The `Use` string decides, because it is where each command already states its
positional contract. A leaf that documents a placeholder keeps the validator it
declares. This is not a repository-wide sweep to `cobra.NoArgs`, which the issue
correctly refuses: 700 leaves take an identifier and continue to accept one.

## Before and after

No credentials are necessary to show this. An `Args` validator runs before
`PersistentPreRunE`, so a refusal replaces the credential error.

The two reproductions in the issue:

```
                                        BEFORE (origin/main 308f7f9)
$ jamf-cli pro categories list junkarg
  exit 2  server URL is required: use --url, JAMF_URL env var, or ...
$ jamf-cli pro report inventory-summary junkarg
  exit 2  server URL is required: use --url, JAMF_URL env var, or ...

                                        AFTER
$ jamf-cli pro categories list junkarg
  exit 2  "jamf-cli pro categories list" takes no positional arguments, but got "junkarg"
          hint: run jamf-cli pro categories list --help for the flags it accepts
$ jamf-cli pro report inventory-summary junkarg
  exit 2  "jamf-cli pro report inventory-summary" takes no positional arguments, but got "junkarg"
```

Both commands exit 2 before the change as well, because this machine holds no
credentials. The message is the evidence: before the change each command passed
argument validation and failed later in authentication, which proves the
positional was accepted and discarded. With credentials the list ran and the
report fetched page 0.

The destructive shape the issue names:

```
$ jamf-cli pro backup /tmp/out
  before: the whole backup ran, and --output ignored the directory
  after:  exit 2  "jamf-cli pro backup" takes no positional arguments, but got "/tmp/out"
```

## Three defects in the refusal itself

**The message described the wrong mistake.** `cobra.NoArgs` answers
`unknown command "x" for "y"`, which is `guardUnknownSubcommands`' message for a
parent command. A leaf has no subcommand the value could have named, so that
wording sends the reader to look for one. `pro backup /tmp/out` reported an
unknown command for a directory that `--output` takes, and named neither the
[...]

PR #359, opened 2026-09-04
fix(output): print every command through the shared output formatter

ktn-jamf:
Closes #349

## Why

`PersistentPreRunE` builds one output formatter and applies six global flags to it. `SetWriter` for `--out-file`, `SetProjector` for `--select` and `--compact`, then `SetQuiet`, `SetNoHints` and `SetExplicitNoColor`. It stores the result as `cliCtx.Output`.

27 commands built a second formatter with `output.New(...)` and wrote to `os.Stdout`. That formatter receives none of the setters, so those six flags were parsed, accepted and discarded.

Nothing reported it. `--out-file` still created the file, the payload still went to standard output, and the exit code was 0. A scheduled job cannot tell that from an empty tenant.

## What changed

Three functions in `internal/commands/output_route.go` are the route:

- `printRows` prints rows through the shared formatter.
- `formatterFor` clones it for a caller whose own argument names the format, which is `multi` and `group-tools export`.
- `printSection` writes a section header above rows, and decides the header before writing it.

A command that renders its own text takes `writerFor(cliCtx)` as its writer. `doctor` already used that helper.

The 27 sites are the 24 #349 names plus three its grep missed: `newCommandsCmd`, `printAggregated` in `multi.go`, and `runGroupToolsExport`. `pro_backup.go` is in the issue's file list and holds no formatter.

## The renderers

Routing the flags exposed four renderer defects, all of them older than this PR and all reachable only once `--select` and `--compact` took effect on these commands.

A projection makes rows heterogeneous, because a row keeps a key only where it matched. `printTable`, `printCSV` and `printPlain` read their column set from row 0, so a key row 0 lacked became a column for no row. `commands -o csv --select command,api` wrote a `command`-only header while `-o json` returned 1375 `api` values. The column set is now the union across rows while a projector is live. Without one, row 0 still decides and every existing table is byte-identical.

A projection that matches nothing in any row now renders nothing. It used to print `RESULTS (N total)` above a blank header, and under `plain` one blank line per row. `Formatter.Print` names the miss on stderr, and `--quiet` and `--no-hints` do not suppress it, because those four renderers write nothing at all and no other output reports it.

`--select` bypasses the table's default-column heuristic. It dropped `api` from `commands -o table --select api` until `--wide` was added.

`--select " "` used to render wider than `--wide`. `Projector.IsZero` now measures the trimmed path list.

## Per-format contracts

[...]

Issue #354, opened 2026-09-04
`platform setup` incorrectly identifies Scope as Platform Environment instead of Tenant


ckay-unsw:
### jamf-cli version

1.28.0

### Product namespace

pro (Jamf Pro / Platform API)

### Auth method

platform (gateway)

### OS / platform

macOS 26

### Command run

```shell
jamf-cli platform setup

Integration is using Level "Tenant" and not "Platform environment"
```

### Expected behavior

Scope should prompt for Tenant ID

### Actual behavior

```shell
Scope prompted for Platform Environment ID and if you paste in the Tenant ID you can't use jamf-cli to access the gateway successfully.
```

neilmartin83:
Hey @ckay-unsw thanks for flagging this, it's more a UX issue. You can hit enter to skip the Environment ID then you're prompted for Tenant ID.

Thought about trying to do something more deterministic with a single ID but there are too many edge cases where the logic could fail (e.g if the tenant is Protect, Pro or JSC and what permissions the integration has). Tenant IDs are also considered legacy and for cases where you can't create an environment for some reason etc. They can't hit Blueprints/Benchmarks and other Platform level services. So Environments are much preferred now.

This is with a Tenant ID and it works end to end - is there anything in the prompts/wording/flow that you think we could improve?

```
% jamf-cli platform setup 
Profile name [platform]: platform-mockingbird-tenant

Platform gateway region:
  1. US (https://us.api.jamfcloud.com)
  2. EU (https://eu.api.jamfcloud.com)
  3. APAC (https://apac.api.jamfcloud.com)
  4. Custom URL
Choose [1-4]: 2

Client ID: REDACTED
Client Secret: 

An API integration is created at one level: organization, platform environment,
or tenant. Supply the ID for the level this integration was created at.
Platform environment ID (Enter to skip): 
Tenant ID (Enter to skip for an organization-scoped integration): 85f69825-dc92-4522-aa3c-34eea2f20bc5

Validating credentials... ok
Checking Jamf Security Cloud access... no (no Security Cloud entitlement)

Profile "platform-mockingbird-tenant" saved and set as default.
  Gateway:     https://eu.api.jamfcloud.com
  Scope:       tenant 85f69825-dc92-4522-aa3c-34eea2f20bc5
  Client ID:   REDACTED
  Secrets stored in system keychain

This scope serves the Pro API and Platform API commands.
```

iampavan:
Just a note on this last line: 

[...]

PR #346, opened 2026-09-03
feat(pro backup): advanced searches, a resource listing, and a typo guard

ktn-jamf:
Closes #341. Closes #340.

Three changes to `pro backup`. The third is a bug fix that the first two made reachable.

## 1. Advanced searches join the backup set (#341)

Two entries under one new `--resources` token, `advanced-searches`.

The computer half reads the Classic API and the mobile half reads the Pro API. No modern advanced-computer-searches spec exists — `specs/` carries only `AdvancedMobileDeviceSearch.yaml` and `AdvancedUserContentSearch.yaml` — so the asymmetry is the resource table's own modern-over-classic rule answering differently for the two halves. It is not a preference. All four list and get paths declare `GET` in `specs/gateway/coverage.json`, which `TestBackupResourcePathsAreServed` asserts, so neither entry adds a request the gateway refuses.

The issue calls these smart searches. Jamf calls them advanced searches, and smart *groups* are a different object that `pro backup` already captures under `smart-groups`. The token follows Jamf.

`advanced-user-content-searches` is not added. The issue asks for computer and mobile only.

## 2. `pro backup list-resources` (#340)

Before this change there were two ways to learn what `--resources` accepts. Read `pro_resources.go`, or pass a wrong value and read the error. The flag had no shell completion either, unlike `protect backup`.

```
$ jamf-cli pro backup list-resources -o table
 RESOURCE                SOURCE
──────────────────────────────────────────────
 accounts                classic api
 advanced-searches       classic api, pro api
 blueprints              platform sdk
 ...
```

`table`, `csv` and `plain` show the token and its API. `json`, `yaml` and `ndjson` add a third field, `objects`, naming the generated commands behind each token.

Every column is derived. The row order and the row set come from `BackupFilterNames()`, so a token the flag accepts cannot go unlisted. `source` comes from `BackupEndpoint.IsClassic`. The three resources handled outside the generated registry have no endpoint to read, so `nonStandardBackupFilter` gains a `Source` field, and a test sweeps the live tables for any token that declares none. No code switches on a filter name.

[...]

PR #365, opened 2026-09-05
feat!: ingest SDK v0.22.2 — restore the Classic patch family, fix device-group update, report declared scope

neilmartin83:
Ingests `jamfplatform-go-sdk` **v0.22.2** (`d4ab4f4`, GitOps build **v2082**) and acts on what it carries. Coverage moves to Jamf Pro **476 paths / 700 operations** and Classic **270 / 589**.

Four things in one branch because the ingest is what forced each: a restoration, a command that would otherwise have vanished, a spec field the CLI had never read, and a repeated-element name the spec got wrong.

## Eight refused commands work again

`public-apis-oas#438` restored the whole Classic patch-management family, on the stated reasoning that patch management is where Classic API callers are most concentrated. So `pro classic-patch-reports` (2), `pro classic-patch-titles` `list`/`get`/`update`/`delete`/`apply` (5) and `pro classic-patch-policies list` are no longer refused — **75 → 67**, the first time that count has moved down. `classic-patch-titles` also picks up a bound body schema, so its writes gain `--scaffold`, `--set` and enum help (Classic write coverage → 114 of 117 commands, 44 of 54 resources).

Nothing on the wire ever changed. The v1993 bundle carried all 606 Classic operations throughout and granted every one of the 31 it withdrew, so those eight commands worked the entire time they were refused. Confirmed through the CLI on an EU environment credential immediately after the ingest lifted the refusal: `classic-patch-titles list` returns a real title, `classic-patch-reports get 1` returns the report.

That is the case `JAMF_CLI_ALLOW_UNPUBLISHED` exists for, and it is still an argument for following the publication rather than second-guessing it: the two withdrawals in this repo's history that looked most like publication defects went **opposite** ways — v1942's was the CLI sending a superseded version, this one was reversed — and following the spec was right both times.

## `security device-groups update` was one commit from vanishing silently

The v2 PUT handler was broken (403 until `authorization-policies#265` deployed on 2026-09-03, then a bare 404 on a group its own `list` had just returned) and was fixed on 2026-09-04. Build v2082 then withdrew the v1 list and PUT outright — so with `platformUnroutedOps` still dropping the v2 PUT, the resource shipped **no update command at all**.

`TestSecurityCloudSpecParity` caught it as 51 subcommands against 52 declared operations. The only other sign was one fewer line in `--help`, which is the lesson: a stale drop is a *silently missing* command, not a broken one.

Re-probed here 2026-09-05 before emptying the table — the v2 PUT answers **204 3/3** and the rename reads back through `GET /v2/groups`, a different operation from the one written, so a handler accepting a write and discarding it would not have passed; `PUT /v1/groups/{id}` at 200 and a bogus id at a field-attributed `404 GROUP_NOT_FOUND` as controls in the same invocation. Verified end to end through the CLI: create → `update --set` → `list` showing the new name → delete.

**Breaking:** a successful `update` now prints nothing (204 no body), where v1 returned 200 with the group. `create` and `delete` stay on v1; `list` and `update` are v2.

## The specs now declare a scope level, and the CLI reports it

v2082 declared `x-scope-types` on every Platform spec that declares one at all — the three `account_*.json` specs declare none, which is why an absent annotation means "the spec is silent" rather than "any level works". Six went tenant → **environment only** with `X-Environment-Id` required; jpapi, capi and all six Security Cloud specs went tenant → **tenant and environment**. Read by `generator/parser`, attached per operation, stamped as `jamf:scopes`, surfaced in `commands -o json` as **`scopes`** on 110 commands.

This is the one requirement a 403 cannot teach: the gateway's refusal names a capability permission and says nothing about the level, so an operator holding a correctly-granted credential at the wrong level is sent to Jamf Account to tick a box that changes nothing.

**Nothing refuses on it, deliberately — the spec is stricter than the gateway.** Probed 2026-09-05 on a tenant credential: `pro platform-devices list` and `pro platform-device-groups list` both answer 200 despite being declared environment-only, with `pro categories list` at 200 as the routed control. (`blueprints` and `compliance-benchmarks` answered 403 on that credential and are deliberately *not* pinned either way — classifying a 403 in this namespace takes two credentials, not two paths.) A pre-flight refusal keyed on this data would have refused at least two working commands.

It retires `annotateAuditScopeError`, which spelled the same fact for one command by hand and had already gone stale once: it used to add that the spec listed organization as allowed while the gateway refused it, a sentence v2056 made false. Every platform command gets the derived note now. Audit's `--help` text stays hand-written — it carries the code each wrong level earns and that audit is served in every region, neither of which an annotation states.

### Two wrong statements it exposed

`platform setup`'s closing summary told a **tenant**-scoped operator the profile "serves the Pro API and Platform API commands" when 13 of the 29 platform resources declare environment scope — the question raised on #354 — and told an **organization**-scoped one that AI Governance was served, which is false: `GET /ai/governance/policies/v1/policies` answers `400 REQUEST_CONTEXT_NOT_PROVIDED` with no scope header (probed in US with `GET /licensing/v1/licenses` at 200 in the same run). The summary is assembled from the annotations now, so it cannot claim a surface the specs rule out. The Security Cloud entitlement probe *qualifies* that list rather than replacing it — the old code printed only the Security Cloud sentence when the probe passed, hiding that the profile drives Pro too.

## A profile's scope level is no longer spliced onto other credentials

**Breaking.** An integration is created at exactly one level in Jamf Account and its credential carries that choice, so `resolveScope`'s per-input ladder was wrong for the scope: `JAMF_URL` + `JAMF_CLIENT_ID` + `JAMF_CLIENT_SECRET` for an organization-scoped integration, with any default profile carrying a `tenant-id`, sent an `X-Tenant-Id` that appeared in no flag, no variable and no command the operator typed. An organization-scoped credential must send no scope header at all.

Now: if the **client ID** comes from `JAMF_CLIENT_ID` — the only way one arrives, there being no `--client-id` flag in this CLI — the profile's `environment-id` and `tenant-id` are **both** dropped. Keyed on the client ID because that is what *names* an integration — a profile holding `client-id` with only `JAMF_CLIENT_SECRET` injected is still the profile's integration, a reasonable CI shape the config's own `env:` refs already serve. The full pair always trips it.

Nothing silently reached the wrong tenant, so this fixes a misattributed error rather than a wrong target: wire-checked, a tenant ID the credential does not own answers `403 OWNERSHIP_FORBIDDEN`. It is also what made `annotateAuditScopeError` look unreachable — the note is gated on `REQUEST_CONTEXT_NOT_PROVIDED`, the probe that "disproved" it had a tenant-carrying default profile, and with a clean config the answer is the code the note was written for all along.

Three details are load-bearing, all about *where* the rule reads its inputs:

- `resolveScope` reads `JAMF_CLIENT_ID` from the environment **directly**, because `PersistentPreRunE` returns for the `security` product before `resolveAuth` folds the env vars — so a rule reading only the folded var would silently not apply to the 52 gateway-served Security Cloud commands. Same structural trap that once left `--tenant-id` unread there while `--url` was honoured.
- `ResolveAuthForProfile` reads `params.ClientID` **only**, because `resolveAuth` has already folded by then and params is the complete answer. Reading the mutable package var there made the function depend on global state its params exist to isolate — which surfaced as `TestResolveAuthScopePrecedence` failing only when the whole package ran.
- Both ladders use the **resolved** profile name from `config.GetProfile`. Passing the requested empty string on made the rule read as "there is no profile" and dropped the scope of every default-profile user. That is the first case in the new table test.

The error now names the profile whose level was passed over and both ways to supply one; `scopeLevelNote` hands it the remedy rather than rendering a second one (the first version said "no scope header was sent" twice and then advised setting an ID on the very profile whose ID had just been ignored).

## A `commandEntry` field does not reach the catalog by existing

`scopes` was populated, unit-tested and **absent** from `commands -o json` for all 110 commands, because `commandEntriesToMaps` is a hand-maintained projection. Fixed, and `TestEveryCommandEntryFieldReachesTheCatalog` walks the struct's json tags against the shipped tree so the next field cannot be lost the same way. It caught `gatewaySuccessor` independently, which #345 has since fixed; that PR's `TestCatalogJSONCarriesTheSuccessorKey` is kept alongside this one, since it pins the positive-only contract in *both* directions and a key-presence sweep cannot see that.

## The `os_x_configuration_profile` scope element is `<user_group>`

v0.22.1's single fix lands entirely inside the Classic body schemas. `os_x_configuration_profile.scope.jss_user_groups` declared `<jss_user_group>` where the wire answers `<user_group>`, so `pro classic-macos-config-profiles create --scaffold` now renders the targets position the same way it has always rendered the exclusions one.

The typo was confined to that one property: all seven sibling Classic resources carrying the same scope block declare `user_group`, and so does that resource's **own** `scope.exclusions.jss_user_groups`. Writes were never affected — the Classic API accepts both spellings and reads the scope back as `<user_group>`. What the mismatch cost was *reads*, in the SDK: a generated Classic type leaves `XMLName xml.Name` untagged, so a mismatched child decodes to nothing and the call reports success, and the scoped user group was dropped silently on every decode. This CLI renders the server's XML rather than decoding it into those types, so the scaffold is the only surface it reached here.

No platform spec moved and no path or operation count changed. The rest of the diff is the recorded SDK revision.

CLAUDE.md's repeated-element note gains the part this proves: reading the element name off the schema is necessary and **not sufficient**, because the schema itself can be wrong about it. A sibling resource carrying the same block, and the resource's own exclusions half, are the cheap cross-checks.

## SDK v0.22.2 — a Self Service category needs `display_in`

Ingested on top of v0.22.1. One `schemaPatches` entry upstream, and the CLI surface is one Classic resource's writes.

`mobile_device_configuration_profile.self_service.self_service_categories[].category` `$ref`'d the shared `category` schema (`{id, name, priority}`) where its five siblings declare the item inline with `display_in`. That field is what decides whether the category is stored **at all**: wire-verified on Jamf Pro 11.31.1 across all six resources carrying the block, a `<category>` holding only `<id>`, or `<id>` plus `<name>`, or `<feature_in>` without `<display_in>`, is **accepted with 201 and silently discarded**, and `display_in=false` is a deletion gesture rather than a stored value. So the law is a property of the Self Service category relation, not a quirk of one resource.

What *is* unique here: `display_in` is **write-only on this resource alone** — the GET echoes `<id>` and `<name>` only, where the other five echo it — so the one resource whose spec omitted the property is also the one whose read surface conceals it. A consumer reasoning from `os_x_configuration_profile` would conclude the field is readable and be wrong here only.

`pro classic-mobile-config-profiles create`/`update`/`apply` now carry `display_in` in `--scaffold`, `--set` and field help. **The scaffold renders `<display_in>false</display_in>`** — the boolean placeholder, which is the discard value — so it joins `general.category.id: 0` and ztna's `categoryName: ""` as a rendered value that is guaranteed wrong rather than merely incomplete. Both facts are carried in the field's description in `specs/classic/schemas.json`, and in CLAUDE.md and the CHANGELOG.

No platform spec moved and no path or operation count changed; the rest of the ingest is the recorded revision.

## Review round 2 — all five blocking findings fixed, all three nice-to-haves done

The review's diagnosis is right and worth restating: each round-1 fix replaced a vague statement with a specific one, and four of the specific ones claimed more than the code knows. That is the PR's own "declares, never requires" discipline failing inside the messages rather than in the refusal logic.

**(1) The note asserted what the gateway would have done, and dropped the remedy that works.** The disagreeing branch's "would not have worked here anyway" is a *requires* claim, against the counter-example CLAUDE.md records — a tenant credential answers 200 on `pro platform-devices list`, which declares environment scope. It also returned before the remedy paragraph, so the withheld level's own variable and flag were never printed. Both halves are stated now, in order of durability, and `withheldReasonClause` factors the shared opening so the two branches cannot drift.

**(2) The school arm recommended the one thing that path cannot do.** It passed `scopesOf(cmd)`; `school blueprints list` declares environment scope (asserted in the new test rather than assumed, so it cannot go vacuous), so the disagreeing branch rendered and advised an integration created at a declared level — while `resolveSchoolClient` has no environment code path at all. `nil` now: no request was sent, so the arm has no opinion on the declared level, and offering the withheld level back is the whole remedy. The new test goes through the same `*cobra.Command` `main.go` passes, which is what the nil-argument sibling could not see.

**(3) One 403 stated a licensing verdict.** Agreed on all three counts — `BAD_PERMISSIONS` is indistinguishable from a missing capability grant, and the recorded `/devices/v1/devices` control proves the *device* grants rather than `content-categories:read`. The subtraction stays; the claim does not. The clause names both causes and sends the reader to the integration's permissions before licensing, and `securityCloudUnentitled`'s doc comment says so too. The probe line itself changed from "no (no Security Cloud entitlement)" to "no — refused (no entitlement, or the integration lacks this permission)".

**(4) The third verdict was inert at its only consumer.** `printScopeSummary` branched on `== securityCloudUnentitled`, so `securityCloudUnknown` took the same path as a confirmed 200. It reports the entitlement as unknown now, guarded so the sentence is not printed for a level that reaches no Security Cloud resource anyway. Both mutations the finding names fail: collapsing the verdicts fails the unknown assertion, and widening to `!= securityCloudEntitled` fails the "must not subtract" assertion in the same test.

**(5) The `file:` read error was discarded and the note blamed the wrong cause.** The justification was false for the reason the finding gives, and that is the sharpest of the five: the branch is reachable only when the invocation supplied a client ID, and in exactly that case both ladders skip `ResolveSecret(p.ClientID)`, so nothing downstream ever opens the file. Recorded and named now — path and read error — with the environment variable no longer blamed for it. The read is bounded too: `readClientIDRefFile` refuses a non-regular file before `open()`, because `os.ReadFile` on a config-named FIFO blocks with no timeout and `/dev/zero` Stats as size 0 and then reads forever; the cap is asserted on a boundary-sized regular file so the test runs the same everywhere.

Nice-to-haves: **(6)** `securityPlatformSDKClient` resets at the top — it returns `nil, nil` *before* the `resolveScope` that was the only reset on that path, which is one of the two paths the reset's own comment names; the assertion is added to the existing reset test and removing the call fails it. **(7)** `main.go`'s ordering comment enumerates both `AnnotateScopeLevelError` triggers. **(8)** `TestCatalogJSONOmitsTheEmptyPositiveOnlyKeys` pins `scopes`, `privileges`, `gatewayPrivileges` and `gatewayPermissions` in both directions, judging each row against the entry field the projection was handed rather than against a second copy of the rule; making the `scopes` assignment unconditional emits `"scopes": null` on every Pro and Classic row and fails it.

### Wire-checked, EU environment-scoped credential, one run each

| shape | answer |
|---|---|
| profile `client-id: file:<matching>` + `environment-id`, `JAMF_CLIENT_ID` exported | **200** — the level is kept |
| profile `client-id: file:<a different id>` + `environment-id` | 400 `REQUEST_CONTEXT_NOT_PROVIDED` + the note naming the profile, the level and `JAMF_ENVIRONMENT_ID` / `--environment-id` |
| profile `client-id: file:<unmounted path>` + `environment-id` | 400 + the note naming **the path and the read error**, with no mention of `JAMF_CLIENT_ID` — finding (5) |
| profile `tenant-id` on the environment-declaring `pro platform-devices list` | 400 + the ranked note: declared level, durable answer, then `JAMF_TENANT_ID` "may still work today" — finding (1) |
| `school blueprints list`, tenant level withheld | names `JAMF_TENANT_ID` / `--tenant-id` and nothing about the environment declaration — finding (2) |
| the same with `JAMF_TENANT_ID` supplied | reaches `403 OWNERSHIP_FORBIDDEN` on a deliberately bogus tenant, i.e. **the remedy produces a platform client** — finding (2)'s "fixed when" |

The two verdict findings are unit-tested rather than probed: this credential *is* Security Cloud entitled (`security content-categories list` returns the catalogue), so neither the unentitled nor the inconclusive branch is reachable on it.

`make test`, `make lint` (0 issues) and all five verify targets clean; `make generate` idempotent.

## Review round 1 — all four blocking findings fixed

Each was reproduced before being fixed, and each is in the messaging and precedence layer rather than in the ingest or the generator.

**(1) A `keychain:` client-id reference always read as a foreign integration.** The predicate compared only an `env:` reference, and the stated reason for not resolving the others is that a keychain read can prompt — true of `keychain:`, not of `file:`, which is a plain read. `file:` is now resolved and compared by value; `keychain:` alone is withheld. Wire-checked on an EU environment credential: the `file:` shape answers 200, and with the branch removed the same invocation answers `400 REQUEST_CONTEXT_NOT_PROVIDED` — exactly the failure the finding described.

That leaves a real cost, since `keychain:<profile>/client-id` is what `platform setup` writes: such a profile, in a shell also exporting `JAMF_CLIENT_ID` for the same integration, still has its level withheld. The CHANGELOG names the shape, and the note names the profile, the level and both remedies. The `keychain:` row is also what makes the predicate's final `return false` load-bearing — returning true from it left the whole `internal/commands` suite green, which is the mutation the review flagged as surviving.

**(2) A withheld school scope reported missing credentials.** `resolveSchoolClient` requires a tenant ID before building a platform client, so a withheld level left it nil and the command answered "this command requires platform gateway auth" with the profile, client ID and secret all present. `platform.ErrNoPlatformClient` is the sentinel both gates now wrap, and `AnnotateScopeLevelError` annotates it — rather than a stderr warning at resolution time, which would fire on every `school` command including the ones that never touch the Platform API.

**(3) The tenant summary contradicted itself in the ordinary case.** Confirmed against the built binary: of the 29 groups carrying `jamf:scopes`, the 16 declaring `tenant` are exactly the Security Cloud ones, so an unentitled Jamf Pro tenant was told it reached sixteen resources and then that it lacked the entitlement for all sixteen. The probe answer now partitions the list. The set is read off the command tree rather than off the declared levels, because the two coincide only by accident — all six Security Cloud specs declare tenant **and** environment, so an environment credential declares all 29 and the entitlement has to subtract there too; a split keyed on "declares tenant" would have left that case claiming the whole surface.

**(4) The ownership refusal was worded tenant-only** while being reachable at either level, so a refused environment ID was called a tenant ID and the operator was told to use the prompt they had just used — with `printScopeSummary` naming the same value an environment ID in the next breath. The level is a parameter now, which removes an assumption rather than adding one.

All five nice-to-haves are done too, each of which was real: the doubled declared-levels sentence (a three-state `withheldState`, since returning `""` whenever anything was withheld would have dropped the level in the *other* branch), the missing `resolvedPlatformScope` reset and its stale contract comment, the inconclusive probe becoming a definite entitlement verdict, the stale Classic write-coverage counts in CLAUDE.md, and the two description inaccuracies above. One more fell out of reading the wire output back: the note printed "carries **a** environment ID".

## Verification

`make test`, `make lint` (0 issues), and `verify-generated` / `verify-gateway-coverage` / `verify-classic-schemas` / `verify-platform-specs` / `verify-site` all clean; `make generate` idempotent. Re-run after the ingest and after every review fix.

Live: an EU environment credential, an EU tenant credential and a US organization credential. The scope-withhold rule is wire-verified across five combinations — default profile with its own credentials sends its `X-Tenant-Id`; the same with an explicit `-p`; env credentials with a tenant-carrying default profile send no header on both the `pro`/`platform` and `security` paths; env credentials with `JAMF_ENVIRONMENT_ID` or `JAMF_TENANT_ID` send that level and answer 200.

## Merge ordering

Nothing here conflicts semantically with the three open PRs, and none of them adds a `commandEntry` field, so the new reflection test fires against none of them. Comments are on each with the specifics. **#346 has since merged and `main` is merged in here** — its `captureOutput` helper won the one conflict (a `(t, fn) string` signature), and `TestRunAnnotatesAScopeLevelError` was adapted to it. In short: this and **#359** in either order (with one post-merge check on `newCommandsCmd`), then **#360**.


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Ingests SDK v0.22.0, un-refuses eight Classic patch commands, moves `security device-groups update` onto the v2 PUT, and reads `x-scope-types` into a reported (never enforced) scope level.
> Blocking: (1), (2), (3), (4). See the collapsed sections for 5 nice-to-have suggestions and the decomposition note.

## Rating: 3/5
- Would be a 5 once the client-ID predicate recognises a `keychain:` reference (1), the school path reports its withheld level (2), and the two `platform setup` summaries stop contradicting themselves (3), (4).

The reporting-not-refusing decision is the right call and the evidence for it is in the code. Every blocking finding below is in the messaging and precedence layer built on top, not in the ingest or the generator.

---

⚠️ **IMPORTANT** (1) (correctness, via devil-advocate+security-reviewer+test-quality-reviewer) — `internal/commands/platform_scope.go:117-123`: a `keychain:` client-id reference always reads as a foreign integration

`profileNamesTheInvocationClientID` recognises only an `env:` reference, so any other form fails the same-integration test and both of the profile's scope IDs are dropped. `platform setup` always writes `keychain:<profile>/client-id` (`platform_gateway_setup.go:298`), and `config add-profile` puts a bare client ID in the keychain too, so this is the shape the CLI's own setup commands produce. The doc comment's reason for not resolving other forms is that a keychain read can prompt. That does not apply to `file:`, which is a plain file read.

**Failure scenario:** a profile written by `platform setup` (client-id `keychain:gw/client-id`, tenant-id `T`), invoked as `jamf-cli -p gw pro platform-devices list` in a shell that also exports `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` for that same integration → no scope header is sent, and the gateway answers `400 REQUEST_CONTEXT_NOT_PROVIDED` where the command exited 0 before this PR.

**Suggested fix:** compare the resolvable non-prompting forms, and treat only `keychain:` as undecidable.

```diff
-	after, ok := strings.CutPrefix(profileClientID, "env:")
-	return ok && os.Getenv(after) == invocationID
+	if after, ok := strings.CutPrefix(profileClientID, "env:"); ok {
+		return os.Getenv(after) == invocationID
+	}
+	// A file: reference is a plain read, so it can answer here. Only
+	// keychain: is withheld, because resolving it can prompt.
+	if path, ok := strings.CutPrefix(profileClientID, "file:"); ok {
+		b, err := os.ReadFile(path)
+		return err == nil && strings.TrimSpace(string(b)) == invocationID
+	}
+	return false
```

A keychain reference stays undecidable under that diff, so the CHANGELOG has to name the shape and the withheld note has to survive it. The note already gives a usable remedy, which is what keeps this off CRITICAL.

**Fixed when:** a `file:` reference naming the invocation's client ID keeps the profile's level; the CHANGELOG names the keychain shape as affected; and `TestAProfileScopeIsUsedOnlyWithThatProfilesCredentials` carries a `keychain:` row and a `file:` row. The `keychain:` row is what makes the surviving mutation fail: returning `true` from the non-`env:` branch currently leaves the whole `internal/commands` suite green.

---

⚠️ **IMPORTANT** (2) (correctness, via silent-failure-hunter) — `internal/commands/root.go:1614-1620`, `:1658`: a withheld school scope reports missing credentials

`resolveSchoolClient` records the withheld tenant ID and leaves `tid` empty. The next guard requires `tid != ""` before constructing `cliCtx.PlatformSDKClient`, so the client is never built and nothing is printed. The command then fails inside `platform.RequirePlatformClient` with "this command requires platform gateway auth", which misstates the cause: the profile and both credentials are present, and only the level was dropped. The recorded note cannot rescue it, because its only reader needs the error text to carry `REQUEST_CONTEXT_NOT_PROVIDED`, and no request is sent. Finding (1) is what makes this reachable on an ordinary profile.

**Failure scenario:** a school profile carrying `platform-url`, a keychain client-id and `tenant-id`, in a pipeline exporting `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` → `jamf-cli school blueprints list` reports "requires platform gateway auth" instead of naming the withheld tenant ID, where it previously worked.

**Suggested fix:** say what was withheld, on the one path that never reaches the gateway.

```diff
 	if platformURL != "" && cid != "" && csecret != "" && tid != "" {
 		...
 		cliCtx.PlatformSDKClient = sdk
 	}
+	if platformURL != "" && cid != "" && csecret != "" && tid == "" && withheldProfileScope.Profile != "" {
+		// No request is sent on this path, so the 400 the other two ladders
+		// rely on never arrives and the note has nowhere else to surface.
+		fmt.Fprintf(os.Stderr, "WARNING: %s\n", withheldScopeNote(nil))
+	}
```

**Fixed when:** a withheld school tenant ID produces a message naming the profile and the level, and `TestSchoolWithholdsAProfileScopeFromForeignCredentials` asserts that message beside the nil client it already pins.

---

⚠️ **IMPORTANT** (3) (correctness, via devil-advocate) — `internal/commands/platform_scope.go:476-500`: the tenant summary contradicts itself in the common case

For a tenant credential the derived `reachable` set is exactly the Jamf Security Cloud groups. Verified against the built binary: of the 29 platform resource groups carrying `scopes`, the 16 declaring `tenant` are precisely `content-categories`, `device-groups`, `dns-*`, `enrollment-activation-profiles`, `uem-*` and `ztna-*`. The `if !securityCloud` sentence at the end of the same function then disclaims every one of them. `securityCloud` never qualifies the printed list, although the function's doc comment at `:429-432` says it does.

**Failure scenario:** `jamf-cli platform setup` for an ordinary Jamf Pro tenant with no Security Cloud entitlement, so the probe answers `BAD_PERMISSIONS` → the summary prints "It also reaches 16 of the 29 Platform API resources: content-categories, device-groups, dns-custom-hostname-mappings and 13 more.", then "This tenant answered no to the Jamf Security Cloud check, so the dns-*, ztna-*, content-categories, device-groups and uem-* commands need an entitlement it lacks."

**Suggested fix:** let the probe answer qualify the partition, which is what the doc comment already promises.

```diff
 	reachable, unreachable := platformResourcesByScope(root, level)
+	if !securityCloud {
+		// A tenant credential's whole reachable set is Security Cloud, so an
+		// unentitled tenant reaches none of it. Move them, do not list and
+		// then disclaim them.
+		unreachable = append(unreachable, reachable...)
+		reachable = nil
+		sort.Strings(unreachable)
+	}
```

The wording below the list needs the same edit, since the unreachable set then holds two different reasons.

**Fixed when:** no resource appears in the reachable list and in the entitlement disclaimer, asserted by a `printScopeSummary(..., securityCloud=false)` case in `TestSetupSummarySaysWhatEachLevelActuallyReaches`.

---

⚠️ **IMPORTANT** (4) (correctness, via devil-advocate) — `internal/commands/platform_gateway_setup.go:263-270`: the ownership refusal is worded tenant-only and fires for an environment ID

`reportSecurityCloudProbe` takes `(w, err)` and cannot see which level was supplied, so its `OWNERSHIP_FORBIDDEN` branch calls the value a tenant ID and advises putting an environment ID in the environment prompt. The branch is reachable with `X-Environment-Id`: `CLAUDE.md:564` records `OWNERSHIP_FORBIDDEN` as "an environment ID for a tenant-scoped integration, or the reverse", and `CLAUDE.md:778` records that "a foreign `X-Environment-Id`" is refused by the ownership check. `printScopeSummary:452` then names the same failure an environment ID, from `creds`.

**Failure scenario:** `jamf-cli platform setup`, environment prompt answered with an environment UUID belonging to another organization → setup calls the value a tenant ID and tells the operator to use the environment prompt they just used, and the closing summary calls the same value an environment ID.

**Suggested fix:** pass the level the operator typed, which the caller already knows.

```diff
-func reportSecurityCloudProbe(w io.Writer, err error) (securityCloud, scopeIDRejected bool) {
+// level is what was typed in, so a refusal names the field the operator
+// filled rather than assuming tenant.
+func reportSecurityCloudProbe(w io.Writer, level string, err error) (securityCloud, scopeIDRejected bool) {
```

Then word the branch from `level`, so `ENVIRONMENT_NOT_FOUND` and an environment-level `OWNERSHIP_FORBIDDEN` read the same way.

**Fixed when:** an `OWNERSHIP_FORBIDDEN` probe on an environment-scoped setup names the environment ID, pinned by an `EnvironmentID` case in `TestValidatePlatformGatewayCredentials_SecurityCloudTenant`.

---

This covers all findings. Addressing the above gets this PR to merge-ready.

<details>

💡 **NICE-TO-HAVE** (5) (code-quality, via usability-reviewer) — `internal/commands/platform_scope.go:197-201`, `:331-333`: the combined note states the declared levels twice

When a withheld level is also not one the command declares, `joinNotes` concatenates two sentences that each say "this command's API declares `<levels>` scope". The earlier fix for the doubled remedy suppressed only the remedy clause. Return `""` from `scopeLevelNote` when `withheld` is true, and let the mismatch branch carry the whole note.

💡 **NICE-TO-HAVE** (6) (design, via devil-advocate+silent-failure-hunter) — `internal/commands/platform_scope.go:41`: `resolvedPlatformScope` has no reset, and a doc comment now misstates the contract

`withheldProfileScope` is reset at the top of all three ladders and `resolvedPlatformScope` is only ever written. `ResolveAuthForProfile`'s doc comment still promises it "does not read or mutate package-level variables, making it safe to call multiple times for different profiles (e.g., in the diff command)", which this PR makes false. Reset it beside its sibling, and correct the comment.

**Promote to IMPORTANT when:** `pro diff` resolves its two sides concurrently, or aggregates both sides' errors instead of returning on the first.

💡 **NICE-TO-HAVE** (7) (reliability) — `internal/commands/platform_gateway_setup.go`, `reportSecurityCloudProbe`'s `default:` arm: an inconclusive probe becomes a definite entitlement verdict

The arm returns `(false, false)` for a timeout, a 500 or a DNS failure, and `printScopeSummary`'s `if !securityCloud` sentence then states the tenant lacks a Security Cloud entitlement. The tenant did not answer no. Return a third bool for "did not complete", and say so. The control read the comment names, `GET /devices/v1/devices`, would also settle whether a suspect scope ID is really rejected before the summary is suppressed.

💡 **NICE-TO-HAVE** (8) (documentation, via fidelity-reviewer) — `CLAUDE.md:286`: the Classic write-coverage counts are stale in the sentence this PR rewrote

The tail gained "binds again as of the v2082 ingest" while the head still reads "43 of 54 manifest resources bind a schema; 111 of 117". `CHANGELOG.md:172` and `docs/guides/platform-api-ga.md:608` both say 44 of 54 and 114 of 117, and `specs/classic/schemas.json` agrees with them.

💡 **NICE-TO-HAVE** (9) (documentation, via fidelity-reviewer) — the PR description names a flag that does not exist, and overstates the spec coverage

"if the client ID comes from `--client-id` or `JAMF_CLIENT_ID`" names no flag this CLI has, and `platform_scope.go:215-218` says so. "v2082 declared `x-scope-types` on every Platform spec" excludes the three `account_*.json` specs, which declare none; `platform_scope.go:222-226` says so. The CHANGELOG has both right.

</details>

<details>
<summary>PR decomposition (via scope-reviewer)</summary>

**Recommendation: split into 2 PRs.** The ingest and the scope-*reporting* work are one artifact: they fall out of parsing one SDK drop, land in the same regenerated files, and the reporting half is additive and advisory. The scope-*withhold* change is architecturally independent. It touches only `internal/commands/root.go`, `pro_platform_helpers.go` and `platform_scope.go`, needs no spec or generated-code change, and carries its own test file. It arrived as its own commit (`a7f3c16`) and applies cleanly as a follow-up.

The PR's own argument, that the ingest forced each concern, holds for the ingest and the reporting and not for the withhold. That rewrite was triggered by a side effect of the new annotation making a pre-existing splice visible, and the fix stands on its own.

**PR 1 — SDK v0.22.0 ingest, Classic patch restoration, device-groups v2, declared-scope reporting.** The spec drop, `internal/gateway/coverage_gen.go`, all regenerated command files, `generator/parser/platform.go`, `generator/parser/types.go`, `generator/platform/emitter.go`, the `commandEntry.Scopes` projection, `AnnotateScopeLevelError`, `scopeMismatchHint`, and the retirement of `annotateAuditScopeError`. Merge first.

**PR 2 — withhold a profile's scope level from a foreign credential (breaking).** The `ResolveAuthForProfile`, `resolveScope` and `resolveSchoolClient` rewrites, the withheld-note machinery, `platform_scope_withhold_test.go`, and the setup probe verdict work. Depends on PR 1, whose annotation and error scaffolding it renders through.

`f57a221` spans both concerns and would need cherry-picking by function rather than by commit, which its own commit message already separates by paragraph. Do not split PR 1 further: the patch restoration and the scope reporting are read off the same regenerated specs.

One reviewer's counterpoint, for the record. The commits are already clean and separable, the branch is green, and splitting now costs a rebase plus a second review pass on work that has had one. Splitting is the better default for the next change of this shape rather than an argument to reopen this one.

</details>

<details>

- [x] Design and architecture: reporting-not-refusing, derived summary, one recording site
- [x] Correctness: the withhold predicate (1), (2), the summary partition (3), (4)
- [x] Security: scope-header provenance, no secret newly resolved or printed
- [x] Test coverage: 3 of 5 mutations survived, see (1), and (5) in the dropped set
- [x] Reliability: probe branches (7), package-var lifetime (6)
- [x] Documentation currency: CHANGELOG, `docs/guides/platform-api-ga.md`, CLAUDE.md (8)
- [x] Cross-repo contracts: `scopes` catalog key confirmed on 110 rows in the built binary
- [x] Usability: error notes, setup summary, two documented breaking changes
- [x] Project rules compliance: `CLAUDE.md` credential and generated-code policies
- [?] Correctness of the v2 device-group PUT: 204 rests on the author's probe, not verified here
- [na] Performance, frontend.

Diff: 75 files, +3560/−740, head `cfa4571`. Verified locally: `go build ./...`, `go test ./...`, `make lint` (0 issues), `verify-generated`, `verify-gateway-coverage`, `verify-classic-schemas`, `verify-platform-specs`, `verify-site` all clean; refusal count 75 → 67 and the `scopes` key on 110 commands both confirmed by diffing the built binaries. Lanes run: security-reviewer, usability-reviewer, fidelity-reviewer, scope-reviewer, silent-failure-hunter, devil-advocate, test-quality-reviewer. Never run: performance-reviewer, inapplicable.

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `generator/parser/platform.go` | 5/5 | `scopeTypes` validates against a closed allowlist, warns on every malformed shape, and normalises order; `TestScopeTypes` covers absent, scalar, empty, miscased and duplicate |
| `internal/commands/platform_scope.go` | 3/5 | The derivation and the never-enforce decision are right and well evidenced; three of four blocking findings live here, see (1), (3) |
| `internal/commands/platform_gateway_setup.go` | 3/5 | Splitting the probe out made its branches testable, and both new verdicts are pinned; the wording is level-blind, see (4) |
| `internal/commands/root.go` | 4/5 | The withhold reads `params.ClientID` only and the resolved profile name, both correct; the school ladder is the one path with no signal, see (2) |
| `internal/commands/privilege_error.go` | 5/5 | The 403 keeps its two causes separate and adds the declared levels without claiming the credential's own |
| `specs/gateway/coverage.json` | 4/5 | Regenerated, provenance recorded at `94cbc28`; the un-refused eight are confirmed against the binary, and the v2 PUT's 204 rests on the author's probe |

</details>

<details>

✅ Reading `x-scope-types` and refusing to enforce it. The spec is stricter than the gateway, the PR says so in the code rather than only in the description, and it names the two commands a refusal would have broken. That is the same judgement `platformUnroutedOps` was built on, applied in the opposite direction.

✅ `TestEveryCommandEntryFieldReachesTheCatalog` fixes the class of bug rather than the instance. A hand-maintained projection lost `scopes` silently, and had lost `gatewaySuccessor` before it. Reflecting the struct's json tags against the shipped tree means the next field cannot go the same way.

✅ `TestRunAnnotatesAScopeLevelError` drives the whole `run()` sequence rather than the annotator alone, which is what pins the `%w` wrap that keeps the exit-code classification alive.

</details>



ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Re-review of `15e033a`. All four prior blocking findings, all five nice-to-haves and both description inaccuracies are addressed, and the v0.22.1 ingest is clean.
> Blocking: (1) to (5), each of them a consequence of a fix rather than a return of the old defect. See the collapsed sections.

## Rating: 3/5
- Would be a 5 once the notes stop asserting what the gateway would have done (1), (2), the entitlement verdict stops speaking for 16 resources on one 403 (3), and the third probe verdict reaches its consumer (4).

The direction of the fix commit is right, and the reasoning in it is checkable. What went wrong is uniform: each fix replaced a vague statement with a specific one, and four of the specific ones claim more than the code knows. That is the same failure the PR's own "declares, never requires" discipline exists to prevent, now inside the messages rather than in the refusal logic.

<details>
<summary>Prior findings status (9 items)</summary>

| # | Location | State | Note |
|---|---|---|---|
| (1) `c=100` | `platform_scope.go:117` | ✅ Fixed | `file:` resolved and compared, `keychain:` withheld with the CHANGELOG naming the shape, and both test rows added. The keychain row kills the mutation that survived last round |
| (2) `c=75` | `root.go:1614` | ⚠️ Partial | The note now surfaces through `platform.ErrNoPlatformClient` and names the profile and the level. The remedy it gives is unusable on that path, see (2) |
| (3) `c=100` | `platform_scope.go:476` | ⚠️ Partial | The probe answer partitions the list and no resource is listed and disclaimed. The sentence introduced with it over-claims, see (3) |
| (4) `c=100` | `platform_gateway_setup.go:263` | ✅ Fixed | `level` is a parameter, and the setup test carries an environment-level row |
| (5) `c=75` | `platform_scope.go:197` | ✅ Fixed | `withheldState` replaces the bool, and all three arms are asserted directly |
| (6) `c=75` | `platform_scope.go:41` | ⚠️ Partial | `resetPlatformScopeRecords` clears both, and the doc comment is corrected. One path named in its own comment skips it, see (6) |
| (7) `c=50` | `platform_gateway_setup.go` | ⚠️ Partial | `securityCloudVerdict` has three values. Only the negative half of the over-claim was fixed, see (4) |
| (8) `c=100` | `CLAUDE.md:286` | ✅ Fixed | 44 of 54 and 114 of 117, recounted from the binary |
| (9) `c=100` | PR description | ✅ Fixed | `--client-id` removed, and the three account specs named as the exception |

</details>

---

⚠️ **IMPORTANT** (1) (correctness, via devil-advocate) — `internal/commands/platform_scope.go:243-248`: the note asserts what the gateway would have done, and drops the remedy that works

`withheldScopeNote`'s disagreeing branch says the withheld ID "would not have worked here anyway, since this command's API declares X. Use an integration created at a declared level." That is a *requires* claim. `AnnotateScopeLevelError`'s own doc comment 80 lines below says the opposite discipline applies, and `CLAUDE.md:438` records the counter-example: probed 2026-09-05, a tenant credential answers 200 on `pro platform-devices list`, which declares environment scope. The branch also returns before the remedy paragraph, so the withheld level's own env var and flag are never printed. The branch predates this commit; the fixes made it the operator's only output on two paths.

**Failure scenario:** a gateway profile with `client-id: keychain:gw/client-id` and an owned `tenant-id`, in a shell exporting `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` for that same integration → `jamf-cli -p gw pro platform-devices list` earns `400 REQUEST_CONTEXT_NOT_PROVIDED`, and the note says the tenant ID would not have worked and to obtain a new integration, when `JAMF_TENANT_ID=<the same value>` restores the 200 this repo recorded.

**Suggested fix:**
```diff
-		return fmt.Sprintf("Profile %q carries %s, which was not used because the client ID came "+
-			"from %s — and would not have worked here anyway, since this command's API declares %s. "+
-			"Use an integration created at a declared level.",
-			withheldProfileScope.Profile, withheldLevelPhrase(), clientIDSource, renderScopeLevels(levels))
+		return fmt.Sprintf("Profile %q carries %s, which was not used because the client ID came "+
+			"from %s. This command's API declares %s, so an integration created at a declared level "+
+			"is the durable answer; the gateway has not followed the specs everywhere, so %s or %s "+
+			"may still work today.",
+			withheldProfileScope.Profile, withheldLevelPhrase(), clientIDSource,
+			renderScopeLevels(levels), envVar, flagName)
```

**Fixed when:** no note in this file states what the gateway would have done with a level it did not send, and every branch that reaches an operator names the withheld level's own env var and flag.

---

⚠️ **IMPORTANT** (2) (correctness, via devil-advocate+test-quality-reviewer) — `internal/commands/platform_scope.go:347-352`: on the school path the new note recommends the one thing that cannot work

The `ErrNoPlatformClient` arm passes `scopesOf(cmd)`. For `school blueprints list`, the command declares `environment` (confirmed from the built binary) and the withheld level is `tenant`, so the branch in (1) renders and advises an integration created at a declared level. `resolveSchoolClient` reads only `JAMF_TENANT_ID` and the profile's `tenant-id` and builds `auth.TenantScope`. It has no environment code path at all, so an operator who follows that advice returns to the identical error. The working input is a tenant ID, which is the one string the branch omits.

The reason this reads as verified is a branch difference. `TestSchoolWithholdsAProfileScopeFromForeignCredentials` calls `AnnotateScopeLevelError(nil, ...)` with a nil command, so `scopesOf(nil)` is nil and the *other* branch renders, which is why its assertion on `JAMF_TENANT_ID` passes. `cmd/jamf-cli/main.go:93` passes the real command. Replacing `scopesOf(cmd)` with `nil` survives the whole suite.

**Failure scenario:** a school profile with `platform-url`, a keychain client-id, a client secret and `tenant-id`, in a CI shell exporting `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` → `jamf-cli school blueprints list` prints a remedy that cannot produce a platform client on that path.

**Suggested fix:** this arm sent no request, so it has no opinion on whether the declared level would have been accepted, and school's resolver accepts only one level.
```diff
 	if errors.Is(err, platform.ErrNoPlatformClient) {
-		if withheld := withheldScopeNote(scopesOf(cmd)); withheld != "" {
+		// No request was sent, so nothing here can say the declared level
+		// would have been accepted, and school's resolver builds a
+		// tenant-scoped client only. Offer the withheld level back.
+		if withheld := withheldScopeNote(nil); withheld != "" {
 			return fmt.Errorf("%w\n\nnote: %s", err, withheld)
 		}
```

**Fixed when:** `school blueprints list` with a withheld tenant level prints a remedy that produces a working platform client when applied literally, asserted through the same `*cobra.Command` `main.go` passes rather than through `nil`.

---

⚠️ **IMPORTANT** (3) (correctness, via devil-advocate) — `internal/commands/platform_gateway_setup.go:311-315`: one 403 now removes 16 resources and states a licensing verdict

`reportSecurityCloudProbe` classifies a single `403 BAD_PERMISSIONS` on `GET /securitycloud/v1/categories` as `securityCloudUnentitled`, and this commit promoted that verdict from a trailing sentence into a structural subtraction: 16 groups leave the reachable list and the summary prints "N are Jamf Security Cloud, which the check above says this scope is not entitled to". This repo's own rule forbids the inference. `CLAUDE.md:551` says "one 403 is not a probe", and `internal/gateway/note.go:154` records `BAD_PERMISSIONS` as indistinguishable from a missing privilege. The third cause the code cannot see is a per-capability grant: Jamf Account grants one capability at a time, so `categories:read` withheld on an entitled tenant returns the same code, and the recorded `/devices/v1/devices` control proves device grants rather than `categories:read`.

**Failure scenario:** a Security Cloud entitled tenant whose integration was created without `content-categories:read` but with the ztna grants → `platform setup` drops all 16 groups and says the scope is not entitled to them, while `jamf-cli security ztna-apps list` on that profile returns data. The operator opens a licensing conversation instead of ticking one box.

**Suggested fix:** keep the subtraction, and say only what one 403 establishes.
```diff
-		_, _ = fmt.Fprintf(w, "%d are Jamf Security Cloud, which the check above says this scope is not\n", len(unentitled))
-		_, _ = fmt.Fprintln(w, "entitled to:")
+		_, _ = fmt.Fprintf(w, "%d are Jamf Security Cloud. The one read above was refused 403 for this\n", len(unentitled))
+		_, _ = fmt.Fprintln(w, "credential, which is either no Security Cloud entitlement or a missing")
+		_, _ = fmt.Fprintln(w, "capability grant. The gateway spells both the same way, so check the")
+		_, _ = fmt.Fprintln(w, "integration's permissions in Jamf Account before assuming licensing:")
```

**Fixed when:** no sentence printed from a single `BAD_PERMISSIONS` names a cause the code cannot distinguish, and `securityCloudUnentitled`'s doc comment names the missing-grant alternative beside the missing-entitlement one.

---

⚠️ **IMPORTANT** (4) (correctness, via test-quality-reviewer+silent-failure-hunter) — `internal/commands/platform_scope.go:648`: the new third verdict is inert at its only consumer

`printScopeSummary` branches on `securityCloud == securityCloudUnentitled`, so `securityCloudUnknown` takes the same path as a confirmed 200. A probe that timed out, 5xx'd or failed DNS therefore yields "It also reaches all N Platform API resources, audit and AI Governance included", over 16 groups whose entitlement nothing observed. Only the negative half of the over-claim the third value was added to fix is actually fixed. No test reaches this branch with `Unknown`: every case passes `Entitled` or `Unentitled`, or passes `Unknown` with `scopeIDRejected=true`, which returns earlier. Collapsing the two verdicts back together survives the whole suite.

**Failure scenario:** `platform setup` on a tenant whose one Security Cloud read times out → the closing summary, which is the artifact the operator keeps, claims the profile reaches all 29 Platform API resources.

**Suggested fix:**
```diff
+	if securityCloud == securityCloudUnknown {
+		_, _ = fmt.Fprintln(w, "The Jamf Security Cloud check did not complete, so whether the dns-*, ztna-*,")
+		_, _ = fmt.Fprintln(w, "content-categories, device-groups and uem-* commands are entitled is unknown.")
+	}
```

**Fixed when:** a `printScopeSummary` case with `(securityCloudUnknown, false)` at tenant level asserts the entitlement is reported as unknown, and flipping the condition to `!= securityCloudEntitled` fails a test.

---

⚠️ **IMPORTANT** (5) (correctness, via silent-failure-hunter) — `internal/commands/platform_scope.go:156-161`: the `file:` read error is discarded, and the note then blames the wrong cause

The new branch answers "different integration" for a missing path, a permission error or an unmounted volume, so the profile's level is withheld and `withheldScopeNote` says it "was not used because the client ID came from the JAMF_CLIENT_ID environment variable". That is not why. The comment justifying the discard says such a profile "is going to fail with a better message further along", and that is false on the only path that reaches this branch: it is reachable only when the invocation supplied a client ID, and in exactly that case both ladders skip `config.ResolveSecret(p.ClientID)`, so nothing downstream reads the file and the error surfaces nowhere.

**Failure scenario:** a profile with `client-id: file:/run/secrets/client-id` and a `tenant-id`, invoked with `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` exported for that same integration, in a container where the secrets path is not mounted → the read fails silently, the level is withheld, and the note blames the environment variable. With the file readable the same invocation keeps its level and succeeds.

**Suggested fix:**
```diff
 	if path, ok := strings.CutPrefix(profileClientID, "file:"); ok {
 		b, err := os.ReadFile(path)
-		// An unreadable file answers no rather than erroring: this is a
-		// same-integration test, and a profile whose client ID cannot be read
-		// is going to fail with a better message further along.
-		return err == nil && strings.TrimSpace(string(b)) == invocationID
+		// Nothing downstream reads this file: when the invocation supplied a
+		// client ID, both ladders skip ResolveSecret(p.ClientID), so this is
+		// the only place the failure is observable.
+		if err != nil {
+			recordUnreadableClientIDRef(path, err)
+			return false
+		}
+		return strings.TrimSpace(string(b)) == invocationID
 	}
```
Bound the read while you are here: `os.ReadFile` on a config-named FIFO blocks in `open()`, and on `/dev/zero` it grows without limit, because `Stat` reports size 0.

**Fixed when:** an unreadable `file:` client-id reference produces a signal naming the path and the read error, rather than reading as a profile that names a different integration.

---

This covers all findings. Addressing the above gets this PR to merge-ready.

<details>

💡 **NICE-TO-HAVE** (6) (design, via silent-failure-hunter) — `internal/commands/pro_platform_helpers.go:493`: the new reset misses one of the two paths its own comment names

`resetPlatformScopeRecords`' doc comment names "a profile with no credentials, or school's tenant-less path". School resets, and so does `ResolveAuthForProfile`, but `securityPlatformSDKClient` returns `nil, nil` before the `resolveScope` call that is the only reset on that path. Call it at the top of that function instead.

**Promote to IMPORTANT when:** `AnnotateScopeLevelError` becomes reachable from a process that resolves per request, such as a long-lived `mcp serve`.

💡 **NICE-TO-HAVE** (7) (code-quality, via devil-advocate) — `cmd/jamf-cli/main.go:87-92`: the ordering comment's premise is now false

It says "AnnotateScopeLevelError fires only on a 400 carrying REQUEST_CONTEXT_NOT_PROVIDED". The new `ErrNoPlatformClient` arm fires with no gateway response at all. Disjointness still holds, but the sentence a future reader would check it against does not describe the function. Enumerate both triggers.

💡 **NICE-TO-HAVE** (8) (test-coverage, via test-quality-reviewer) — `internal/commands/root.go:1319`: the `scopes` positive-only guard is unpinned for a second round

Making the assignment unconditional leaves the whole suite green, and the binary then emits `"scopes": null` on 1646 rows, which is the reading the code comment exists to prevent. `TestCatalogJSONCarriesTheSuccessorKey` pins both directions for `gatewaySuccessor`; the same case can cover `scopes` and the sibling positive-only keys.

</details>

<details>

- [x] Design and architecture: the three-state enum and the derived Security Cloud set
- [x] Correctness: the note branches (1), (2), the probe verdicts (3), (4), the `file:` read (5)
- [x] Security: the new filesystem read on the auth path, no widening of the fail-closed 403
- [x] Test coverage: 17 mutations, 12 killed; the survivors are (2), (4), (8)
- [x] Reliability: the reset coverage (6), the probe's default arm (4)
- [x] Documentation currency: CHANGELOG, CLAUDE.md and the PR body all corrected
- [x] Cross-repo contracts: the v0.22.1 ingest moves one Classic element name, and the generated scaffold matches
- [x] Project rules compliance: `CLAUDE.md` credential and generated-code policies
- [na] Performance, frontend.

Diff since the reviewed `cfa4571`: 25 files, +2346/−189, head `15e033a`, of which the merge from `main` (PR #346) is the bulk and the new work is 2 commits, +739/−125. Verified locally: `go build ./...`, `go test ./...` (34 packages, all pass), `make lint` (0 issues), and `verify-generated`, `verify-gateway-coverage`, `verify-classic-schemas`, `verify-platform-specs`, `verify-site` all clean. Lanes run this round: devil-advocate, security-reviewer, silent-failure-hunter, test-quality-reviewer. Not re-dispatched: fidelity-reviewer, scope-reviewer, usability-reviewer, all three having run in round 1 on unchanged ground. Never run: performance-reviewer, inapplicable.

</details>

<details>

✅ The fix commit reproduced each finding before fixing it, and says so per finding. The `keychain:` case is the sharpest: it is named as a cost the fix does not remove, with the CHANGELOG and the note carrying the remedy instead of the code pretending to solve it.

✅ `securityCloudResourceGroups` is read off the command tree rather than from "declares tenant", and the commit explains why the two only coincide by accident. An environment credential declares all 29, so a split on the declared level would have left that case claiming the whole surface.

✅ The v0.22.1 ingest is one property in one Classic schema, and the reasoning separates what it cost this CLI (a scaffold) from what it cost the SDK (a silent read-side drop). `CLAUDE.md`'s repeated-element note gains the part the incident proves: reading the element name off the schema is necessary and not sufficient.

</details>



neilmartin83:
@ktn-jamf ready for re-review — `1a20778`.

All five blocking findings and all three nice-to-haves are addressed, and the branch also picks up **SDK v0.22.2** (`d4ab4f4`). The PR description carries the detail; the short version:

**Your diagnosis was the useful part.** Each round-1 fix swapped a vague statement for a specific one and four of the specific ones claimed more than the code knows — the "declares, never requires" discipline failing inside the messages instead of in the refusal logic. Every fix here is a subtraction from a claim, not a new mechanism.

- **(1)** The "would not have worked here anyway" branch is gone, and it no longer returns before the remedy: the note ranks the declared level and the durable answer first, then names the withheld level's own env var and flag as what *may still work today*. `withheldReasonClause` factors the shared opening so the two branches cannot drift apart again.
- **(2)** The `ErrNoPlatformClient` arm passes `nil`. No request was sent, so it has no opinion on the declared level, and school's resolver builds tenant scope only. The new test goes through the same `*cobra.Command` `main.go` passes — and asserts the premise (`school blueprints list` declares environment without tenant) so it cannot go vacuous the way the nil-argument sibling did.
- **(3)** Right on all three counts, including that the `/devices/v1/devices` control proves the device grants rather than `content-categories:read`. The subtraction stays; the claim does not. The clause names both causes, the probe line reads "no — refused (no entitlement, or the integration lacks this permission)", and `securityCloudUnentitled`'s doc comment names the alternative.
- **(4)** `securityCloudUnknown` reports the entitlement as unknown. Both mutations you named fail: collapsing the verdicts fails the unknown assertion, widening to `!= securityCloudEntitled` fails the "must not subtract" assertion in the same test.
- **(5)** The sharpest one — the justification was false for exactly the reason you give, since both ladders skip `ResolveSecret(p.ClientID)` on the only path that reaches the branch. Recorded and named now, path and read error. The read is bounded as suggested: non-regular files are refused before `open()`, so a config-named FIFO cannot block the invocation.
- **(6)** `securityPlatformSDKClient` resets at the top; removing the call fails the existing reset test. **(7)** `main.go` enumerates both triggers. **(8)** `TestCatalogJSONOmitsTheEmptyPositiveOnlyKeys` pins all four positive-only keys in both directions, judged against the entry field the projection was handed.

**Wire-checked on an EU environment-scoped credential**, one run each: a matching `file:` client-id keeps its level and answers 200; a foreign one earns the 400 plus the profile/level/env-var note; an **unreadable** one earns the 400 plus a note naming the path and the read error with no mention of `JAMF_CLIENT_ID`; a withheld tenant level on `pro platform-devices list` gets the ranked note; and `school blueprints list` names `JAMF_TENANT_ID`, which when supplied reaches `403 OWNERSHIP_FORBIDDEN` on a deliberately bogus tenant — i.e. the remedy produces a platform client, which is finding (2)'s "fixed when". The two verdict findings are unit-tested rather than probed: this credential *is* Security Cloud entitled, so neither branch is reachable on it.

**v0.22.2** is one `schemaPatches` entry upstream. `mobile_device_configuration_profile`'s Self Service category `$ref`'d the shared `category` schema where its five siblings declare the item inline with `display_in` — and that field decides whether the category is stored **at all**: a `<category>` with only `<id>`, or `<id>` + `<name>`, is accepted with 201 and silently discarded. Wire-verified across all six resources carrying the block, so the law belongs to the relation. What is unique here is that `display_in` is **write-only on this resource alone**, so the resource whose spec omitted it is the one whose read surface conceals it. `pro classic-mobile-config-profiles` writes gain the field in `--scaffold`, `--set` and help — and the scaffold renders `<display_in>false</display_in>`, the discard value, which puts it alongside `general.category.id: 0` as a rendered value that is guaranteed wrong rather than merely incomplete. No platform spec moved.

`make test`, `make lint` (0 issues) and all five verify targets clean; `make generate` idempotent.

ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Round 3 of `1a20778`. Six of the eight prior findings are fixed and verified by re-mutation, two are partial, and the v0.22.2 ingest is clean.
> Blocking: (1) to (4), all four inside the two code paths this round added. See the collapsed sections.

## Rating: 3/5
- Would be a 5 once the unreadable-reference note stops contradicting itself and stops advising the one change that breaks resolution (1), the unknown sentence is derived rather than hand-listed (2), and the two silent read outcomes are recorded (3).

Every claim the fix commit makes about its own tests holds. Ten mutations were run against it and eight were killed, including all four the author named. Round 2's diagnosis landed: the notes no longer assert what the gateway would have done. What is left is narrower and of one kind. The two paths added this round, the unreadable-reference clause and the unknown-entitlement sentence, each carry one instance of the defect they were written to remove.

<details>
<summary>Prior findings status (8 items)</summary>

| # | Location | State | Note |
|---|---|---|---|
| (1) `c=75` | `platform_scope.go:243` | ✅ Fixed | The requires-claim is gone, the durable answer is ranked first, and the withheld level's env var is named. `withheldReasonClause` factors the opening |
| (2) `c=75` | `platform_scope.go:347` | ✅ Fixed | The arm passes `nil`. Mutating it back fails a test, and the new test asserts its own premise so it cannot go vacuous |
| (3) `c=100` | `platform_gateway_setup.go:311` | ✅ Fixed | Both causes named in the clause, the probe line and the verdict's doc comment |
| (4) `c=75` | `platform_scope.go:648` | ⚠️ Partial | `Unknown` now acts, and both named mutations fail. The sentence it prints is hand-listed and incomplete, see (2); the claim above it stays unqualified, see (5) |
| (5) `c=75` | `platform_scope.go:156` | ⚠️ Partial | The unreadable case is recorded and named, and the read is bounded. Two other silent outcomes remain, see (3), and the note it feeds contradicts itself, see (1) |
| (6) `c=75` | `pro_platform_helpers.go:493` | ✅ Fixed | Resets at the top; removing the call fails a test |
| (7) `c=75` | `main.go:87` | ✅ Fixed | Both triggers enumerated |
| (8) `c=75` | `root.go:1319` | ✅ Fixed | `TestCatalogJSONOmitsTheEmptyPositiveOnlyKeys` pins all four keys in both directions, judged against the entry field the projection was handed. Three mutations killed, and its non-vacuity checked. Closed after three rounds |

</details>

---

⚠️ **IMPORTANT** (1) (correctness, via devil-advocate) — `internal/commands/platform_scope.go:328-332`: the unreadable-reference note asserts what it just disclaimed, and its remedy makes the failure worse

On the agreeing branch, `withheldReasonClause` says "nothing here could tell whether it names the same integration as the credentials in hand", and the next sentence asserts "that ID belongs to the profile's own integration". That is the proposition just declared undecidable, and `profileNamesTheInvocationClientID` returning false here is a fail-closed default rather than a determination. The trailing remedy is worse than merely inconsistent: "use the profile's own client ID as well as its level" means unsetting `JAMF_CLIENT_ID`, and with it gone `ResolveAuthForProfile` runs `config.ResolveSecret(p.ClientID)`, opens the same unreadable file and returns a hard "resolving client-id from profile" error. The note never names the file as the thing to fix, which is the one remedy the newly recorded path and error uniquely enable.

**Failure scenario:** a platform profile with `client-id: file:/run/secrets/cid` whose mount is absent, plus `JAMF_CLIENT_ID` and `JAMF_CLIENT_SECRET` exported → `pro categories list` earns the 400 and this note → the operator follows the trailing clause and unsets `JAMF_CLIENT_ID` → the next run fails outright with "resolving client-id from profile", worse than what it replaced.

**Suggested fix:** branch the remedy on the reason, since the reason is already recorded.
```diff
+	if unreadableClientIDRef.Path != "" {
+		// The comparison was never made, so no claim about whose integration
+		// the ID names may be printed. And dropping JAMF_CLIENT_ID makes
+		// ResolveSecret open the same unreadable file and fail outright.
+		return fmt.Sprintf("Profile %q carries %s, %s Supply the level for these credentials with "+
+			"%s or %s, or make file:%s readable so the profile's own client ID resolves. Do not drop "+
+			"%s while that file is unreadable: the profile cannot then resolve a client ID at all.",
+			withheldProfileScope.Profile, withheldLevelPhrase(), withheldReasonClause(),
+			envVar, flagName, unreadableClientIDRef.Path, clientIDSource)
+	}
 	return fmt.Sprintf("Profile %q carries %s, %s An integration is created at one level in Jamf "+
```

**Fixed when:** the unreadable-reference branch makes no claim about which integration the profile's ID names, names the unreadable path as the thing to fix, does not advise dropping `JAMF_CLIENT_ID`, and a test renders that exact combination and asserts both properties.

---

⚠️ **IMPORTANT** (2) (correctness, via devil-advocate) — `internal/commands/platform_scope.go:823-828`: the new unknown sentence is hand-listed, and it already omits one of the sixteen

The sentence names five literal patterns, `dns-*`, `ztna-*`, `content-categories`, `device-groups` and `uem-*`, while calling `securityCloudResourceGroups` on the line above only for the `anyIn` guard and discarding its contents. It is incomplete as shipped. I confirmed against the built binary that 16 groups under `security` carry `scopes`, and that filtering them by those five patterns leaves exactly one unnamed: `enrollment-activation-profiles`, wired at `security.go:67` with `jamf:scopes: environment,tenant`. Three doc comments in this file argue against a hand-written copy of this set, and the adjacent unentitled branch subtracts the same group correctly because that list is derived.

**Failure scenario:** `platform setup` on a tenant credential whose one Security Cloud read times out → the summary names five patterns as unknown and leaves `enrollment-activation-profiles` standing as an unqualified reach, which is the sixteenth group the third verdict exists to stop over-claiming about.

**Suggested fix:** render it from the intersection the function already computes.
```diff
 	if securityCloud == securityCloudUnknown {
-		if sc := securityCloudResourceGroups(root); anyIn(reachable, sc) {
-			_, _ = fmt.Fprintln(w, "The Jamf Security Cloud check did not complete, so whether the dns-*, ztna-*,")
-			_, _ = fmt.Fprintln(w, "content-categories, device-groups and uem-* commands are entitled is unknown.")
+		sc := securityCloudResourceGroups(root)
+		var unknown []string
+		for _, g := range reachable {
+			if sc[g] {
+				unknown = append(unknown, g)
+			}
+		}
+		if len(unknown) > 0 {
+			_, _ = fmt.Fprintf(w, "The Jamf Security Cloud check did not complete, so whether these %d are\n", len(unknown))
+			_, _ = fmt.Fprintf(w, "entitled is unknown:\n  %s.\n", summariseResources(unknown))
 		}
 	}
```
That also retires `anyIn`, whose guard has no reachable false case in the shipped tree and which survives being replaced with `true`.

**Fixed when:** the unknown sentence is rendered from `securityCloudResourceGroups` intersected with `reachable`, and a test asserts `enrollment-activation-profiles` is covered by it, which fails today.

---

⚠️ **IMPORTANT** (3) (correctness, via silent-failure-hunter) — `internal/commands/platform_scope.go:229-233`: two `file:` reads succeed and record nothing, so the note blames the variable again

`readClientIDRefFile` returns a nil error for two inputs whose value cannot be the profile's client ID. An empty regular file passes `IsRegular`, `io.ReadAll` returns no error, and `TrimSpace("")` compares false against a guaranteed non-empty invocation ID. A file over `maxClientIDRefBytes` reaches EOF on the `io.LimitReader` with no error, so a truncated prefix compares false. Neither sets `unreadableClientIDRef`, so `withheldReasonClause` takes its else branch and the note says the level was not used "because the client ID came from the JAMF_CLIENT_ID environment variable". That is the misattribution the record was added this round to remove, on the neighbouring input. Zero length is what an interrupted write, a touched placeholder, or a secrets volume mounted ahead of its content leaves.

**Failure scenario:** a profile with `client-id: file:/mnt/secrets/client-id` at 0 bytes and a `tenant-id`, in a shell exporting `JAMF_CLIENT_ID` for that same integration → the level is withheld, `pro platform-devices list` earns `400 REQUEST_CONTEXT_NOT_PROVIDED`, and the note sends the operator to drop the one variable that is correct.

**Suggested fix:**
```diff
-	b, err := io.ReadAll(io.LimitReader(f, maxClientIDRefBytes))
+	b, err := io.ReadAll(io.LimitReader(f, maxClientIDRefBytes+1))
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(string(b)), nil
+	if len(b) > maxClientIDRefBytes {
+		return "", fmt.Errorf("larger than %d bytes, so it cannot be compared", maxClientIDRefBytes)
+	}
+	id := strings.TrimSpace(string(b))
+	if id == "" {
+		return "", errors.New("empty")
+	}
+	return id, nil
```

**Fixed when:** an empty or over-cap `file:` reference returns an error, so the record is set and the note names the path and the reason; the existing over-cap test asserts that error rather than a 4096-byte success.

---

⚠️ **IMPORTANT** (4) (test-coverage, via test-quality-reviewer) — `internal/commands/platform_scope.go:221`: the `IsRegular` refusal survives its own test

Neutralizing the branch leaves `TestAClientIDReferenceThatIsNotARegularFileIsRefused` and the whole `internal/commands` package green. The only non-regular input the test supplies is a directory, and with the guard gone `os.Open` succeeds and `io.ReadAll` returns `EISDIR`, so the test still sees a non-nil error while asserting nothing about the guard. The input the guard exists for is a FIFO, which blocks inside `os.Open` and returns no error at all, and it is the one case a downstream read cannot stand in for. The guard is correct; nothing holds it in place.

**Failure scenario:** the guard is later removed or refactored past, and a profile carrying `client-id: file:/path/to/fifo` hangs every invocation that resolves it while `JAMF_CLIENT_ID` is set, with no timeout and no output, and no test fails.

**Suggested fix:** assert on an input a downstream read error cannot answer.
```diff
+	// A FIFO is the input the guard exists for: os.Open blocks on it and
+	// never returns an error, so a downstream read cannot stand in.
+	fifo := filepath.Join(dir, "fifo")
+	if err := syscall.Mkfifo(fifo, 0o600); err != nil {
+		t.Skipf("mkfifo unsupported here: %v", err)
+	}
+	if _, err := readClientIDRefFile(fifo); err == nil {
+		t.Error("a named pipe must be refused before open()")
+	}
```
Cheaper and nearly as good: assert the directory case's error text contains `not a regular file`, which can only come from the guard.

**Fixed when:** neutralizing `!info.Mode().IsRegular()` makes a test in `./internal/commands/` fail.

---

This covers all findings. Addressing the above gets this PR to merge-ready.

<details>

💡 **NICE-TO-HAVE** (5) (code-quality, via devil-advocate) — `internal/commands/platform_scope.go:768-778`: the unqualified claim still prints first, and the qualification is appended after it

"It also reaches all N Platform API resources, audit and AI Governance included" prints verbatim on the unknown path, because it is reached whenever the unreachable and unentitled sets are both empty and only `securityCloudUnentitled` populates the latter. A reader who stops early gets the strongest sentence, which is the unverified one. That is the shape the adjacent unreachable clause's own comment says it was moved to fix: "Beside the list it is about, not after both of them."

**Promote to IMPORTANT when:** the qualification can fail to print while the claim still does, which finding (2)'s fix would otherwise make possible.

💡 **NICE-TO-HAVE** (6) (code-quality, via silent-failure-hunter) — `internal/commands/platform_scope.go:768-778` against `platform_gateway_setup.go:241-248`: the entitled verdict now claims more than the corrected unentitled wording allows

The unentitled branch was corrected this round to say one `BAD_PERMISSIONS` cannot separate no entitlement from a missing capability grant. `securityCloudEntitled` rests on the same single `content-categories` read, and the summary then lists all 16 Security Cloud groups unqualified. Improving one branch exposed the asymmetry rather than creating it, and per-command grants do surface at runtime, which is why this is not blocking.

💡 **NICE-TO-HAVE** (7) (usability, via devil-advocate) — `internal/commands/pro/generated/classic_mobile_config_profiles.go:186-190`: the scaffold ships the discard value pre-filled, on the resource whose read surface hides the loss

The v0.22.2 ingest adds `display_in`, and because the schema gives the property a description but no `example` (unlike its `id` and `name` siblings), `ScaffoldXML` renders `<display_in>false</display_in>`, which that same description says discards the category. The commit likens it to `general.category.id: 0`, but that answers 409 while this is accepted with 201 and silently dropped, and this resource's GET echoes only `id` and `name`, so a read-back cannot tell stored from discarded. `classicBodySpec` carries no description channel, so the law reaches `specs/`, CLAUDE.md and the CHANGELOG and no `--help`. An `example: true` upstream would make the shared walker prefer it.

</details>

<details>

- [x] Design and architecture: the reason clause, the third record, the unknown branch
- [x] Correctness: the note composition (1), the hand-listed set (2), the read outcomes (3)
- [x] Security: the read is now bounded and refuses non-regular files, which closes round 2's dropped item
- [x] Test coverage: 10 mutations, 8 killed; the survivors are (2) and (4)
- [x] Reliability: the record's lifecycle, the reset coverage, the probe's default arm
- [x] Documentation currency: CHANGELOG, CLAUDE.md and the guide all carry v0.22.2
- [x] Cross-repo contracts: the v0.22.2 element gains `display_in`, and the generated scaffold matches, see (7)
- [x] Usability: every reachable rendering of the two notes composed and read
- [x] Project rules compliance: `CLAUDE.md` credential and generated-code policies
- [na] Performance, frontend.

Diff since the reviewed `15e033a`: 15 files, +582/−74, head `1a20778`. Verified locally: `go build ./...`, `go test ./...` (34 packages, all pass), `make lint` (0 issues), and `verify-generated`, `verify-gateway-coverage`, `verify-classic-schemas`, `verify-platform-specs`, `verify-site` all clean. The 16 Security Cloud groups and the one omitted from (2)'s sentence were counted from the built binary. Lanes run this round: devil-advocate, silent-failure-hunter, test-quality-reviewer. Not re-dispatched: security-reviewer, whose one open item was the unbounded read this commit bounds; fidelity-reviewer, scope-reviewer and usability-reviewer, all three having run in round 1. Never run: performance-reviewer, inapplicable.

</details>

<details>

✅ Every verification claim in the commit message held under re-mutation, including the one that had survived two rounds. `TestCatalogJSONOmitsTheEmptyPositiveOnlyKeys` judges each row against the entry field the projection was handed rather than against a second copy of the rule, and it fails if it finds no rows of the positive kind, so it cannot go quietly vacuous.

✅ The new school test asserts its own premise. It fails if `school blueprints list` stops declaring environment without tenant, and it reaches the command through the tree `main.go` executes. That is the specific reason its round-2 predecessor passed while production printed something else.

✅ Fix (5) states plainly that the earlier justification was false and why, rather than replacing one confident sentence with another. The same instinct is what makes the `keychain:` cost readable two rounds later.

</details>



neilmartin83:
@ktn-jamf ready for re-review — `1d7b64a2`.

Round 3's four findings are addressed, but the bigger change came from re-litigating a decision you didn't flag: **`platform setup` no longer probes Jamf Security Cloud, or any product's access or permissions.** That dissolves finding (2) and both nice-to-haves rather than fixing them.

## The entitlement verdict asked the wrong question and answered it wrongly

Wrong question: a capability permission is granted **per operation** when the integration is created, so one read establishes nothing about the other 28 resources. And the runtime answer is both already there and more specific — `EnrichPrivilegeError` maps a gateway 403 onto exit 5 and names the permission in the section-and-picker wording Jamf Account searches by. On the ungranted tenant credential:

```
"hint": "grant the Jamf Platform API integration these permissions in Jamf Account:
         Secure enterprise access > Content categories: Read (content-categories:read)."
"exitCode": 5
```

Wrong answer: wire-checked 2026-09-08 in **one organization**, a *tenant* credential answered `BAD_PERMISSIONS` on `/securitycloud/v1/categories` while an *environment* credential in the same organization answered **200 with 36 categories**. So the summary told a demonstrably entitled organization it had no entitlement and subtracted all sixteen Security Cloud resources from what the profile reaches. That is your finding (6)'s asymmetry resolved by deleting the claim: `BAD_PERMISSIONS` cannot separate no entitlement from one missing grant, so nothing here may report either. That credential's setup now reports **16 of the 29** reachable where it reported none.

Gone: `securityCloudVerdict` and its three constants, `securityCloudResourceGroups`, `anyIn`, the unentitled partition, the unknown sentence. Net 42 lines lighter.

## What survives is a scope-ID check, and it is not a product check

The token exchange sends no scope header, so credentials that authenticate say nothing about the ID just typed — #354's mis-paste saves cleanly and then refuses every command. **The gateway resolves the scope at the edge, before routing and before capability**, which is what makes this level-only rather than product-specific. Wire-probed 2026-09-08 on an EU tenant, an EU environment and a US organization credential:

| scope header | answer |
|---|---|
| the credential's own tenant or environment ID | **200** |
| an unknown `X-Environment-Id`, or a tenant ID pasted there (#354) | **404 `ENVIRONMENT_NOT_FOUND`**, naming the value |
| an `X-Tenant-Id` it will not accept, including an environment or organization ID pasted there | **403 `OWNERSHIP_FORBIDDEN`**, naming the value |
| none | **400 `REQUEST_CONTEXT_NOT_PROVIDED`** |

Both rejection codes arrive unchanged on `/pro/v1/definitely-not-a-real-endpoint` and on a Security Cloud path, and only an unrouted *namespace* escapes scope evaluation (Tyk's bare `404 page not found`). So the probe reads `jamf-pro-version` purely as a carrier — one of the 44 Jamf Pro endpoints declaring no privilege, so a correct ID answers a clean 200 instead of the `BAD_PERMISSIONS` every grant-bearing path returns. Everything else leaves the ID unjudged, `BAD_PERMISSIONS` included: it means the header resolved. Skipped entirely for organization scope.

Two caveats now in `CLAUDE.md`: the tenant refusal is a **403**, so it is only unambiguous matched on the structured `code` and not the status; and `OWNERSHIP_FORBIDDEN` means unusable *by this credential*, not malformed, so another org's real tenant reads the same as a typo.

**The runtime half was missing.** `ENVIRONMENT_NOT_FOUND` is a 404, so it reached neither `EnrichPrivilegeError` (403 only) nor the missing-scope arm and arrived bare — the one mis-paste direction with no explanation anywhere. `AnnotateScopeLevelError` now annotates it ahead of the level arms, since the gateway not recognising the value makes the declared levels beside the point.

## Round 3 findings

- **(1)** Right, and the remedy half was the sharper part. The note makes no claim about whose integration the ID names, names the unreadable path as the thing to fix, and warns against dropping `JAMF_CLIENT_ID` — which is worse than the failure it replaces, since `ResolveSecret(p.ClientID)` then opens the same unreadable file. `withheldNoteState` returns `withheldNoteBeside` on that branch so `scopeLevelNote` still carries the declared levels; asserted rather than assumed.
- **(2)** Dissolved. The hand-listed sentence is gone, `enrollment-activation-profiles` with it, and so is `anyIn`.
- **(3)** Both outcomes fixed as you wrote them: over-cap and empty now return errors, so the record is set and the note names the path instead of blaming the variable. The at-cap boundary still reads.
- **(4)** Held in place twice. The directory case asserts `not a regular file`, which a downstream `EISDIR` cannot produce, and a new `//go:build unix` test drives a real FIFO against a deadline — calling `readClientIDRefFile` directly, as suggested, would *hang* the package on mutation rather than fail it, so the timeout is the assertion.
- **(7)** Unchanged and still true: `display_in`'s scaffold value needs an upstream `example: true` for the shared walker to prefer it. Both facts stay in `specs/`, CLAUDE.md and the CHANGELOG.

**Ten mutations run, ten killed** — the IsRegular guard (both tests fire), the empty and over-cap refusals, the unreadable-note branch, the `withheldNoteState` special case, the `ENVIRONMENT_NOT_FOUND` arm, and reading `BAD_PERMISSIONS` as a rejected ID.

## stop-slop pass

Em dashes out of every user-facing string in `platform_gateway_setup.go`, `platform_scope.go` and the 403 hint in `internal/privileges`; passive voice into active ("This invocation sent no scope header", "Jamf Account creates an integration at one level"); and the historical archaeology out of the comments that had accumulated over three rounds.

## Verification

`platform setup` driven on a pty at all four shapes, one run each:

| shape | result |
|---|---|
| EU environment, correct ID | `ok: the gateway accepts this environment ID`, reaches all 29 |
| EU tenant, correct ID | `ok: ... tenant ID`, **16 of the 29** (was: none) |
| tenant ID at the environment prompt (#354) | `rejected: the gateway does not know this environment ID`, summary suppressed |
| environment ID at the tenant prompt | `rejected: ... will not accept this tenant ID`, summary suppressed |
| US organization, both prompts skipped | probe skipped, Jamf Account surfaces named |

Plus `pro blueprints list` with a tenant ID in `JAMF_ENVIRONMENT_ID`, which returns the gateway's own 404 and the new note.

`make test`, `make lint` (0 issues) and all five verify targets clean; `make generate` idempotent.

grahampugh (approved):

PR #372, opened 2026-09-08
feat!: unify the request-body flag on --from-file, add stdin and apply to the gateway products

neilmartin83:
## Summary

Three changes, one theme: the request-body flag now has one name across the whole CLI, and the two products that spelled it differently also gain the two things that name implies — piped stdin, and `apply`.

**⚠️ Breaking, no deprecation window.** `--file` answers `unknown flag` (exit 2) on Platform and Security Cloud writes. Anyone with `--file` in a script or in `JAMF_CLI_ARGS` breaks on upgrade. The migration is a rename, and the error names the replacement.

## 1. `--from-file` everywhere (breaking)

Platform and Security Cloud commands took `--file`; Jamf Pro, Classic, Protect and School took `--from-file`. `pro platform-device-groups` had **both, in one file** — `patch` and `patch-members` on `--file`, `apply` on `--from-file`, three flag registrations apart.

41 registrations renamed, at the two generator templates plus that one hand-written file. All 361 commands that read a body from a path now agree.

```bash
# Before
jamf-cli security ztna-gateways create --file gateway.yaml
# After
jamf-cli security ztna-gateways create --from-file gateway.yaml
```

**No compat alias, deliberately.** Two spellings surviving in parallel is the state this removes. A script passing the old name fails loudly rather than drifting.

**`--file` is kept on the 11 upload/import commands** (`pro packages upload`, `pro icons upload`, `protect analytics import`, …) and must not be renamed with them. `--from-file` promises the body can arrive on a pipe instead; a multipart upload cannot keep that promise, because the transport needs a filename and a length. Renaming those too would produce one flag name with two capabilities — the same inconsistency, spelled the other way round.

`TestRequestBodyFlagIsUniformlyFromFile` walks the assembled tree and refuses any `--file` not on the named upload list, so the split cannot recur silently.

## 2. Platform and Security Cloud bodies can be piped

This was the larger defect. `ReadBody` in both `internal/platform` and `internal/security` was `os.ReadFile(file)` and nothing else, where Pro's `readApplyInput` and Protect's `readInput` had always checked `os.ModeCharDevice`. Those two products were the only ones where `--scaffold | edit | apply` had to route through a temp file — so renaming the flag alone would have shipped one name with two capabilities.

```bash
jamf-cli security ztna-gateways apply --scaffold | vipe | jamf-cli security ztna-gateways apply --yes
```

**An empty pipe is not a body; an empty *named file* is.** A CI runner hands every process a non-character-device stdin carrying nothing, so reading that as "a body was supplied" breaks every bodyless write under CI. A file the caller named and a file they did not are different instructions, and `bodyinput.Normalize` already errors on empty input — the `file != ""` term in `hasInput` is what keeps that error reachable.

## 3. `apply` on the platform gateway resources

`apply` existed on Pro and Classic and on no gateway resource, so `--scaffold` had no idempotent terminus and every caller wrote the exists-check by hand.

`buildApplySpec` composes a resource's own list, collection POST and item PUT/PATCH, and emits nothing unless **all** of: an own list path (a `crossResourceNameLookupPath` means the `{id}` belongs to a sibling, so there is no collection here to create into), a non-destructive non-multipart collection POST, an item update taking exactly one path param, and a name-ish property actually present in the create body. Six resources qualify: `dns-zones`, `ztna-apps`, `ztna-gateways`, `ztna-grouped-gateways`, `device-groups`, `ai-policies`. PUT is preferred over PATCH where both exist, because replace is what `apply` means elsewhere.

Three properties worth reviewing:

- **The name comes from the body, never a flag.** The input is the desired state and carries its own identity; a `--name` beside it would be a second source of truth with no correct resolution when the two disagree. A non-string or empty name is refused before anything is sent, because the value is matched against list results and `"1"` matching an item named `1` is a coincidence, not a resolution.
- **Only a genuine `ErrNotFound` takes the create branch.** Treating an auth error or a 5xx during the lookup as absence is how a 403 becomes a duplicate.
- **The help sentence is generated from the update method**, not written once — a partial body over PATCH keeps omitted fields, over PUT clears them, and that is the one thing about this verb an operator gets wrong. `platformPatchDoesNotMerge` overrides it where the server contradicts its own method: `ai-policies` sends `application/merge-patch+json` and replaces `settings` wholesale anyway.

**`blueprints` and `platform-device-groups` are blocklisted.** Both already have hand-written `apply` commands that do strictly more (portable scope resolution and ID randomisation; `--device-type` disambiguation), and both parents copy in *every* generated subcommand — so without the blocklist cobra would carry two subcommands named `apply` and dispatch by declaration order.

**The Radar half has no `apply` and should not.** Its whole surface is singletons whose `update` is already an idempotent create-or-replace (`stream`, `status`), actions (`risk override`, `verification trigger`, `device-lifecycle purge`) and read-only documents (`well-known`, `jwks`) — no named collection to resolve against. That is a property of the API rather than an omission, so `TestNoSecurityResourceQualifiesForApply` asserts it and fails if a spec refresh adds a real collection.

## 4. The migration hint is looked up, not measured

`suggestFlag` takes the nearest flag within two edits, and `file` is two edits from `field` and five from `from-file` — so every caller migrating off `--file` was pointed at `--field`, an output-field selector, on a breaking change whose entire value is being navigable.

```
$ jamf-cli security dns-zones create --file zone.yaml
hint: did you mean --from-file?
unknown flag: --file
```

`renamedFlags` carries both directions (`--from-file` is the name most callers know, so it gets typed at the uploads too) and each entry fires **only where its destination is a real flag on the command in hand**. No command declares both, and a test asserts that rather than trusting it.

## Verification

Wire-verified against a live tenant (Elevate EU, `platform ai-policies`), **writes included**:

- Two runs of the same piped body — the first created and reported the new id, the second updated in place, and the collection still held exactly one policy of that name with the changed field read back.
- `--set` overlays onto a piped body and the resulting name resolves the same way.
- Without `--yes` on a non-terminal stdin the update is refused rather than performed.
- Piped JSON and YAML both parse; a nameless body is refused before any request is built.
- Probe policy archived afterwards; tenant left as found.

Also verified: `--dry-run` previews the correct branch (PATCH against the resolved real id vs POST to the collection), and the old flag fails with the new hint.

All CI steps pass locally: `golangci-lint` (0 issues), `make test`, `make verify-generated`, `make verify-platform-specs`, `make verify-security-specs`, `make lint-dead`, `make build`, `make verify-site-output`, site JS parse.

## Caught during development, worth knowing

- `singularize("content-status")` returned `content-statu`. A trailing `s` is not a plural in `status`, `analysis` or `address`; those endings are now named ahead of the general rule.
- My first cut of `ReadBody` broke `TestReadBody_EmptyFileErrors` — it treated an empty named file like an empty pipe. That test existed for a good reason and caught it.
- `TestSecurityCloudSpecParity` pairs each subcommand 1:1 with a spec endpoint, so six correctly-generated `apply` commands read as six *missing* operations. Taught it about synthesized verbs rather than bumping the count, so the next one does not re-break it.
- `--from-file` has a **third** sense I had not accounted for: a newline-delimited list of IDs or names (the bulk deletes, `multi`), which never read stdin and should not claim to. The uniformity test asserts the flag name only; stdin behaviour is asserted where it lives, in the two `ReadBody` test suites.
- The apply confirmation passed a whole sentence where `confirmStmt` passes a bare verb, so `ConfirmAction`'s `"%s on %q requires --yes"` rendered as *"update existing ai-policy on "x" requires --yes"*. Now passes `"update"`.

## Noted during verification, not this PR's to fix

- **`platform ai-policies --scaffold` emits `schemaVersion: 2026-05-12` and the tenant answers `422 SCHEMA_VERSION_UNKNOWN`** (it wants `2026-06-03`). That value is the published spec's own example, so it is upstream, not a CLI defect — recorded here only because it is why the first probe run failed and because it means `--scaffold | apply` needs the field edited on that one resource. `apply` built and sent the POST correctly.
- The CLI reads `JAMF_CLIENT_ID`/`JAMF_CLIENT_SECRET`/`JAMF_ENVIRONMENT_ID`, while the platform SDK and terraform provider use `JAMFPLATFORM_*`. Exporting the SDK names is silently ignored and the CLI falls back to the default profile — which presents as a permissions problem rather than a config one. Separate concern, raised and deferred.


ktn-jamf (commented):
> [!WARNING]
> ⚠️ **Needs changes**. Unifies the Platform/Security Cloud request-body flag to `--from-file`, adds stdin support, and synthesizes a new `apply` verb for six gateway resources.
> Blocking: (1), (2), (3). See the collapsed sections for 2 nice-to-have suggestions and a PR-decomposition recommendation.

## Rating: 2/5
- Would be a 4 once (1)–(3) are fixed. All three are fixable in place. (4)–(13) are further IMPORTANT gaps in the new `apply` verb worth fixing in the same pass.

---

❌ **CRITICAL** (1) (repo-hygiene) — `site` (repo root): a 4.2 MB compiled binary is committed

The diff adds a file named `site` at the repo root. `file site` reports `Mach-O 64-bit executable arm64`. It is not in `.gitignore` (only `bin/`, `dist/`, and `/jamf-cli` are), it did not exist on `origin/main`, and it is unrelated to any of this PR's three changes. It reads as a local `go build -o site ...` accidentally `git add`-ed.

**Failure scenario:** merging as-is adds 4.2 MB of binary history to the repo permanently, and anyone cloning fresh downloads an executable that has nothing to do with the CLI's own build output.

**Suggested fix:**
```diff
- site (binary, committed)
```
Drop the file from the commit (`git rm --cached site` before the final push, or amend it out) and confirm `git status` shows nothing untracked matching build output.

**Fixed when:** `site` is gone from the diff and `git ls-tree` at the PR head has no top-level binary.

---

❌ **CRITICAL** (2) (correctness, via security-reviewer+silent-failure-hunter) — `internal/commands/platform/generated/device-groups.go:399`: `security device-groups apply` can never find an existing group, so every re-apply creates a duplicate

The exists check calls `platform.ResolveIDByName(ctx, client, "/securitycloud/v2/groups", name)`. CLAUDE.md records, as a wire-verified fact, that this endpoint answers `{"groups":[{"name":"Default Group"}]}` — items carry only `name`, no `id`. `extractID` (`internal/platform/resolve_generic.go:204`) recognizes only `id`/`blueprintId`/`groupId`/`deviceId`/`benchmarkId`/`baselineId`, so it returns `""` for every item, `firstMatch` never records a match, and the lookup always returns `ErrNotFound` — whether or not the group exists. `apply`'s `id == ""` branch then always takes CREATE; the overwrite-confirmation prompt on the UPDATE branch is dead code for this resource.

**Failure scenario:** running `security device-groups apply --from-file group.yaml` twice with an unchanged `name` (the ordinary reconciliation use of `apply`) silently creates a second device-group with the identical name on the second run, exit 0, no confirmation, no warning.

**Suggested fix:**
```diff
 var platformNoApply = map[string]bool{
 	"blueprints":             true,
 	"platform-device-groups": true,
+	// device-groups (Security Cloud): GET /securitycloud/v2/groups returns
+	// {"groups":[{"name":"..."}]} with no id, so ResolveIDByName can never
+	// match an existing group and apply always creates a duplicate.
+	"device-groups": true,
 }
```

**Fixed when:** `security device-groups apply` run twice on the same name either updates the existing group or fails loudly, rather than creating a second one.

---

❌ **CRITICAL** (3) (correctness, via usability-reviewer+test-quality-reviewer) — `generator/platform/emitter.go:654`: `apply`'s PATCH sends a different `Content-Type` than the resource's own `patch` command for the identical endpoint

`UpdateMergePatch: update.RequestBody.IsMergePatch` (`emitter.go:654`) reflects only whether the spec literally declares `application/merge-patch+json`. `specs/platform/ai_governance_policies_api.json` declares `application/json` for this PATCH, so `UpdateMergePatch` is `false` and `apply`'s update sends `Content-Type: application/json` (`internal/commands/platform/generated/ai-policies.go:706`). The standalone `ai-policies patch` command hits the identical `PATCH .../{policyId}` through plain `DoExpect` with no explicit content type, and the SDK's transport (`jamfplatform-go-sdk@v0.20.1/internal/client/client.go:670-671`) defaults *any* PATCH with no explicit override to `application/merge-patch+json`. CLAUDE.md records that default as wire-verified CLI behavior for this exact endpoint. So the two commands send different headers to the same URL, and `apply`'s own generated help text ("Other top-level fields do keep their current values when omitted") describes the merge-patch behavior `apply` itself does not send. A test-quality mutation (flip the literal, rerun the full scoped suite) confirmed no test catches either value.

**Failure scenario:** `platform ai-policies apply` on an existing policy sends `Content-Type: application/json` where `ai-policies patch` on the same URL sends `application/merge-patch+json` — a provable divergence on the wire, untested in either direction.

**Suggested fix:**
```diff
- UpdateMergePatch: update.RequestBody.IsMergePatch,
+ UpdateMergePatch: update.RequestBody.IsMergePatch || strings.EqualFold(update.Method, http.MethodPatch),
```

**Fixed when:** `ai-policies apply`'s update sends the same `Content-Type` as `ai-policies patch`, and a test asserts the rendered Content-Type literal for both `UpdateMergePatch` branches.

---

⚠️ **IMPORTANT** (4) (correctness, via devil-advocate) — `generator/platform/emitter.go:592`: `apply`'s list-then-create has no duplicate protection on 4 of 6 resources

`buildApplySpec`'s only stated safety net is "the collection's names are unique... `ResolveIDByName` already errors on a duplicate." That is a client-side check, not a server one. `specs/platform/securitycloud_device_groups_api.json` and `securitycloud_dns_api.json` both declare a 409 on a duplicate create; `securitycloud_ztna_api.json`'s create-409 codes never include a name-uniqueness check, and `ai_governance_policies_api.json`'s `POST /policies` declares no 409 at all.

**Failure scenario:** a CI retry or a concurrent run of `apply --from-file` against `ztna-apps`, `ztna-gateways`, `ztna-grouped-gateways`, or `ai-policies` both list, both see "not found", both POST-create — two resources with the identical name, exit 0 on both.

**Suggested fix:** name the gap in the generated help for the four unprotected resources, or add a post-create existence re-check before reporting success.

**Fixed when:** the four resources' `apply --help` discloses the retry/race risk, or a re-check catches the duplicate.

---

⚠️ **IMPORTANT** (5) (correctness, via devil-advocate) — `internal/platform/resolve_generic.go:89`: ambiguous-name detection stops at the first page that matches

`resolveIDByName` returns as soon as `firstMatch` finds a match on any page; the pagination loop only continues when the *current* page has zero matches. A 150-item collection with the same name on page 1 and page 2 never surfaces the ambiguity — page 1's single match wins.

**Failure scenario:** `apply` targets a name that exists twice, once per side of a 100-item page boundary. The lookup returns the page-1 item as if it were unique, and `apply` upserts the wrong object with no ambiguity error.

**Suggested fix:** accumulate matches across every page before deciding, rather than returning on the first page that matches.

**Fixed when:** a two-page fixture with a same-named item on each page returns an ambiguous-match error instead of the first one found.

---

⚠️ **IMPORTANT** (6) (correctness, via devil-advocate) — `internal/commands/platform/generated/ai-policies.go:618`: `ai-policies apply` never publishes, so the change it reports never takes effect

CLAUDE.md documents ai-policies as having a draft/publish split: `create`/`patch` write a draft, and a separate `publish` (no body) makes it the effective version. `newAiPoliciesApplyCmd`'s RunE calls only the create and update (draft) operations — no reference to publish appears in its body at all.

**Failure scenario:** `platform ai-policies apply --from-file policy.yaml --yes` exits 0 with "Updated ai-policy", but the enforced policy is whatever was last *published* — the change sits as an unpublished draft with nothing in the output saying so.

**Suggested fix:** have `apply` call `publish` after a successful create/update, or add a line to `applyLong` for this resource stating the draft/publish distinction explicitly.

**Fixed when:** `ai-policies apply --help` states the draft/publish split, or `apply` itself publishes and a test asserts the publish call on both branches.

---

⚠️ **IMPORTANT** (7) (usability, via usability-reviewer+devil-advocate) — `internal/platform/resolve_generic.go:180`: the ambiguous-match error tells `apply` to do something `apply` cannot do

`firstMatch`'s duplicate-name error reads "pass the positional ID to disambiguate." Every generated `apply` command takes no positional argument (`RunE: func(cmd *cobra.Command, _ []string) error`) and its whole design, stated in its own `Long` text, is that the name comes only from the body. The error is not wrapped in `ErrNotFound`, so it reaches the caller verbatim.

**Failure scenario:** two `ztna-gateways` (or any apply-eligible resource) share a name. `apply` fails with an instruction — "pass the positional ID" — that the command has no flag or argument to satisfy.

**Suggested fix:**
```diff
  id, err := platform.ResolveIDByName(cmd.Context(), cliCtx.PlatformSDKClient, listPath, name)
  if err != nil && !platform.IsNotFound(err) {
-     return err
+     return fmt.Errorf("apply: %w (apply has no id to pass; resolve the conflict with this resource's update command)", err)
  }
```

**Fixed when:** the ambiguous-match error surfaced through `apply` names a remedy `apply` itself supports.

---

⚠️ **IMPORTANT** (8) (usability, via usability-reviewer) — `generator/platform/emitter.go:487`: `apply`'s annotations drop the `jamf:scopes` tag every sibling verb on the same resource carries

`applyAnnotations` emits only `jamf:api` and `jamf:privileges`. Every one of the six qualifying resources' own `create`/`update`/`list` commands carries `jamf:scopes` (e.g. `ai-policies.go`'s `patch` command: `"jamf:scopes": "environment"`). `scopesOf` (`internal/commands/platform_scope.go:402`) reads `cmd.Annotations["jamf:scopes"]` and returns `nil` when absent, which is exactly `apply`'s case — the scope-mismatch enrichment in `AnnotateScopeLevelError` can never fire for `apply`, and `apply` also drops out of the `commands -o json` catalog's `scopes` field.

**Failure scenario:** a tenant-scoped credential runs `device-groups apply` against a resource build v2082 moved to environment-only scope. `create` on the same body gets the enriched "declares environment, you're using tenant" note; `apply` gets the bare gateway 400 with no such note.

**Suggested fix:**
```diff
  Privileges:       unionPrivileges(create, update),
+ ScopeTypes:       unionScopeTypes(create, update),
```
and add the corresponding `jamf:scopes` pair in `applyAnnotations`.

**Fixed when:** `apply`'s generated `Annotations` carries `jamf:scopes` matching its resource's other verbs.

---

⚠️ **IMPORTANT** (9) (usability, via usability-reviewer) — `internal/commands/root.go:994`: the `--file`→`--from-file` migration hint bypasses this CLI's own structured error envelope

`fmt.Fprintf(os.Stderr, "hint: did you mean --%s?\n", s)` fires from `SetFlagErrorFunc` as a raw stderr write, then `exitcode.Wrap(exitcode.Usage, ferr)` returns an error with `Hint` left empty — so `envelope["hint"]` (the field every other hint in this CLI populates, per CLAUDE.md's own documented convention) is never set for this case. Verified by building and running the PR binary: the hint line prints *before* the JSON error block in combined output, and `-o json` carries no `hint` field at all.

**Failure scenario:** a CI job wrapping `jamf-cli ... --file x -o json` and reading `.hint` to auto-remediate gets nothing; a job piping `2>&1` into a JSON parser gets a corrupting stray line ahead of the JSON.

**Suggested fix:**
```diff
- fmt.Fprintf(os.Stderr, "hint: did you mean --%s?\n", s)
- return exitcode.Wrap(exitcode.Usage, ferr)
+ e := exitcode.Wrap(exitcode.Usage, ferr)
+ e.Hint = fmt.Sprintf("did you mean --%s?", s)
+ return e
```

**Fixed when:** `-o json` on the rename scenario shows the hint inside the JSON `hint` field.

---

⚠️ **IMPORTANT** (10) (reliability, via security-reviewer+silent-failure-hunter) — `internal/platform/body.go:90`, `internal/security/body.go:92`: piped stdin over 10 MB is silently truncated, not refused

`io.ReadAll(io.LimitReader(os.Stdin, 10<<20))` returns a plain `io.EOF` at the cap, indistinguishable from real end-of-input — no length check follows. A truncated YAML/JSON body can still parse cleanly past the cut point.

**Failure scenario:** piping a >10 MB desired-state document into `apply` on a PUT-semantics resource (whose own help says omitted fields are cleared) silently drops trailing fields and reports success.

**Suggested fix:**
```diff
- raw, err := io.ReadAll(io.LimitReader(os.Stdin, 10<<20))
+ raw, err := io.ReadAll(io.LimitReader(os.Stdin, 10<<20+1))
  if err != nil {
      return nil, fmt.Errorf("reading stdin: %w", err)
  }
+ if len(raw) > 10<<20 {
+     return nil, fmt.Errorf("stdin body exceeds 10MB; use --from-file for larger payloads")
+ }
```

**Fixed when:** piping a body over the cap errors explicitly instead of silently truncating, in both `internal/platform` and `internal/security`.

---

⚠️ **IMPORTANT** (11) (correctness, via silent-failure-hunter) — `internal/platform/body.go:99`: `applySet` silently overwrites a non-object field, where the hand-duplicated `internal/security/body.go:103` refuses the same input

Platform's `applySet` replaces any existing scalar at a `--set` path segment with a fresh map, no error. Security's version checks `present` first and returns `"cannot set nested key under non-object field %q"`. The divergence predates this PR, but this PR's new `apply` command is now the newest, most consequential caller of the permissive one.

**Failure scenario:** `apply --from-file body.yaml --set scope.owner=alice` where `body.yaml` already has `scope: "team-a"` (a string) silently replaces it with `{"owner":"alice"}`, discarding the original value.

**Suggested fix:**
```diff
-func applySet(m map[string]any, path []string, v any) {
+func applySet(m map[string]any, path []string, v any) error {
 	for i, segment := range path {
 		if i == len(path)-1 {
-			m[segment] = v
-			return
+			m[segment] = v
+			return nil
 		}
-		next, ok := m[segment].(map[string]any)
+		existing, present := m[segment]
+		next, ok := existing.(map[string]any)
 		if !ok {
+			if present {
+				return fmt.Errorf("cannot set nested key under non-object field %q", segment)
+			}
 			next = map[string]any{}
 			m[segment] = next
 		}
 		m = next
 	}
+	return nil
 }
```

**Fixed when:** `internal/platform/body.go`'s `applySet` errors on the same input `internal/security/body.go`'s already refuses.

---

⚠️ **IMPORTANT** (12) (reliability, via silent-failure-hunter) — `internal/platform/body.go:86`, `internal/security/body.go:88`: a genuine `os.Stdin.Stat()` error is treated identically to "definitely no input"

`if err != nil || (stat.Mode()&os.ModeCharDevice) != 0 { return nil, nil }` collapses a stat failure into the same silent "empty" result the function's own doc comment reserves for the empty-pipe-under-CI case. The comment never addresses `err != nil`, a materially different condition.

**Failure scenario:** on a platform or sandbox where `Stat()` fails while real piped content is available, a create/update command silently sends no body instead of the caller's intended payload, with no error at any verbosity level.

**Suggested fix:**
```diff
 stat, err := os.Stdin.Stat()
-if err != nil || (stat.Mode()&os.ModeCharDevice) != 0 {
+if err != nil {
+    return nil, fmt.Errorf("checking stdin: %w", err)
+}
+if stat.Mode()&os.ModeCharDevice != 0 {
     return nil, nil
 }
```

**Fixed when:** a `Stat()` error surfaces as an explicit error in both files, rather than collapsing into "no input".

---

⚠️ **IMPORTANT** (13) (test-coverage, via test-quality-reviewer) — `generator/platform/template.go:366`: the generated `apply` RunE has zero test coverage

`internal/commands/platform/generated/` has no `_test.go` files at all. This PR's own `internal/platform/apply_test.go` exercises only the pure `ApplyName` string-extraction helper. Nothing constructs a mocked list endpoint to exercise the ambiguous-match path, a non-"not found" lookup error, or the create-vs-update fork; `internal/commands/platform_generated_test.go`'s existing `TestGeneratedPlatformMutationsHonourDryRun` covers other `device-groups` verbs but was not extended to `apply`.

**Failure scenario:** a future refactor that mishandles the not-found-vs-error distinction, or the create/update fork, ships silently — the full scoped suite passes today with 0% coverage of this branch logic, confirmed by a mutation test that flipped a literal in this exact code with no test failure either direction.

**Suggested fix:** add `httptest`-backed tests mirroring `TestGeneratedPlatformMutationsHonourDryRun`'s pattern: one-match→update, zero-match→create, two-matches→ambiguous error with no write, list-500→error with no write, `--dry-run` sends zero requests on both branches.

**Fixed when:** a mutation to the not-found check, the ambiguity handling, or the create/update fork fails a test.

---

<details>

💡 **NICE-TO-HAVE** (14) (correctness, via devil-advocate) — `generator/platform/emitter.go:644`: `NameLookupField` is set unconditionally, independent of whether `NameField` actually used it

`buildApplySpec` can fall back `NameField` to the standard `name`/`title`/`displayName` when the create body lacks the resource-specific override, but `applySpec.NameLookupField` is still set to the raw override value — so a future resource in that shape would read identity from one field and search for it under another. Not live today: `platformNameLookupFields` holds only `sso/sso-domains`, which doesn't qualify for `apply`.

**Fixed when:** `NameLookupField` is set only when `NameField` actually resolved to it, with a test pinning the fallback case.

💡 **NICE-TO-HAVE** (15) (usability, via usability-reviewer) — `internal/commands/pro_packages_upload.go` and the other 10 leaves in `bodyFileUploadLeaves`: `--help` gives no signal these are the deliberate exception to the new `--from-file` convention

CHANGELOG.md and the platform-api-ga guide explain the split at length, but a user reading only `--help` sees nothing about why this command keeps `--file`.

**Fixed when:** each upload leaf's `Long` or flag description states in one sentence why it uses `--file`, not `--from-file`.

</details>

---

<details>

**Recommendation: split into 3 PRs.** The author's own CHANGELOG entry already partitions the diff into "Breaking" (flag rename), "Behaviour" (stdin), and "Added" (`apply`) — evidence the three concerns are cleanly separable, not entangled. None needs the others present as a stub.

1. **Stdin support** (`internal/platform/body.go`, `internal/security/body.go` + tests) — no dependencies, smallest, purely additive.
2. **The breaking rename** (`generator/platform/template.go`, `generator/security/template.go`'s flag lines, `pro_platform_device_groups.go`, `root.go`'s `renamedFlags`, the uniformity tests) — no dependencies, sequenced so `apply`'s generated code is born with the final flag name.
3. **The `apply` verb** (`generator/platform/emitter.go`, the new `apply` template block, `internal/platform/apply.go`, the six affected generated files) — depends on (2).

The stray `site` binary (finding 1) must be dropped before merge regardless of the split decision. If the team judges the coordination cost of three reviews not worth it for a trusted, already-wire-verified change, keeping one PR is defensible — but the binary removal is not optional either way.

</details>

<details>

- [x] Design and architecture: `buildApplySpec`'s qualification rules, `platformNoApply` blocklist, findings (2), (4), (8)
- [x] Correctness: create/update branch, name resolution, Content-Type derivation, findings (2)-(7), (11), (14)
- [x] Security: stdin credential exposure (no new violation), 10MB cap, finding (10)
- [x] Test coverage: generated `apply` RunE, finding (13)
- [x] Reliability: `Stat()` error handling, findings (10), (12)
- [x] Documentation currency: CLAUDE.md/CHANGELOG/platform-api-ga.md all updated accurately
- [x] Project rules compliance: `.claude/rules/` absent; CLAUDE.md's Credential Input Policy checked against the new stdin path
- [na] Performance, frontend.

Diff: 65 files, +2826/−130, head `745afeb`. Lanes run: security-reviewer, usability-reviewer, scope-reviewer, silent-failure-hunter, devil-advocate, test-quality-reviewer. Never run: `fidelity-reviewer` (no linked spec/issue), `performance-reviewer` (no DB/loop/caching surface in the diff).

</details>

<details>

| File | Score | Notes |
|------|-------|-------|
| `site` | 0/5 | Committed compiled binary, unrelated to the PR (see (1)) |
| `internal/commands/platform/generated/device-groups.go` | 1/5 | New `apply` can never detect an existing group (see (2)) |
| `generator/platform/emitter.go` | 2/5 | `buildApplySpec`'s Content-Type derivation, scope-annotation gap, TOCTOU gap (see (3), (4), (8), (14)) |
| `internal/commands/platform/generated/ai-policies.go` | 2/5 | Wrong Content-Type and no publish step on `apply` (see (3), (6)) |
| `internal/platform/resolve_generic.go` | 3/5 | Pre-existing code newly load-bearing for `apply`'s safety story (see (5), (7)) |
| `internal/platform/body.go` | 2/5 | Truncates stdin silently, clobbers non-object fields, swallows stat errors (see (10)-(12)) |
| `internal/security/body.go` | 4/5 | Shares the stdin truncation and stat-error gaps but keeps the `applySet` guard (see (10), (12)) |
| `internal/commands/root.go` | 3/5 | Rename hint bypasses the structured error envelope (see (9)) |
| `internal/platform/apply.go` | 4/5 | `ApplyName` itself is clean and well-tested; the caller it serves is not (see (13)) |

</details>



_Generated by `pr-review` v1.33.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

neilmartin83:
Addressed in 504b88b4 — 12 of 13 findings fixed, both nice-to-haves done, and one CRITICAL not actioned because it did not survive wire verification.

## Fixed

**(1) committed `site` binary** — `git rm --cached`, `/site` added to `.gitignore`.

**(3) `apply`'s PATCH Content-Type** — confirmed against the SDK (`internal/client/client.go:713`: any bodied PATCH with no override defaults to `application/merge-patch+json`). `UpdateMergePatch` now derives from the method as suggested, so `ai-policies apply` sends the same header as `ai-policies patch`. `TestGeneratedApplySendsThePatchContentTypeItsSiblingDoes` pins it; mutation-checked (flipping the literal fails it).

**(4) list-then-create race** — disclosed in every `apply --help`, not only the four. A declared 409 is not a promise the server enforces it *on the name* (ztna's 409 codes are hostname and reference conflicts), so a per-resource claim of protection would be worse than the caveat.

**(5) ambiguity across pages** — matches now accumulate across every page before the decision (`collectMatches`). `TestResolveIDByName_AmbiguousAcrossPages` uses a two-page fixture with a copy on each.

**(6) `ai-policies apply` never publishes** — help states the draft/publish split (`platformApplyCaveats`). Not auto-publishing: publish is non-idempotent (`409 NO_DRAFT_TO_PUBLISH`), and folding it in would leave no way to write a draft alone, which is the workflow the split exists for.

**(7) ambiguity remedy** — fixed at the resolver rather than wrapped at the call site, so every caller gets a remedy it can satisfy: "identify it by ID instead". `TestResolveIDByName_AmbiguityRemedyIsNotAPositional` fails on the word "positional".

**(8) `jamf:scopes`** — `unionScopeTypes(create, update)` + the annotation pair. `TestGeneratedApplyCarriesTheScopeAnnotationItsSiblingsDo` compares each `apply` against its own resource's `create` and fails if the create has none, so it can't pass vacuously.

**(9) rename hint bypasses the envelope** — hint now travels on the `*exitcode.Error`, so it reaches both the JSON `hint` field and `FprintError`'s `hint:` line.

**(10) stdin truncation**, **(11) platform `applySet`**, **(12) `Stat()` error** — all three applied as suggested, in both `internal/platform` and `internal/security`.

**(13) zero coverage on the generated `apply` RunE** — `internal/commands/platform_generated_apply_test.go`: create/update fork, ambiguity and idless-match refusals (asserting zero writes), list-500 (zero writes), `--dry-run` on both branches, the PATCH Content-Type, the scope annotation. Assertions are on which requests reached the server, not on the message printed.

**(14) `NameLookupField`** — set only when `NameField` actually resolved to it.

**(15) upload leaves** — each `--file` description now says why it isn't `--from-file` ("a multipart upload needs a name and a length, so it cannot be piped").

## Not actioned: (2)

`GroupListItem` in `specs/platform/securitycloud_device_groups_api.json` declares `id` and requires only `name` — "omitted for the implicit *Default Group* entry, which is not a stored group". So `apply` on a stored group resolves normally.

Wire-checked through the CLI, EU environment credential, 2026-09-08:

```
$ jamf-cli security device-groups list -o json
[ { "name": "Default Group" } ]

$ jamf-cli security device-groups create --set name=pr372-idprobe -o json
{ "href": "...", "id": "8cf9dafc-...", "name": "pr372-idprobe" }

$ jamf-cli security device-groups list -o json
[ { "name": "Default Group" },
  { "id": "8cf9dafc-...", "name": "pr372-idprobe" } ]

$ jamf-cli security device-groups apply --set name=pr372-idprobe --yes
Updated device-group "pr372-idprobe" (id: 8cf9dafc-...)
```

Second run updated; no duplicate. The probe group was deleted afterwards.

You were reading CLAUDE.md correctly — **the note was wrong.** Its 2026-09-05 probe ran on a tenant holding only the Default Group and generalised from a collection of one, so it could not tell a property of the endpoint from a property of that row. Corrected, along with the ZTNA nameless-app note the same paragraph pointed at.

The Default Group *is* a real gap, and it's the one your failure scenario describes — before this pass it reported `ErrNotFound`, which `apply` reads as "create it". Now:

```
$ jamf-cli security device-groups apply --set "name=Default Group" --yes
found 1 item(s) named "Default Group" in /securitycloud/v2/groups,
but the list returns no ID for them; identify the item by ID instead
```

Fixed at the resolver rather than by blocklisting the resource, because the same shape recurs (CLAUDE.md had already flagged it as wanted for ZTNA's nameless apps) and a blocklist would cost the five stored-group cases that work.

## Decomposition

Keeping it as one PR. The three concerns are separable, as you say, but they were wire-verified together against one tenant and the `apply` half is only safe *because* of the resolver change in this same diff — splitting would ship `apply` on top of a lookup that still reads an unaddressable match as absence.

`make test`, `make lint` (0 issues) and `make verify-generated` all pass.

ktn-jamf (approved):
> [!TIP]
> ✅ **Merge-ready**. All 15 prior findings addressed, 12 fixed as suggested, one fix built at the resolver rather than the resource, two nice-to-haves done. No new issues.

## Rating: 5/5
- Clean. Every fix is tested against the exact failure scenario the finding described, and the corrected diagnosis on (2) is backed by a fresh wire probe and an honest CLAUDE.md correction.

<details>
<summary>Prior findings status (15 items)</summary>

| (1) | `site` | ✅ Fixed | `git rm --cached`, `/site` added to `.gitignore` |
| (2) | `internal/platform/resolve_generic.go` | ✅ Fixed | See below — diagnosis corrected, underlying duplicate-creation bug fixed at the resolver |
| (3) | `generator/platform/emitter.go:701` | ✅ Fixed | `UpdateMergePatch` now derives from the method; `TestGeneratedApplySendsThePatchContentTypeItsSiblingDoes` |
| (4) | `generator/platform/emitter.go` (`applyLong`) | ✅ Fixed | Disclosed in every `apply --help`, not only the 4 originally flagged |
| (5) | `internal/platform/resolve_generic.go` (`collectMatches`) | ✅ Fixed | Matches accumulate across pages; `TestResolveIDByName_AmbiguousAcrossPages` |
| (6) | `generator/platform/emitter.go` (`platformApplyCaveats`) | ✅ Fixed | Help states the draft/publish split; publish stays manual (non-idempotent) |
| (7) | `internal/platform/resolve_generic.go:124` | ✅ Fixed | "identify it by ID instead"; `TestResolveIDByName_AmbiguityRemedyIsNotAPositional` |
| (8) | `generator/platform/emitter.go` (`unionScopeTypes`) | ✅ Fixed | `TestGeneratedApplyCarriesTheScopeAnnotationItsSiblingsDo` |
| (9) | `internal/commands/root.go:988` | ✅ Fixed | Hint now travels on `*exitcode.Error`, reaches both JSON and text output |
| (10) | `internal/platform/body.go`, `internal/security/body.go` | ✅ Fixed | Reads one byte past the cap; errors explicitly over the limit |
| (11) | `internal/platform/body.go` (`applySet`) | ✅ Fixed | Now errors on a non-object clobber, matching `internal/security` |
| (12) | `internal/platform/body.go`, `internal/security/body.go` | ✅ Fixed | `Stat()` error now returned explicitly, not collapsed into "no input" |
| (13) | `internal/commands/platform_generated_apply_test.go` | ✅ Fixed | New `httptest`-backed suite: fork, dry-run, Content-Type, scope annotation |
| (14) | `generator/platform/emitter.go:673` | ✅ Fixed | `NameLookupField` set only when `NameField` actually resolved to it |
| (15) | 11 upload leaves | ✅ Fixed | Each `--file` description now says why it isn't `--from-file` |

</details>

**On (2):** the fix is real, but my original diagnosis overstated its scope. `security device-groups apply` does not *always* fail to find an existing group — a stored group's list entry carries an `id`, confirmed by re-probing create→list on a live tenant. Only the implicit "Default Group" row lacks one. The CLAUDE.md note I cited as evidence was itself wrong: it generalized from a tenant holding a single row, and the diff corrects it in place with a fresh two-row probe. The gap that note *did* correctly point at — a name match with no ID silently reading as "not found," which `apply` treats as "create it" — is fixed at `resolveIDByName` itself (`collectMatches`' nameless branch) rather than by blocklisting the resource, so the fix also closes the CLAUDE.md-documented ZTNA nameless-app case wherever a name match with no ID recurs. `TestResolveIDByName_MatchedWithNoID` pins it directly: the Default Group now refuses with "the list returns no ID for them," never `ErrNotFound`, so `apply` can no longer read it as absent and create a duplicate.

No new findings from the fix commits. The one shared-predicate change worth flagging as such — `resolveIDByName`/`collectMatches` is consulted by every `--name` lookup across the Platform surface, not only `apply` — I read in full myself rather than re-dispatching a specialist for it: the existing `--name`-lookup tests (`TestResolveIDByName_NotFound`, `_Ambiguous`, `_MatchesByTitle`, etc.) all still pass unchanged, and the new nameless-branch behavior is additive (a new error case) rather than a change to any path those tests already cover.

<details>

- [x] Correctness: every fix re-verified against its cited lines and a passing, targeted test
- [x] Test coverage: `internal/commands/platform_generated_apply_test.go`, `internal/platform/resolve_generic_test.go` read in full
- [x] Documentation currency: CLAUDE.md's device-groups and ZTNA notes corrected and consistent with the new code
- [x] Shared-predicate risk: `resolveIDByName`/`collectMatches` read directly; no regression in existing `--name`-lookup tests
- [x] Build and test: `go build ./...` and `go test ./generator/platform/... ./internal/platform/... ./internal/security/... ./internal/commands/...` all green at `504b88b`

Diff since baseline: 29 files, +665/−75, `745afeb..504b88b`. Lanes re-dispatched this round: none — the fix commits were directly responsive to prior findings and verified in full by direct reading. Never run: `fidelity-reviewer` (no linked spec/issue), `performance-reviewer` (still inapplicable — no new DB/loop/caching surface).

</details>



_Generated by `pr-review` v1.33.0, a [Jamf Claude Code skill](https://github.com/jamf/claude-plugins/blob/main/plugins/pr-review/README.md)_

